PR #17, opened 2021-02-06
- Add head http method.

plombard:
Hello, I was looking for a drop-in replacement for httpie, which is fantastic but add a 300ms overhead to every request on my work machine, and I think ht should be ideal for helping me develop my APIs. I use HEAD endpoints in my APIs so I needed this method included. I'm not a seasoned Rust developer so it may not be _perfect_ code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ht"
-version = "0.4.0"
+version = "0.5.0"
 authors = ["ducaale <sharaf.13@hotmail.com>"]
 edition = "2018"
 license = "MIT"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ atty = "0.2"
 jsonxf = "1.0"
 indicatif = "0.15.0"
 lazy_static = "1.4.0"
+predicates = "1.0.7"
 regex = "1"
 reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "stream", "json", "gzip", "brotli", "multipart"] }
 rpassword = "5.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ atty = "0.2"
 jsonxf = "1.0"
 indicatif = "0.15.0"
 lazy_static = "1.4.0"
-predicates = "1.0.7"
 regex = "1"
 reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "stream", "json", "gzip", "brotli", "multipart"] }
 rpassword = "5.0.0"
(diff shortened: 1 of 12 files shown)


ducaale:
Thanks, @plombard. It seems this PR is overlapping with another [PR](https://github.com/ducaale/ht/pull/16) that is adding support for the `HEAD` HTTP method. Maybe you could update this PR to support the `OPTIONS` method instead? Thanks again.

plombard:
Sure, I'll give it a try.

tests/cli.rs
```diff
@@ -74,3 +74,25 @@ fn basic_head() -> Result<(), Box<dyn std::error::Error>> {
...
+
+#[test]
+fn basic_options() -> Result<(), Box<dyn std::error::Error>> {
+    let mut cmd = Command::cargo_bin("ht")?;
+    cmd.arg("-v")
+        .arg("--offline")
+        .arg("--ignore-stdin")
+        .arg("--pretty=format")
+        .arg("options")
+        .arg("httpbin.org");
```
ducaale, line 86: Can we change the URL `httpbin.org` to `httpbin.org/json` since the latter has more interesting headers?

```
$ http options
HTTP/1.1 200 OK
access-control-allow-methods: GET, POST, PUT, DELETE, PATCH, OPTIONS
access-control-max-age: 3600
allow: GET, HEAD, OPTIONS
connection: keep-alive
content-length: 0
content-type: text/html; charset=utf-8
date: Sun, 07 Feb 2021 11:28:45 GMT
server: gunicorn/19.9.0
```

ducaale (changes requested):

tests/cli.rs
plombard: Sure, but in that case, we won't be able to test with the _--offline_ flag, as these headers are the ones sent back by the server. Or maybe we should launch a dedicated http server during the tests to send our tests against, instead of verifying the requests sent ? For the time being I'll remove the offline flag and test directly againt httpbin.org.

ducaale: Oh, right. It slipped my mind that we were using the _--offline_ flag for other tests.

> maybe we should launch a dedicated HTTP server during the tests to send our tests against, instead of verifying the requests sent?

Sounds good. Personally, I've been using a small Deno script for my local testing. But it would be nice if there was an automated solution that works with `cargo test`.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ atty = "0.2"
 jsonxf = "1.0"
 indicatif = "0.15.0"
 lazy_static = "1.4.0"
+predicates = "1.0.7"
```
ducaale, line 20: I think this could be moved to the `dev-dependencies` section.

ducaale (changes requested):
One last nitpick. Other than that, everything looks good.

ducaale (approved):
🚀🚀

PR #19, opened 2021-02-06
Add control for redirects

Till--H:
Use same arguments as HTTPie for following redirects (`--follow` and `--max-redirects`).


ducaale:
Thanks, @Till--H!

PR #21, opened 2021-02-06
Render white text as the default foreground color

blyxxyz:
The default theme is unfortunately hard to read with a light background:
image
Treating white text as uncolored fixes that:
image
That also seems to match HTTPie.


ducaale:
Thanks for the fix @blyxxyz!

PR #22, opened 2021-02-06
Init parsing regex for RequestItem only once

jRimbault:
I thought maybe you might appreciate that. Perhaps also investigate replacing `lazy_static` with the safer `once_cell::sync::Lazy` ?

And BTW in Rust 2018 edition you don't need :

```rust
#[macro_use]
extern crate name;
```


ducaale:
Thanks for PR, @jRimbault.

>Perhaps also investigate replacing lazy_static with the safer once_cell::sync::Lazy?

I agree. Also, it seems that [once_cell](https://github.com/rust-lang/rust/issues/74465) will be part of std in the future which is great!

PR #24, opened 2021-02-06
Use write_all instead of write to ensure everything is written

blyxxyz:
`write` doesn't always write everything, but `write_all` makes sure you don't drop any output.

Found by `cargo clippy`: https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount

Clippy also gives some warnings about small inefficiencies and code that could be shorter, but this was the only functional thing.


ducaale:
Thanks for mentioning `cargo clippy`. I should consider adding it to the CI pipeline along with `cargo fmt`.

PR #25, opened 2021-02-06
- Add basic handling of sessions.

plombard:
This one has a bit more changes. 
It emulates the session handling of httpie, by dumping the headers (including auth headers) in a file in the user home directory, and try to load and send again those headers if a session of the same name is requested. 
If additional headers are added to the next requests, there is a merge between those new (or modified) headers and the previous ones, so the user can keep relevant old headers (like auth token) but provide new info. 
There is no way to clear a session yet, except by deleting the json file in ~/.ht/.


blyxxyz:
Maybe it's better to put the files in `$XDG_CONFIG_HOME/ht` (i.e. `~/.config/ht`)? See for example [how HTTPie does it](https://github.com/httpie/httpie/blob/e944dbd7fa467da4fba49f00253a6ea30015fba2/httpie/config.py#L51).

It seems like one of those things that are best to overengineer from the start, so you don't have to change it later and keep supporting legacy configuration. HTTPie goes as far as `~/.config/httpie/sessions/<domain name>/<session name>.json`.

plombard:
> It seems like one of those things that are best to overengineer from the start, so you don't have to change it later and keep supporting legacy configuration.

You are absolutely right, and this is a great suggestion. I'll change the directory right now. If it is to be moved/modified later, it should have minimal impact on the users this way.

src/main.rs
```diff
@@ -45,7 +46,40 @@ async fn main() -> std::result::Result<(), Box<dyn std::error::Error>> {
...
+        Some(identifier) => {
+            match Session::load(&identifier, &host) {
+                Err(why) => panic!("couldn't load session {}: {}", &identifier, why),
+                Ok(result) => result,
+            }
+        },
+    };
+    let session_for_merge = previous_session.clone();
+    // Use auth from previous session if no auth present
+    match (auth.is_none(), previous_session) {
```
phrohdoh, line 64: If being rid of the empty match arm is desired, I believe something like the following could be used.
```rust
if let (true, Some(p)) = (auth.is_none(), previous_session) {
    auth = p.auth;
}
```

src/main.rs
```diff
@@ -45,7 +46,40 @@ async fn main() -> std::result::Result<(), Box<dyn std::error::Error>> {
...
+    };
+    let saved_auth = auth.clone();
+    // Merge headers from parameters and previous session
+    let (headers, headers_to_unset) = request_items.headers(&session_for_merge);
+    // Save the current session if present
+    match arg_session {
+        None => (),
+        Some(identifier) => {
+            let new_session = Session::new(identifier, &host, request_items.export_headers(&session_for_merge), saved_auth);
+            match new_session.save() {
```
phrohdoh, line 76: This could be `if let Err(why) = new_session.save() { ... }` if removing the `Ok` arm (which doesn't have a meaningful body) is desired.

src/request_items.rs
```diff
@@ -31,7 +39,7 @@ impl RequestItems {
         count
     }
 
-    pub fn headers(&self) -> (HeaderMap<HeaderValue>, Vec<HeaderName>) {
+    pub fn headers(&self, session: &Option<Session>) -> (HeaderMap<HeaderValue>, Vec<HeaderName>) {
```
phrohdoh, line 42: Any particular reason to use `&Option<Session>` here instead of `Option<&Session>`?

src/session.rs
```diff
@@ -0,0 +1,83 @@
...
+pub struct Session {
+    pub identifier: String,
+    pub host: String,
+    pub headers: Vec<Parameter>,
+    pub auth: Option<Auth>,
+    ht_version: String,
+}
+
+impl Session {
+    pub fn new(identifier: String, host: &str, headers: Vec<Parameter>, auth: Option<Auth>) -> Session {
```
phrohdoh, line 22: If we're going to pay the cost of converting `host` to a `String` unconditionally, I believe we should have the calling code do that explicitly (which'd mean we'd use `host: String` here) instead of hiding it from the user.

src/session.rs
```diff
@@ -0,0 +1,83 @@
...
+            host: host.to_string(),
+            headers,
+            auth,
+            ht_version: String::from(VERSION),
+        }
+    }
+
+    pub fn save(&self) -> std::io::Result<String> {
+        let json = serde_json::to_string(&self)?;
+        let identifier = &self.identifier;
```
phrohdoh, line 34: You can have values in a `String` that are not permitted by some filesystems.
Could this go through some check/conversion first?

src/session.rs
```diff
@@ -0,0 +1,83 @@
...
+        Session {
+            identifier,
+            host: host.to_string(),
+            headers,
+            auth,
+            ht_version: String::from(VERSION),
+        }
+    }
+
+    pub fn save(&self) -> std::io::Result<String> {
```
phrohdoh, line 32: Documenting what value this returns (the session's identifier) would be helpful.
Currently the function signature tells us _some `String`_ but without docs that may as well be a hardcoded `"test".into()` given what I know as a user.

src/session.rs
```diff
@@ -0,0 +1,83 @@
...
+        let json = serde_json::to_string(&self)?;
+        let identifier = &self.identifier;
+        let mut config_dir = match ensure_session_dir_exists(&self.host) {
+            Err(why) => panic!("couldn't get config directory: {}", why),
+            Ok(dir) => dir,
+        };
+        config_dir.push(identifier);
```
phrohdoh, line 39: IMO it'd be more clear if we did `config_dir.push(&self.identifier)` and in for tail expression `Ok(identifier.clone())` (which'd mean no longer having the `let identifier = &self.identifier` line).

src/main.rs
plombard: Sure, thanks.

src/main.rs
plombard: Right, thanks again.

src/session.rs
plombard: Ok, got it.

src/session.rs
plombard: True. Besides, the `Result<String>` is no longer needed in this version of the code (I did use the returned identifier in a message at one point), so I replaced it with a `Result<()>`. I hope it makes more sense.

src/session.rs
plombard: Ok, I handle this below, it's no longer needed I think.

src/session.rs
plombard: Ok, I think this is a valid concern. I tried to find a suitable validator for filenames, but most wouldn't fail something like `--session "http://microsoft.com/"`, as I think they should, to prevent the creation of a file with an error-prone filename. (note : httpie's `http  httpbin.org/basic-auth/me/pass --auth-type basic --auth me:pass --session "http://httpbin.org/" --verbose` doesn't fail either).
So I... gave up, and went for an easier path 😄 
I changed the sessions filenames to be 10-chars hashes (at the price of a bit more complexity and another dependency, on the blake2 crate, as it seems the most popular). So no more directly user-defined filenames, and as the identifier is still included as a json field, a user can still find which file contains which session.
What do you think ?

phrohdoh: That sounds great to me!

The owner/maintainer (not me) will need to approve the new dependency of course.

blyxxyz: Hmm. HTTPie checks for a path separator in the session name. If there is one, then it points directly to a file, so it's not relative to `~/.config/httpie/sessions/`.

It accepts `http://httpbin.org/` because it helpfully creates a directory called `http:` (in your current directory) to put the session file in.

But as long as we only ever put sessions in `~/.config` we don't have to handle that. So maybe this logic?
- If there's a path separator in the session name (`name.contains(std::path::is_separator)`), reject with a message saying explicit session filenames are not supported.
- Otherwise, just try creating the file, and if there's something else wrong with the filename you'll get an error to show to the user.

src/request_items.rs
```diff
@@ -43,9 +51,50 @@ impl RequestItems {
                 _ => {}
             }
         }
+        // handle session additional headers
+        match session {
```
phrohdoh, line 55: `if let Some(s) = session { /* use `s` */ }`

src/main.rs
```diff
@@ -63,7 +64,38 @@ async fn main() -> Result<()> {
...
+    let session_for_merge = previous_session.clone();
+    // Use auth from previous session if no auth present
+    if let (true, Some(p)) = (auth.is_none(), previous_session) {
+        auth = p.auth;
+    }
+    let saved_auth = auth.clone();
+    // Merge headers from parameters and previous session
+    let (headers, headers_to_unset) = request_items.headers(session_for_merge.as_ref())?;
+    // Save the current session if present
+    match arg_session {
```
phrohdoh, line 89: `if let Some(identifier) = arg_session { ... }`

src/request_items.rs
```diff
@@ -43,9 +51,50 @@ impl RequestItems {
...
+            }
+        }
         Ok((headers, headers_to_unset))
     }
 
+    pub fn export_headers(&self, session: Option<&Session>) -> Vec<Parameter> {
+        let mut headers = vec![];
+        let mut headernames_present = vec![];
+        for item in &self.0 {
+            match item {
```
phrohdoh, line 73: `if let RequestItem::HttpHeader(key, value) = item { ... }`

src/request_items.rs
```diff
@@ -43,9 +51,50 @@ impl RequestItems {
...
+                        name: String::from(key),
+                        value: String::from(value),
+                    });
+                    headernames_present.push(String::from(key));
+                }
+                _ => {}
+            }
+        }
+        // handle session additional headers
+        match session {
```
phrohdoh, line 85: `if let Some`

src/utils.rs
```diff
@@ -20,6 +23,8 @@ use tokio_util::codec::{BytesCodec, FramedRead};
 use crate::Body;
 use crate::Theme;
 
+extern crate dirs;
```
phrohdoh, line 26: Not necessary in non-`2015` edition Rust.

src/cli.rs
```diff
@@ -188,6 +193,19 @@ fn parse_method_url(s: &str) -> Result<(Option<Method>, String)> {
     }
 }
 
+fn validate_session_name(s: &str) -> Result<String> {
```
phrohdoh, line 196: If we aren't tying session names (identifiers? unclear due to seemingly-mixed terminology) to paths, do we want to validate that the name is a valid path?

src/request_items.rs
```diff
@@ -43,9 +51,50 @@ impl RequestItems {
...
+                        .or_insert(HeaderValue::from_str(&h.value).unwrap());
+                }
+            }
+        }
         Ok((headers, headers_to_unset))
     }
 
+    pub fn export_headers(&self, session: Option<&Session>) -> Vec<Parameter> {
+        let mut headers = vec![];
+        let mut headernames_present = vec![];
```
phrohdoh, line 71:
```suggestion
        let mut header_names_present = vec![];
```

src/session.rs
```diff
@@ -0,0 +1,83 @@
+extern crate dirs;
```
phrohdoh, line 1: same here (not necessary in [non-`2015` edition](https://github.com/ducaale/ht/commit/f4332e23b5a425049c0f76d1e31136006262b683#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R5) Rust)

tests/cli.rs
```diff
@@ -101,3 +103,141 @@ fn basic_options() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+fn clear_session_file(name: String) -> std::io::Result<()> {
```
phrohdoh, line 107:
```suggestion
fn delete_session_file(name: String) -> std::io::Result<()> {
```

src/utils.rs
```diff
@@ -170,3 +175,22 @@ macro_rules! vec_of_strings {
         vec![$(String::from($str),)*] as Vec<String>
     });
 }
+
+// Create the directory ~/.config/ht/sessions/{domain} if not present.
```
phrohdoh, line 179: This comment isn't always correct (if `dirs::config_dir()` returns something other than (an equivalent to) `~/.config/ht/sessions/{domain}`).

```suggestion
// Create the directory which will contain sessions for `domain` if not present.
```

src/utils.rs
```diff
@@ -170,3 +175,22 @@ macro_rules! vec_of_strings {
...
+    config_dir.push("ht");
+    config_dir.push("sessions");
+    config_dir.push(domain);
+    fs::create_dir_all(&config_dir)?;
+    Ok(config_dir)
+}
+
+// Calculate an hash from session identifier to use as the session filename.
+pub fn session_filename(identifier: &str) -> String {
+    let hash = Blake2b::new().chain(identifier).finalize();
```
phrohdoh, line 194: `identifier` here but `name` [here](https://github.com/ducaale/ht/pull/25/files#diff-bcb711f1b6fb02677c65dec5c9027deaf435a7fc7b5191a8e25e7256d37773deR116)

src/utils.rs
```diff
@@ -170,3 +175,22 @@ macro_rules! vec_of_strings {
...
+    format!("{:x}", hash).get(0..10).unwrap().to_string()
```
phrohdoh, line 195: Why only `0..=9` instead of the entire hash?

tests/cli.rs
```diff
@@ -101,3 +103,141 @@ fn basic_options() -> Result<(), Box<dyn std::error::Error>> {
...
+    // Clear session file
+    let mut config_dir = match dirs::config_dir() {
+        None => panic!("couldn't get config directory"),
+        Some(dir) => dir,
+    };
+    config_dir.push("ht");
+    config_dir.push("sessions");
+    config_dir.push("httpbin.org");
+    let hash = Blake2b::new().chain(name).finalize();
+    config_dir.push(format!("{:x}", hash).get(0..10).unwrap().to_string());
```
phrohdoh, line 117: Possibility for this and the actual utility function to get out of sync.
Can this test utility function just use the actual util logic?

src/utils.rs
plombard: Honestly, it's just a matter of having a shorter filename while still being mostly collision-free, a bit like what docker is doing with the short ids of the running containers. I tried with the full hash and it was _really_ long when listing the sessions directory.

src/utils.rs
plombard: Whoops, sorry.

src/cli.rs
plombard: Yes, in theory if we choose to use hashes we shouldn't need to validate the name of the session (which is the filename no more) for being a valid path.  Plus, it clears the context of both name and identifier. Identifier is the user-defined name of the session, and the filename is its calculated hash.

tests/cli.rs
plombard: Well, I'd like to, but I don't know how to do that in Rust without having to create a new crate for the `utils` functions. :(

src/session.rs
plombard: Ok, it's simpler, and it fits well with the validation already in place, thanks.

src/request_items.rs
plombard: No particular reason, I'm afraid I just wanted to use a ref on the function argument, the built-in mechanism to do this was lost on me. This is really helpful (as every comment on this PR), thanks.

tests/cli.rs
ducaale: Maybe the `config` dir could be made to point to the `temp` directory during tests? That way, any created file will be deleted regardless of whether tests have failed or not.

src/main.rs
```diff
@@ -62,7 +63,34 @@ async fn main() -> Result<()> {
...
+    // load previous session if present
+    let previous_session = match args.session {
+        None => None,
+        Some(ref identifier) => {
+            match Session::load(identifier, &host) {
+                Err(why) => panic!("couldn't load session {}: {}", &identifier, why),
+                Ok(result) => result,
+            }
+        },
+    };
```
ducaale, line 77: Could you replace this panic with an `anyhow` error? we've been replacing all our panics with errors that are bubbled up.
```rust
use anyhow::Context;

let previous_session = match args.session {
    None => None,
    Some(ref identifier) => Session::load(identifier, &host)
        .context(format!("couldn't load session {}", identifier))?
};
```

src/main.rs
```diff
@@ -62,7 +63,34 @@ async fn main() -> Result<()> {
...
+    let session_for_merge = previous_session.clone();
+    // Use auth from previous session if no auth present
+    if let (true, Some(p)) = (auth.is_none(), previous_session) {
+        auth = p.auth;
+    }
+    let saved_auth = auth.clone();
+    // Merge headers from parameters and previous session
+    let (headers, headers_to_unset) = request_items.headers(session_for_merge.as_ref())?;
```
ducaale, line 85: I would have liked it if we could limit `request_items.rs`'s responsibility to parsing current request_items.
1. `request_items.headers()` would produce `reqwest::header::HeaderMap` from user input.
2. `session.rs` would construct `reqwest::header::HeaderMap` from prev session.
3. `session.rs` would take the two headers and merge them.

src/main.rs
```diff
@@ -62,7 +63,34 @@ async fn main() -> Result<()> {
...
+    // Save the current session if present
+    if let Some(identifier) = args.session {
+        let new_session = Session::new(identifier, host, request_items.export_headers(session_for_merge.as_ref()), saved_auth);
+        if let Err(why) = new_session.save() {
+            panic!("couldn't save session {}: {}", new_session.identifier, why);
```
ducaale, line 90: Can we replace this with an `anyhow` error?

ducaale:
Superseded by https://github.com/ducaale/xh/pull/125

PR #29, opened 2021-02-07
README - update macOS instructions to use homebrew-core

dtrodrigues:
`ht` is now included in [Homebrew core](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ht-rust.rb) so the installation instructions can be updated to point there instead of needing a separate tap.


ducaale:
Woohoo🎉! Thank you very much @dtrodrigues!

PR #33, opened 2021-02-07
Add User-Agent header

blyxxyz:
This also adds a rudimentary way to detect an integration test. If that's useful for other things it could be extracted to a function (or lazy static) in `utils`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,9 @@
+use std::env;
+
 use atty::Stream;
-use reqwest::header::{HeaderValue, ACCEPT, ACCEPT_ENCODING, CONNECTION, CONTENT_TYPE, RANGE};
+use reqwest::header::{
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -21,6 +25,17 @@ use reqwest::redirect::Policy;
 use url::Url;
 use utils::body_from_stdin;
 
+fn get_user_agent() -> &'static str {
```
ducaale, line 28: I usually prefer to move any function that doesn't belong in any module to `utils.rs`. What do you think?

blyxxyz: I think this one is best off in the same module as where it's used, since it isn't used anywhere else. That also means it can stay private.

But I would also be ok with moving it, if you'd like.

ducaale (approved):
LGTM

PR #37, opened 2021-02-07
Bubble errors up to main() instead of panicking

### LINKED ISSUE #14: Broken pipe panic when piping output
Thanks for this, it's dramatically faster than the original when pretty printing big JSON responses!

When piping to `head` or `less` I'm seeing a broken pipe-related panic when the pager exits. Seems likely to be related to this? https://github.com/rust-lang/rust/issues/46016

Reproducing it requires a response of sufficient length:

```
22:02:25 ~ > RUST_BACKTRACE=1 ht get 'http://httpbin.org/stream/150' | head
{"url": "http://httpbin.org/stream/150", "args": {}, "headers": {"Host": "httpbin.org", "X-Amzn-Trace-Id": "Root=1-601e06c4-02a347b5507c64fe408cf5b6", "Accept-Encoding": "gzip, deflate", "Accept": "*/*"}, "origin": "104.162.100.223", "id": 0}

{"url": "http://httpbin.org/stream/150", "args": {}, "headers": {"Host": "httpbin.org", "X-Amzn-Trace-Id": "Root=1-601e06c4-02a347b5507c64fe408cf5b6", "Accept-Encoding": "gzip, deflate", "Accept": "*/*"}, "origin": "104.162.100.223", "id": 1}

{"url": "http://httpbin.org/stream/150", "args": {}, "headers": {"Host": "httpbin.org", "X-Amzn-Trace-Id": "Root=1-601e06c4-02a347b5507c64fe408cf5b6", "Accept-Encoding": "gzip, deflate", "Accept": "*/*"}, "origin": "104.162.100.223", "id": 2}

{"url": "http://httpbin.org/stream/150", "args": {}, "headers": {"Host": "httpbin.org", "X-Amzn-Trace-Id": "Root=1-601e06c4-02a347b5507c64fe408cf5b6", "Accept-Encoding": "gzip, deflate", "Accept": "*/*"}, "origin": "104.162.100.223", "id": 3}

{"url": "http://httpbin.org/stream/150", "args": {}, "headers": {"Host": "httpbin.org", "X-Amzn-Trace-Id": "Root=1-601e06c4-02a347b5507c64fe408cf5b6", "Accept-Encoding": "gzip, deflate", "Accept": "*/*"}, "origin": "104.162.100.223", "id": 4}

thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', library/std/src/io/stdio.rs:895:9
stack backtrace:
   0: rust_begin_unwind
             at /build/rustc-71Ymwo/rustc-1.47.0+dfsg1+llvm/library/std/src/panicking.rs:475
   1: std::panicking::begin_panic_fmt
             at /build/rustc-71Ymwo/rustc-1.47.0+dfsg1+llvm/library/std/src/panicking.rs:429
   2: std::io::stdio::print_to
             at /build/rustc-71Ymwo/rustc-1.47.0+dfsg1+llvm/library/std/src/io/stdio.rs:895
   3: std::io::stdio::_print
             at /build/rustc-71Ymwo/rustc-1.47.0+dfsg1+llvm/library/std/src/io/stdio.rs:907
   4: ht::buffer::Buffer::write
   5: ht::main::{{closure}}
   6: std::thread::local::LocalKey<T>::with
   7: tokio::runtime::enter::Enter::block_on
   8: tokio::runtime::thread_pool::ThreadPool::block_on
   9: tokio::runtime::Runtime::block_on
  10: ht::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```

Installed from cargo, rustc 1.47.0
Ubuntu 20.10 / Linux 5.8.0-41-generic

blyxxyz:
In theory, all panics after this should be bugs instead of normal error conditions.

The error reporting may still need finetuning. For example, if the hostname doesn't resolve you get a redundant traceback:

```
Error: error sending request for url (http://bad-hostname/): error trying to connect: dns error: failed to lookup address information: Name or service not known

Caused by:
    0: error trying to connect: dns error: failed to lookup address information: Name or service not known
    1: dns error: failed to lookup address information: Name or service not known
    2: failed to lookup address information: Name or service not known
```

This commit also fixes all of the remaining Clippy warnings.

Resolves #14 (?)

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -1,5 +1,6 @@
-use std::fmt::Write;
+use std::io::{self, Write};
 
+use anyhow::Result;
 use reqwest::header::{HeaderMap, HeaderName, HeaderValue, CONTENT_LENGTH, HOST};
 use reqwest::{Request, Response};
 
@@ -29,7 +30,7 @@ pub struct Printer {
 
 impl Printer {
     pub fn new(pretty: Option<Pretty>, theme: Option<Theme>, stream: bool, buffer: Buffer) -> Self {
-        let pretty = pretty.unwrap_or(Pretty::from(&buffer));
+        let pretty = pretty.unwrap_or_else(|| Pretty::from(&buffer));
         let theme = theme.unwrap_or(Theme::Auto);
 
         Printer {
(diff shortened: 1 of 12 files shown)


src/printer.rs
```diff
@@ -86,16 +85,20 @@ impl Printer {
...
         for (key, value) in headers {
-            let key = key.to_string();
-            let value = value.to_str().unwrap();
-            writeln!(&mut header_string, "{}: {}", key, value).unwrap();
+            header_string.push_str(key.as_str());
+            header_string.push_str(": ");
+            match value.to_str() {
+                Ok(value) => header_string.push_str(value),
+                Err(_) => header_string.push_str(&format!("{:?}", value)),
+            }
```
ducaale, line 93: Rather than being silent, I think it is ok to panic if an invalid header value is detected
```
pub fn to_str(&self) -> Result<&str, ToStrError>
------------------------------------------------
Yields a &str slice if the HeaderValue only contains visible ASCII chars.
```

blyxxyz: Hmm. Such a value is weird but technically valid:

> In practice, HTTP header field values are usually valid ASCII. However, the HTTP spec allows for a header value to contain opaque bytes as well. In this case, the header field value is not able to be represented as a string.

I figure that if reqwest allowed it to be constructed it's best to just print it. It's probably only relevant for responses though.

And making the program more verbose should probably never cause it to fail, it'd be bad for it to crash when you tried to see more. But I didn't think of that while writing so I'm not sure that's upheld.

blyxxyz: On second thought, it's upheld in functions that return an `io::Result`, because those can only fail for I/O reasons.

`print_json()` is the only one that can fail for non-I/O reasons, if `indent_json()` fails, because the JSON is invalid. But I think the user would want to know what's invalid about it, and rejecting it based on configuration is bad. Maybe it should just fall back to the unindented original? With or without a warning? What do you think?

ducaale: >And making the program more verbose should probably never cause it to fail, it'd be bad for it to crash when you tried to see more.

👍

>Maybe it should just fall back to the unindented original? With or without a warning? What do you think?

Sounds good. Since `print_json()` will not write anything until `indent_json()` succeeds, I think the simplest option would be to do the fallback inside `print_json()` i.e `indent_json(text).unwrap_or(text)`.

ducaale (approved):
Approved 🎉🎉🎉

PR #41, opened 2021-02-08
Remove use of async, make --stream work consistently

### LINKED ISSUE #39: Either remove or increase async
`ht` uses async, but it uses it inconsistently and I'm not sure that it needs it.

Is it likely that it will start doing multiple requests at a time in the future? If so, then blocking APIs like `std::fs` should be avoided. If not, then maybe [reqwest's blocking API](https://docs.rs/reqwest/0.11.0/reqwest/blocking/index.html) would be easier to use.

blyxxyz:
This also
- Makes streaming supported under all circumstances
- Makes binary data detection and trailing newlines more consistent
- Removes some obsolete dependencies from Cargo.toml

Resolves #39.


src/printer.rs
```diff
@@ -43,43 +42,48 @@ impl Printer {
...
     }
 
-    fn print_html(&mut self, text: &str) -> io::Result<()> {
+    fn print_syntax(&mut self, body: &mut impl Read, syntax: &'static str) -> io::Result<()> {
         if self.color {
-            colorize(text, "html", &self.theme, &mut self.buffer)
+            let mut buf = Vec::new();
+            body.read_to_end(&mut buf)?;
+            let text = String::from_utf8_lossy(&buf);
+            self.colorize(&text, syntax)
```
blyxxyz, line 78: Leaving a note to myself:

This (as well as `print_json()`'s coloring) is wrong, because it doesn't pay attention to `self.stream`.

A [`BufReader`](https://doc.rust-lang.org/std/io/struct.BufReader.html) could be used to read [line by line](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.lines). That's better than manually calling `read()`, which could split at awkward places, even inside a UTF-8 character. Using `read_until()` might be best so the final newline isn't lost.

blyxxyz: [`LineWriter`](https://doc.rust-lang.org/std/io/struct.LineWriter.html) turned out to be easier to hook up to `jsonxf` and somehow performs much better.

blyxxyz:
This kept getting more complicated as I found new difficulties with `--stream` :grimacing: 

But I think it's about ready. Streaming output is now decoded even if it isn't UTF-8, and it's printed immediately as it arrives even if it has to be formatted or highlighted. That didn't work before.

This depends on gamache/jsonxf#8. I've put my own repository in `Cargo.toml` for the time being, but it can't be published on crates.io this way, so that should be replaced before this is merged (or failing that, before the next release).

blyxxyz:
I've vendored JSONXF, so this would be publishable on crates.io.

src/printer.rs
```diff
@@ -159,79 +206,107 @@ impl Printer {
...
+    pub fn print_response_body(&mut self, mut response: Response) -> anyhow::Result<()> {
+        if !self.buffer.is_terminal() {
+            // No trailing newlines, no binary detection, no decoding, direct streaming
+            self.print_body_stream(get_content_type(&response.headers()), &mut response)?;
+        } else if self.stream {
+            self.print_body_stream(
+                get_content_type(&response.headers()),
+                &mut decode_stream(&mut response),
+            )?;
+            self.buffer.print("\n")?;
```
ducaale, line 261: Is this newline required for streaming to work properly?

src/printer.rs
```diff
@@ -159,79 +206,107 @@ impl Printer {
...
+            let content_type = get_content_type(&response.headers());
+            // Note that .text() behaves like String::from_utf8_lossy()
+            let text = response.text()?;
+            if text.contains('\0') {
+                self.buffer.print(BINARY_SUPPRESSOR)?;
+                return Ok(());
             }
-        };
+            self.print_body_text(content_type, &text)?;
+            self.buffer.print("\n")?;
```
ducaale, line 271: Do we need this newline?

src/printer.rs
```diff
@@ -34,52 +41,92 @@ impl Printer {
...
+    }
+
+    fn dump(&mut self, reader: &mut impl Read) -> io::Result<()> {
+        copy_largebuf(reader, &mut self.buffer)?;
+        Ok(())
+    }
+
+    fn print_json_text(&mut self, text: &str) -> io::Result<()> {
+        if !self.indent_json {
+            self.print_syntax_text(text, "json")
```
ducaale, line 79: So whether an output should colorized can be determined by either `print_json_text()` or `print_syntax_text`. I wonder if there is a way to restructure this so both indent and coloring are controlled by `print_json_text()`.

src/formatting.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.0.out.flush()
+    }
+}
+
+fn write_style(
+    highlighter: &mut HighlightLines,
+    text: &str,
+    out: &mut impl Write,
+) -> io::Result<()> {
+    // TODO: this text may contain multiple lines, is that ok?
```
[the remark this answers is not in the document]
blyxxyz: If `LineWriter` receives multiple lines in a single write it forwards all of them at once, I believe. And `LineWriter` is only used in streaming mode.

I haven't noticed any problems, so it should be fine, though I haven't dug very deep into `syntect`.

src/buffer.rs
```diff
@@ -34,18 +50,27 @@ impl Buffer {
...
-            Buffer::File(file) => file.write(buf),
-            Buffer::Redirect | Buffer::Stdout => stdout().write(buf),
-            Buffer::Stderr => stderr().write(buf),
+        if self.dirty {
+            return Ok(buf.len());
+        }
+        if self.is_terminal() && buf.contains(&b'\0') {
+            self.print("\x1b[0m")?;
+            self.print(BINARY_SUPPRESSOR)?;
+            self.dirty = true;
```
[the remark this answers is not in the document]
blyxxyz: `print_response_body` doesn't detect it when streaming, this is intended to catch that. That unfortunately does mean it checks twice when it isn't streaming.

I also noticed just now that when HTTPie finds binary data while streaming it immediately exits, which seems sensible if it's not going to show anything.

I think that suggests the streaming binary detection should be moved up a layer, so I'll try to revise that.

src/printer.rs
```diff
@@ -159,79 +206,107 @@ impl Printer {
...
         match get_content_type(&request.headers()) {
             Some(ContentType::Multipart) => {
                 self.buffer.print(MULTIPART_SUPPRESSOR)?;
-                self.buffer.print("\n\n")?;
             }
-            Some(ContentType::Json) => {
-                if let Some(body) = get_body() {
-                    self.print_json(&body)?;
-                    self.buffer.print("\n\n")?;
+            // TODO: Should this print BINARY_SUPPRESSOR?
```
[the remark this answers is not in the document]
blyxxyz: Doing that actually gives me `Error: stream did not contain valid UTF-8`. I'm pretty sure that's wrong, it works fine in HTTPie. So I'll try to fix that.

HTTPie does print `BINARY_SUPPRESSOR` in that scenario, so I think we should print it too.

src/printer.rs
blyxxyz: No, it's just to close off the output nicely, same as the newline printed without streaming.

src/printer.rs
blyxxyz: Yes, otherwise if the output doesn't end in a newline it messes with the shell prompt. Sufficiently advanced shells have a [disgusting trick](https://www.vidarholen.net/contents/blog/?p=878) to fix it, but if you run `xh httpbin.org/json` with the current release in `bash` or `busybox ash` you get something like this at the end:
```
        "title": "Sample Slide Show"
    }
}~ $ 
```

src/printer.rs
blyxxyz: I think that'd just mean duplicating `print_syntax_text` inside `print_json_text`, so it wouldn't really be cleaner.

Right now the logic is that if it doesn't have to do anything specialized it falls back to the generic version.

ducaale: Fair enough. Could you add your last sentence as a comment? I think it would help any future who is skimming the code.

ducaale: Thanks. Can we also add the same comment to `print_json_stream()`?

blyxxyz: Ah, sure.

blyxxyz:
I reworked the `BINARY_SUPPRESSOR` system some more. It feels like there's a lot of awkward duplication and indirection but I can't figure out a better way to structure it.

Non-UTF8 input from stdin is now supported, and a binary suppressor is used for printing the request body.

I reworked `RequestItems` somewhat so it doesn't have to clone all the time.

I changed the URL handling to not use an extra type, not require a hostname, and to add the query parameters earlier (so `RequestItems` doesn't have to be kept around).

src/printer.rs
```diff
@@ -160,78 +264,94 @@ impl Printer {
...
-                while let Some(bytes) = response.chunk().await? {
-                    if is_first_chunk
-                        && matches!(self.buffer, Buffer::Stdout | Buffer::Stderr)
-                        && bytes.contains(&b'\0')
-                    {
-                        self.buffer.print(BINARY_SUPPRESSOR)?;
-                        break;
+                Err(err) if err.kind() == io::ErrorKind::InvalidData => {
+                    if self.color {
+                        self.buffer.print("\x1b[0m")?;
```
ducaale, line 301: Since Terminal output is the only thing that will ever be suppressed, we can include the reset ASCII code at the beginning of the binary suppressor string.

src/printer.rs
```diff
@@ -160,78 +264,94 @@ impl Printer {
...
                     }
-                    is_first_chunk = false;
-
-                    self.buffer.write_all(&bytes)?;
+                    self.buffer.print(BINARY_SUPPRESSOR)?;
                 }
+                Err(err) => return Err(err.into()),
+            }
+        } else {
+            let content_type = get_content_type(&response.headers());
```
ducaale, line 308: Can we move this to the top of the function?

src/buffer.rs
```diff
@@ -1,51 +1,110 @@
...
+
+struct BinaryGuard<'a>(&'a mut Buffer);
+
+impl BinaryGuard<'_> {
+    fn check_dirty(&mut self, buf: &[u8]) -> io::Result<()> {
+        if buf.contains(&b'\0') {
+            Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "Found binary data",
+            ))
```
ducaale, line 89: Have you considered printing the binary suppressor here? (I know it doesn't feel right but I still think it would reduce some of the repetition)

src/printer.rs
```diff
@@ -160,78 +264,94 @@ impl Printer {
...
+            // Note that .text() behaves like String::from_utf8_lossy()
+            let text = response.text()?;
+            if text.contains('\0') {
+                self.buffer.print(BINARY_SUPPRESSOR)?;
+                return Ok(());
```
ducaale, line 313: What would happen if we used `BinaryGuard` for non-streamed responses?

src/printer.rs
blyxxyz: The main reason I do that is that test output becomes *incredibly* confusing if you leave it out. `assert_cmd` just prints the escape code as-is (invisible), so if the test fails because you forgot to put it in you can't see the difference :smile:

But I can also imagine a scenario where someone is using a terminal that doesn't support escape codes and they run `xh` with colors turned off.

src/buffer.rs
blyxxyz: This error is only caught in one place, so it wouldn't be less repetitive. The other places where the suppressor is printed detect it in a different way.

src/printer.rs
ducaale: Okay, so `BinaryGuard` is only necessary when we are streaming a response but for non-streamed responses, we can just scan the whole text?

blyxxyz: Yep, `BinaryGuard` does the same detection incrementally.

ducaale: Can you add that as a comment to `BinaryGuard`?

ducaale (approved):
Approved 🎉🎉🎉🎉🎉

PR #42, opened 2021-02-08
Adds --headers option

sanpii:
This also replaces the `-h` option, like httpie.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -53,6 +53,10 @@ pub struct Cli {
     #[structopt(short = "d", long)]
     pub download: bool,
 
+    /// Print only the response headers, shortcut for --print=h.
+    #[structopt(short = "h", long)]
+    pub headers: bool,
+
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -77,6 +77,10 @@ pub struct Cli {
     #[structopt(short = "S", long)]
     pub stream: bool,
 
+    /// Always stream the response body
```
ducaale, line 80: Can you update this comment?

ducaale (changes requested):
Looks good. I've one comment

src/cli.rs
blyxxyz, line 80: You forgot to change this one :slightly_smiling_face:

sanpii: Oops, done!

ducaale (approved):

ducaale:
Thanks, @sanpii!

PR #44, opened 2021-02-09
Add HTTPS relevant supports: host's SSL certificate verification, client side SSL certificate, etc.

jihchi:
### Description

This PR attempts to implement same/similar functionalities of [HTTPS](https://httpie.io/docs#https) of HTTPie:

- [x] `--verify` To skip the host’s SSL certificate verification.
- [x] `--cert`/`cert-key` To use a client side certificate for the SSL communication.

---

@otaconix have contributed a lot of work on this PR, thank you.


src/cli.rs
```diff
@@ -96,6 +96,10 @@ pub struct Cli {
     /// Optional key-value pairs to be included in the request.
     #[structopt(name = "REQUEST_ITEM")]
     pub request_items: Vec<RequestItem>,
+
+    /// Skip the host's SSL certificate verification
+    #[structopt(long)]
+    pub verify: Option<VerifyHttps>,
```
otaconix, line 102: I was working on adding these exact features, and I see you've chosen a slightly different solution. I thought it might make more sense to have the `--verify` flag have a default value (which is `yes`), instead of making it optional.

To achieve this, I used the `default_value` macro _parameter_ (dunno if that's what they're called, I'm a bit of a Rust noobie), and had an implementation for the `Default` trait, like so:
```rust
impl Default for VerifyHttps {
  fn default() -> Self {
    VerifyHttps::Yes
  }
}
```

I feel this better resembles what httpie does, and it rids the codebase of an `Option` you'll have to deal with.

src/cli.rs
```diff
@@ -96,6 +96,10 @@ pub struct Cli {
     /// Optional key-value pairs to be included in the request.
     #[structopt(name = "REQUEST_ITEM")]
     pub request_items: Vec<RequestItem>,
+
+    /// Skip the host's SSL certificate verification
```
otaconix, line 100: As you've correctly implemented, the flag isn't only used to disable certificate verification, but possibly also to use an alternative CA bundle. It'd be nice if this were documented.

src/cli.rs
```diff
@@ -358,3 +362,25 @@ impl FromStr for RequestItem {
...
+    No,
+    PrivateCerts(String),
+}
+
+impl FromStr for VerifyHttps {
+    type Err = Error;
+    fn from_str(verify: &str) -> Result<VerifyHttps> {
+        match verify.to_lowercase().as_str() {
+            "no" => Ok(VerifyHttps::No),
+            "yes" => Ok(VerifyHttps::Yes),
```
otaconix, line 378: `true` and `false` are also valid values in httpie. I don't know if @ducaale is aiming for an interface that is as close as possible to httpie's, but if he is, it might make sense to replicate that behavior here.

src/cli.rs
otaconix: And seeing how I had the part ready that disables verification, here's the diff I had, which you could use in this PR if you want:

```diff
diff --git a/src/main.rs b/src/main.rs
index cc896ef..91d4534 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,7 @@ mod utils;
 use anyhow::{anyhow, Result};
 use auth::Auth;
 use buffer::Buffer;
-use cli::{AuthType, Cli, Method, Pretty, Print, RequestItem, Theme};
+use cli::{AuthType, Cli, Method, Pretty, Print, RequestItem, Theme, VerifyHttps};
 use download::{download_file, get_file_size};
 use printer::Printer;
 use request_items::{Body, RequestItems};
@@ -68,8 +68,15 @@ async fn main() -> Result<()> {
         true => Policy::limited(args.max_redirects.unwrap_or(10)),
         false => Policy::none(),
     };
+    let disable_certificate_verification = match args.verify {
+        VerifyHttps::No => true,
+        _ => false,
+    };
 
-    let client = Client::builder().redirect(redirect).build()?;
+    let client = Client::builder()
+        .redirect(redirect)
+        .danger_accept_invalid_certs(disable_certificate_verification)
+        .build()?;
     let request = {
         let mut request_builder = client
             .request(method, url.0)
```

**NB:** I haven't yet finished much work on implementing the custom CA bundle functionality, so this would still have to be done.

otaconix (commented):
I hope you don't mind these comments. I'm not the maintainer, but I happen to have started work on exactly this, and found your pull request before I was going to submit mine :)

jihchi:
Hi @otaconix ,

I am happy to see comments and feedback for this PR, appreciate your comments!

I am wondering that are you finished your implementation of these features? I am fine to step down and close this PR since I just get started.

otaconix:
> I am wondering that are you finished your implementation of these features?

Nope, not at all.  You have tests, I don't. I'm pretty much in a research phase when it comes to implementing the custom CA bundle functionality. And finally: I haven't even begun to think about the `--cert` use-case :smile:

To summarize: I don't think I'm much further along than you are, if at all. So don't drop your PR :wink:

src/cli.rs
jihchi: Thanks a lot! I've updated the code.

src/cli.rs
jihchi: I've added `"true"` and `"false"`as valid value.

I can remove those if it is not @ducaale aiming for.

src/cli.rs
jihchi: Added!

src/cli.rs
ducaale: I believe most of the current `ht` users are coming from `httpie`, so it's nice to have a compatible interface to reduce friction. In the future, I am planning to list any divergence between `ht` and `httpie` in the `README.md` file.

otaconix:
To whomever ends up reviewing this: take a look at https://github.com/ctz/rustls/issues/402

Unfortunately, this means we can't easily support encrypted private keys for client certificates. I feel like ctz's rationale for not implementing support for encrypted RSA private keys makes sense. Would this be an acceptable incompatibility with _httpie_?

ducaale:
>Unfortunately, this means we can't easily support encrypted private keys for client certificates. I feel like ctz's rationale for not implementing support for encrypted RSA private keys makes sense. Would this be an acceptable incompatibility with httpie?

I think that is fairly reasonable. We can revisit it in the future if there is a demand for it.

jihchi:
Hi @ducaale , I wonder that are we fine to merge PR and upgrade  `reqwest` later on?

src/main.rs
```diff
@@ -80,6 +82,58 @@ async fn inner_main() -> Result<i32> {
...
+
+                let mut file = File::open(&cert_key).with_context(|| {
+                    format!("Failed to open the cert key file: {}", cert_key.display())
+                })?;
+                file.read_to_end(&mut buffer).with_context(|| {
+                    format!("Failed to read the cert key file: {}", cert_key.display())
+                })?;
+            }
+
+            let identity = reqwest::Identity::from_pem(&buffer)
```
blyxxyz, line 131: This binds us to `rustls`. That's ok, I think, but it's relevant to #49.

tests/cli.rs
```diff
@@ -317,3 +317,127 @@ fn proxy_multiple_valid_proxies() {
...
+    cmd.assert()
+        .stdout(predicates::str::contains("GET / HTTP/1.1"));
+
+    cmd.assert().stderr(indoc! {r#"
+    Error: error sending request for url (https://self-signed.badssl.com/): error trying to connect: invalid certificate: UnknownIssuer
+
+    Caused by:
+        0: error trying to connect: invalid certificate: UnknownIssuer
+        1: invalid certificate: UnknownIssuer
+    "#});
```
blyxxyz, line 338: Every time you call `.assert()` it runs the command again, so it's bests to put it all in a single chain.

`.failure()` makes extra sure the command failed. (There's also `.success()`.)

Checking for the precise phrasing and formatting is probably brittle.

So I'd do:
```suggestion
    get_command()
        .arg("-v")
        .arg("--pretty=format")
        .arg("get")
        .arg("https://self-signed.badssl.com");
        .assert()
        .failure()
        .stdout(predicates::str::contains("GET / HTTP/1.1"));
        .stderr(predicates::str::contains("UnknownIssuer));
```
(and so on for the other tests)

blyxxyz (changes requested):
I have a few nitpicks but this looks good.

@ducaale, do you have a plan for when to do the next release? I think it's possible to incorporate the new `reqwest` feature later and leave it out of this merge, but if we slow down that would make things simpler.

src/main.rs
```diff
@@ -80,6 +82,58 @@ async fn inner_main() -> Result<i32> {
     let mut client = Client::builder().redirect(redirect);
     let mut resume: Option<u64> = None;
 
+    if url.0.scheme() == "https" {
+        client = client.danger_accept_invalid_certs(args.verify == Verify::No);
+
+        if let Verify::CustomCABundle(path) = args.verify {
+            // FIXME: Change the version of reqwest in Cargo.toml to an official released version (> 0.11.0) which should contain this function
+            // See: https://github.com/seanmonstar/reqwest/pull/1150
+            client = client.tls_built_in_root_certs(false);
```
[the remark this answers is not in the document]
jihchi: Fixed in commit 65f0f07b.

src/cli.rs
```diff
@@ -449,6 +464,44 @@ impl FromStr for RequestItem {
...
+impl FromStr for Verify {
+    type Err = Error;
+    fn from_str(verify: &str) -> Result<Verify> {
+        match verify.to_lowercase().as_str() {
+            "no" | "false" => Ok(Verify::No),
+            "yes" | "true" => Ok(Verify::Yes),
+            "" => Err(Error::with_description(
+                &format!("{:?} is not a valid value", verify),
+                ErrorKind::InvalidValue,
+            )),
```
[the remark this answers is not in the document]
jihchi: Right, this is redundant. 

Fixed in commit 4143067.

tests/cli.rs
```diff
@@ -317,3 +317,127 @@ fn proxy_multiple_valid_proxies() {
 // TODO: test implicit download filenames
 // For this we have to pretend the output is a tty
 // This intersects with both #41 and #59
+
+#[test]
+fn verify_default_yes() -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
jihchi: Fixed in commit 8000be0.
---
By the way, may I know what clippy arguments do you supply? I ran `cargo clippy` and didn't get any warnings.

src/main.rs
jihchi: Should we document it on somewhere?

tests/cli.rs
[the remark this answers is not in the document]
blyxxyz: Aha, seems it's new in Rust 1.50.
https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#new-lints-1
https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps

src/main.rs
blyxxyz: Just having this mention show up in that issue is enough, I think. We already had `rustls` unconditionally enabled, so it's not a user-visible change.

tests/cli.rs
blyxxyz: Can you also change the other tests?

blyxxyz: Woops, sorry.

jihchi:
Hi @blyxxyz , I've addressed comments and updated the PR.

Thanks so much for the review.

blyxxyz (approved):

ducaale:
>@ducaale, do you have a plan for when to do the next release? I think it's possible to incorporate the new reqwest feature later and leave it out of this merge, but if we slow down that would make things simpler.

@blyxxyz If I understand correctly, the only downside to merging this with `client.tls_built_in_root_certs(false)` commented out is that `Verify::CustomCABundle(path)` would slightly deviate from HTTPie's current behavior, right? In that case, we could mention that in the docs until a new version of reqwest is released.

blyxxyz:
That's my understanding, yes.

I'm now looking at HTTPie and see that its `--verify` flag is taken from [`merge_environment_settings` in `requests`](https://github.com/psf/requests/blob/master/requests/sessions.py#L701), which can take a `bool` or a `str`. That explains why it's so weird, I'd expect argumentless `--verify` and `--no-verify` flags and a separate `--ca-bundle=...` flag.

And then `requests`'s reason to ignore the default CA bundle if you specify one is that it's reimplementing the `CURL_CA_BUNDLE` environment variable.

We should copy weird parts of HTTPie so people's commands keep working, but all this does mean that its exact behavior is a historical peculiarity and temporarily diverging from it like this isn't so bad.

ducaale:
@jihchi Do you mind modifying the `Verify` doc comment to mention how it differs from HTTPie's _--verify_ flag?

ducaale:
>We should copy weird parts of HTTPie so people's commands keep working, but all this does mean that its exact behavior is a historical peculiarity and temporarily diverging from it like this isn't so bad.

@blyxxyz deviations are bound to happen but as long we are documenting them, it should be okay.

otaconix:
It's possible I'm biased due to having been exposed to the way httpie does it, but it makes sense to me that the default bundle isn't used when I explicitly specify my own: maybe I don't trust the default bundle, and I want to be certain that the host responding to my request is one I/the CA's in the bundle trust.  

I can see this being a factor in a corporate network, where s default bundle isn't going to do much good anyway (which is the reason I contributed to this PR: I want to be able to use `xh` on such a network).

src/main.rs
otaconix: I hadn't noticed this function being behind a feature flag. And now that you mention it: the other `Identity` constructor is behind another feature, which is for `native-tls`. This is a little surprising, as that means there is no way to use client certificates with reqwest in a tls implementation agnostic way.

jihchi:
`reqwest` just released an new version [`0.11.1`](https://github.com/seanmonstar/reqwest/releases/tag/v0.11.1)!
I've put move the part of "don't use built-in root certificates when custom CA bundle is used" back.

src/cli.rs
```diff
@@ -173,6 +175,20 @@ A backslash can be used to escape special characters (e.g. weird\:key=value).
...
+    #[structopt(long, default_value, name = "VERIFY")]
+    pub verify: Verify,
+
+    /// Use a client side certificate for SSL.
+    #[structopt(long, name = "CERT")]
+    pub cert: Option<PathBuf>,
+
+    /// Pass the path to a private key file if the private key is not contained in the cert file.
+    #[structopt(long = "cert-key", name = "CERT_KEY")]
+    pub cert_key: Option<PathBuf>,
```
blyxxyz, line 191: This would make the help easier to skim by pointing out the type of the argument:
```suggestion
    /// Use a client side certificate for SSL.
    #[structopt(long, name = "FILE")]
    pub cert: Option<PathBuf>,

    /// Pass the path to a private key file if the private key is not contained in the cert file.
    #[structopt(long = "cert-key", name = "FILE")]
    pub cert_key: Option<PathBuf>,
```

src/cli.rs
```diff
@@ -173,6 +175,20 @@ A backslash can be used to escape special characters (e.g. weird\:key=value).
     /// Optional key-value pairs to be included in the request.
     #[structopt(skip)]
     pub request_items: Vec<RequestItem>,
+
+    /// Set to "no" (or "false") to skip checking the host's SSL certificate.
+    /// Defaults to "yes" ("true"). You can also pass the path to a CA bundle
+    /// file for private certs.
+    #[structopt(long, default_value, name = "VERIFY")]
+    pub verify: Verify,
```
blyxxyz, line 183: This help text is pretty long. This way it still fits on one line if you run `--help`:
```suggestion
    /// If "no", skip SSL verification. If a file path, use it as a CA bundle.
    ///
    /// Specifying a CA bundle will disable the system's built-in root certificates.
    ///
    /// "false" instead of "no" also works. The default is "yes" ("true").
    #[structopt(long, default_value, hide_default_value = true, name = "VERIFY")]
    pub verify: Verify,
```

src/cli.rs
blyxxyz: My bad, turns out they aren't allowed to have the same name like that :grimacing:

If you use `value_name =` instead of `name =` it works.

jihchi: `value_name` works!

blyxxyz:
I think this is ready now. Thank you for bearing with us!

ducaale (approved):

PR #48, opened 2021-02-09
update README with official Arch Linux package installation instructions

saghm:
While updating my packages this morning, I noticed that `ht-bin` was no longer present in the AUR. At first I thought the name might have been changed, but it turns out that it was moved to the official pacman repository! I figured the installation instructions should be updated, and since the changes were small, I threw them in a PR to make it easier for you (although if you'd prefer to make different changes, feel free to close this PR or push directly to this branch!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,11 @@ Yet another [HTTPie](https://httpie.io/) clone in Rust.
 brew install ht-rust
 ```
 
-### On Arch Linux via AUR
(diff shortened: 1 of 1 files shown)


ducaale:
Thanks, @saghm. There is https://github.com/ducaale/ht/pull/43 which also updates the Arch Linux package installation instructions so I will close this PR in favor of that one.

saghm:
Oops, completely missed that one when I checked to see if there already was one. Sorry about that!

PR #55, opened 2021-02-12
Detect the method and URL arguments more reliably

### LINKED ISSUE #54: query parameter values are not escaped properly
Hi,
I'm passing a query parameter of the form `geometry=POINT(1 2)` to a request : 

```zsh
ht post ".../api/...doStuff?exporttype=legacy&geometry=POINT(1 2)" <<< "[]"
```

please note the request is within double quotes

Using `ht` version `3.5.0`, my app receives `POINT(1` as the value of parameter geometry

blyxxyz:
Determining whether the HTTP method is explicit or implicit is tricky. I've changed it from pre-processing `structopt`'s input to post-processing its output, which is a bit more robust.

Resolves #54.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,3 +1,4 @@
+use std::mem;
 use std::str::FromStr;
 
 use regex::Regex;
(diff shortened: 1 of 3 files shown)


ducaale (approved):
LGTM

PR #56, opened 2021-02-12
Improve automatically generated download filenames

blyxxyz:
- Parse some non-conforming Content-Disposition headers
- Ignore trailing slashes when generating from URL
- Generate from the original URL instead of the URL after redirects
- If there's no extension, pick one using the MIME type


ducaale (approved):
Approved 🎉.

Getting file extension from mime-type was something I've been looking forward to. Thanks!

PR #58, opened 2021-02-13
add macro for lazy regex initialization

ducaale:


blyxxyz:
Nice!

I'm unsure about this usage:
```rust
let re1 = regex!(r"^(.+?)@(.+?);type=(.+?)$");
```
At a glance it looks like the regex isn't static.

But I can't figure out a better way.

ducaale:
Do you mean from the reader's perspective of view? In that case, the macro could be renamed to `lazy_regex!`, what do you think?

blyxxyz:
Yes, from the reader's perspective. But it's probably not worth the extra characters.

I tried to make an alternative form `regex!(RE = r"...")`:
```rust
    ($name:ident = $re:expr) => {
        lazy_static::lazy_static! {
            static ref $name: Regex = Regex::new($re).unwrap();
        }
    };
```
But the surrounding code then can't see `RE`.

ducaale:
Your code does work for me in Rust Playground
```rust
macro_rules! regex {
    ($name:ident = $re:expr) => {
        lazy_static::lazy_static! {
            static ref $name: regex::Regex = regex::Regex::new($re).unwrap();
        }
    };
}

fn main() {
    regex!(RE = r"[^.]+");
    dbg!(RE.captures("hello.world"));
}
```

blyxxyz:
Aha, thank you! It didn't work for me because I kept the old pattern (using it inline is nice), and assignment is an expression so it still had precedence. I just had to switch the order.

How about this?

```rust
#[macro_export]
macro_rules! regex {
    ($name:ident = $($re:expr)+) => {
        lazy_static::lazy_static! {
            static ref $name: regex::Regex = regex::Regex::new(concat!($($re,)+)).unwrap();
        }
    };
    ($($re:expr)+) => {{
        lazy_static::lazy_static! {
            static ref RE: regex::Regex = regex::Regex::new(concat!($($re,)+)).unwrap();
        }
        &RE
    }};
}
```

The `concat!` makes it so you can use multiple strings, e.g. for this long regex I'm working on:

```rust
    regex!(RE_RANGE =
        r"^bytes (?P<first_byte_pos>\d+)-(?P<last_byte_pos>\d+)"
        r"/(?:\*|(?P<complete_length>\d+))$"
    );
```

ducaale:
Looks good to me

PR #59, opened 2021-02-14
Fix --continue, add --check-status

blyxxyz:
- `--continue` now works for resuming downloads. It was incomplete before.
- `--check-status` is supported, and is automatically active for downloads (so you don't download error pages).
- `--auth` and `--download` are now documented.

Internal work:
- Improved regex macro (#58)
- Fixed Clippy warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ FLAGS:
 
 OPTIONS:
     -A, --auth-type <auth-type>              Specify the auth mechanism [possible values: Basic, Bearer]
-    -a, --auth <USER[:PASS]>                 Authenticate as USER. PASS will be prompted if missing
+    -a, --auth <USER[:PASS] | TOKEN>
     -o, --output <output>                    Save output to FILE instead of stdout
         --max-redirects <max-redirects>      Number of redirects to follow, only respected if `follow` is set
     -p, --print <print>                      String specifying what the output should contain
--- a/README.md
+++ b/README.md
@@ -35,19 +35,20 @@ FLAGS:
     -m, --multipart       Similar to --form, but always sends a multipart/form-data request (i.e., even without files)
     -I, --ignore-stdin    Do not attempt to read stdin
     -F, --follow          Do follow redirects
-    -d, --download
+    -d, --download        Download the body to a file instead of printing it
     -h, --headers         Print only the response headers, shortcut for --print=h
     -b, --body            Print only the response body, Shortcut for --print=b
     -c, --continue        Resume an interrupted download
     -v, --verbose         Print the whole request as well as the response
     -q, --quiet           Do not print to stdout or stderr
     -S, --stream          Always stream the response body
+        --check-status    Exit with an error status code if the server replies with an error
         --help            Prints help information
     -V, --version         Prints version information
 
 OPTIONS:
     -A, --auth-type <auth-type>              Specify the auth mechanism [possible values: Basic, Bearer]
-    -a, --auth <auth>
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -39,6 +39,17 @@ fn get_user_agent() -> &'static str {
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    std::process::exit(inner_main().await?);
```
ducaale, line 42: Seems we would have to live with this until Rust can support [returning an exit code from `main()`](https://stackoverflow.com/a/54174631/5915221). Or we can we could use this [crate](https://docs.rs/quit/1.1.2/quit/) instead.

README.md
```diff
@@ -35,19 +35,20 @@ FLAGS:
...
     -q, --quiet           Do not print to stdout or stderr
     -S, --stream          Always stream the response body
+        --check-status    Exit with an error status code if the server replies with an error
         --help            Prints help information
     -V, --version         Prints version information
 
 OPTIONS:
     -A, --auth-type <auth-type>              Specify the auth mechanism [possible values: Basic, Bearer]
-    -a, --auth <auth>
+    -a, --auth <USER[:PASS]>                 Authenticate as USER. PASS will be prompted if missing
```
ducaale, line 51: I was a bit hesitant to add documentation for _--auth_  since what it takes is determined by the `auth-type` which defaults to `bearer` instead of `basic`.

* when auth-type is bearer (default):
  * token (a shortcut for `authorization:bearer token`)
* when auth-type is basic
  * username:password
  * username (a password prompt will be displayed later)
  * username: for passwordless authentication

blyxxyz: Aha, I missed that part. I wonder if it would make more sense to split that off into a separate flag? HTTPie's only alternative `auth-method` still works with username and password.

src/main.rs
blyxxyz: `quit` looks very nice. Maybe something for after #41? I don't know if it composes with `tokio::main`.

Another option might be to attach an exit code as context to the `anyhow` error. That would work better in cases where we also want to show a message.

ducaale: if it could be done using `anyhow` then that is good. Otherwise, we could wait for https://github.com/ducaale/xh/pull/41 👍

README.md
ducaale: >HTTPie's only alternative auth-method still works with username and password.

Could you clarify a bit more?

src/main.rs
blyxxyz: I looked at it again and I think the current approach is better after all. `--check-status` doesn't always print a message, and the program has to continue even after the exit code is known. So it's not like an error and you don't get much mileage out of `quit`'s stack unwinding.

README.md
blyxxyz: HTTPie has `--auth-type basic` and `--auth-type digest`. But both of them use a username and a password, so `--auth` takes the same kind of argument either way.

`--auth-type bearer` changes how `--auth` works, so maybe it's better to split it off completely into a `--bearer` flag that doesn't interact with `--auth-type`. That would be easier to explain.

You'd have:
- `--auth <USER[:PASS]>`
- `--bearer <TOKEN>`

And no more `--auth-type` (unless e.g. `--auth-type digest` is added).

ducaale: Sounds good. Thanks!

blyxxyz: Great! Probably something for another PR, for now I've just removed the description again.

ducaale (approved):
Approved 🎉

PR #60, opened 2021-02-14
Use httpmock for tests

blyxxyz:
This makes it possible to run integration tests without `--offline`. More tests are needed, but this is a start.

I looked into both [httpmock](https://docs.rs/httpmock/0.5.5/httpmock/) and [mockito](https://docs.rs/mockito/0.29.0/mockito/). A difference is that mockito has its own barebones HTTP implementation while httpmock pulls in hyper and a bunch of other libraries. That bloats up `Cargo.lock` but I figured something fully-featured would be better to test a tool that has to talk to many different kinds of servers.

HTTPie apparently just tests with a local httpbin, since httpbin is also written in Python.


ducaale (approved):
Approved 🎉

PR #62, opened 2021-02-15
Support the --proxy option

otaconix:
The reqwest crate does all the heavy lifting (including using the
http_proxy and https_proxy environment variables when the --proxy option
isn't present).
As noted, it's possible to use different proxies for different
protocols.

The "socks" configuration for the reqwest crate was enabled to allow
users to use a socks proxy.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -95,7 +95,7 @@ pub struct Cli {
     /// <PROTOCOL>:<PROXY_URL>. You can specify proxies for multiple protocols by repeating this option.
     /// Note that when this option is absent, the environment variables `http_proxy` and
     /// `https_proxy` can be used to set proxies to use.
-    #[structopt(long, value_name = "<PROTOCOL>:<PROXY_URL>", number_of_values = 1)]
+    #[structopt(long, value_name = "PROTOCOL:PROXY_URL", number_of_values = 1)]
     pub proxy: Option<Vec<Proxy>>,
 
     /// The default scheme to use if not specified in the URL.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -97,13 +97,18 @@ pub struct Cli {
     #[structopt(long)]
     pub check_status: bool,
 
-    /// Proxy to use for a specific protocol. The value passed to this option should take the form of
-    /// <PROTOCOL>:<PROXY_URL>. You can specify proxies for multiple protocols by repeating this option.
-    /// For example, if you want to use a HTTP proxy for HTTPS URLs: `--proxy
-    /// https:http://proxy.host:8080`. If your proxy requires credentials, put them in the URL,
-    /// like so: `--proxy http:socks5://user:password@proxy.host:8000`.
-    /// Note that when this option is absent, the environment variables `http_proxy` and
-    /// `https_proxy` can be used to set proxies to use.
+    /// Proxy to use for a specific protocol. The value passed to this option should take the form
+    /// of <PROTOCOL>:<PROXY_URL>. For example: `--proxy https:http://proxy.host:8080`.
+    ///
+    /// PROTOCOL can be `http`, `https` or `all`.
+    ///
+    /// If your proxy requires credentials, put them in the URL, like so: `--proxy
+    /// http:socks5://user:password@proxy.host:8000`.
+    ///
+    /// You can specify proxies for multiple protocols by repeating this option.
+    ///
+    /// When omitting this option, the environment variables `http_proxy` and `https_proxy` can
+    /// also be used.
     #[structopt(long, value_name = "PROTOCOL:PROXY_URL", number_of_values = 1)]
     pub proxy: Vec<Proxy>,
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,6 +1,8 @@
 use std::str::FromStr;
 
 use regex::Regex;
+use reqwest::Url;
+use std::convert::TryFrom;
 use structopt::clap::AppSettings;
 use structopt::clap::{arg_enum, Error, ErrorKind, Result};
 use structopt::StructOpt;
@@ -88,6 +90,13 @@ pub struct Cli {
     /// Output coloring style.
     #[structopt(short = "s", long = "style", possible_values = &Theme::variants(), case_insensitive = true)]
     pub theme: Option<Theme>,
+
+    /// Proxy to use for a specific protocol. The value passed to this option should take the form of
+    /// <PROTOCOL>:<PROXY_URL>. You can specify proxies for multiple protocols by repeating this option.
+    /// Note that when this option is absent, the environment variables `http_proxy` and
+    /// `https_proxy` can be used to set proxies to use.
+    #[structopt(long, value_name = "<PROTOCOL>:<PROXY_URL>", number_of_values = 1)]
+    pub proxy: Option<Vec<Proxy>>,
 
     /// The default scheme to use if not specified in the URL.
     #[structopt(long = "default-scheme")]
@@ -311,6 +320,47 @@ impl FromStr for Print {
             response_body,
         };
         Ok(p)
+    }
+}
+
+#[derive(Debug)]
+pub enum Proxy {
+    Http(Url),
+    Https(Url),
+    All(Url),
+}
+
+impl FromStr for Proxy {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        let split_arg: Vec<&str> = s.splitn(2, ":").collect();
+        match split_arg[..] {
+            [protocol, url] => {
+                let url = reqwest::Url::try_from(url).map_err(|e| {
+                    Error::with_description(
+                        &format!(
+                            "Invalid proxy URL '{}' for protocol {}: {}",
+                            url, protocol, e
+                        ),
+                        ErrorKind::InvalidValue,
+                    )
+                })?;
+                match protocol.to_lowercase().as_str() {
+                    "http" => Ok(Proxy::Http(url)),
+                    "https" => Ok(Proxy::Https(url)),
+                    "all" => Ok(Proxy::All(url)),
+                    _ => Err(Error::with_description(
+                        &format!("Unknown protocol to set a proxy for: {}", protocol),
+                        ErrorKind::InvalidValue,
+                    )),
+                }
+            }
+            _ => Err(Error::with_description(
+                "The value passed to --proxy should be formatted as <PROTOCOL>:<PROXY_URL>",
+                ErrorKind::InvalidValue,
+            )),
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/cli.rs
```diff
@@ -89,6 +91,13 @@ pub struct Cli {
     #[structopt(short = "s", long = "style", possible_values = &Theme::variants(), case_insensitive = true)]
     pub theme: Option<Theme>,
 
+    /// Proxy to use for a specific protocol. The value passed to this option should take the form of
+    /// <PROTOCOL>:<PROXY_URL>. You can specify proxies for multiple protocols by repeating this option.
+    /// Note that when this option is absent, the environment variables `http_proxy` and
+    /// `https_proxy` can be used to set proxies to use.
```
blyxxyz, line 97: Can you add an example value? HTTPie has `http:http://foo.bar:3128`, but maybe something where the traffic's protocol and the proxy's own protocol don't match would be clearer. (It confused me at first.)

src/cli.rs
```diff
@@ -89,6 +91,13 @@ pub struct Cli {
...
+    #[structopt(long, value_name = "<PROTOCOL>:<PROXY_URL>", number_of_values = 1)]
```
blyxxyz, line 98: This looks less noisy in the help text:
```suggestion
    #[structopt(long, value_name = "PROTOCOL:PROXY_URL", number_of_values = 1)]
```

blyxxyz (changes requested):
Thank you!

Can you target the `develop` branch instead? It also changed that part of the README, so you'll have to fix a (trivial) conflict. (We should include that in a `CONTRIBUTING.md` file.)

src/cli.rs
otaconix: Agreed :)

src/cli.rs
otaconix: What do you think about my latest change? I'm not quite sure if making the output of `--help` much longer is useful, but I agree that the values you should pass to this option aren't very intuitive.

otaconix:
I haven't coded anything to handle multiple proxies being specified for the same URL. I'm thinking maybe that should result in an error, but looking at the behavior _httpie_ displays, maybe it'd be better to use the last provided proxy for some protocol?

src/cli.rs
blyxxyz: I think the amount of information is good. It helps to reorder it slightly and insert blank lines, so everything most people need to know is in the first or second section. After revising it a little I have this:
```
    /// Proxy to use for a specific protocol. The value passed to this option should take the
    /// form of PROTOCOL:PROXY_URL. For example: `--proxy https:http://proxy.host:8080`.
    ///
    /// PROTOCOL can be `http`, `https`, or `all`.
    ///
    /// If your proxy requires credentials, put them in the URL, like so:
    /// `--proxy http:socks5://user:password@proxy.host:8000`.
    ///
    /// You can specify proxies for multiple protocols by repeating this option.
    ///
    /// The environment variables `$http_proxy` and `$https_proxy` can also be used.
```
HTTPie's help text is often pretty wordy.

blyxxyz:
I can think of at least one case where using the last proxy would be helpful: a shell alias that specifies a proxy to use by default, that you sometimes override. The users types `myalias --proxy http:override`, and `xh` sees `xh --proxy http:default --proxy http:override`. In that case `override` should be used.

You'd have to test it, but based on `src/connect.rs` in `reqwest` I think adding the proxies in reverse order should make that work.

src/cli.rs
```diff
@@ -95,6 +97,16 @@ pub struct Cli {
...
 
+    /// Proxy to use for a specific protocol. The value passed to this option should take the form of
+    /// <PROTOCOL>:<PROXY_URL>. You can specify proxies for multiple protocols by repeating this option.
+    /// For example, if you want to use a HTTP proxy for HTTPS URLs: `--proxy
+    /// https:http://proxy.host:8080`. If your proxy requires credentials, put them in the URL,
+    /// like so: `--proxy http:socks5://user:password@proxy.host:8000`.
+    /// Note that when this option is absent, the environment variables `http_proxy` and
+    /// `https_proxy` can be used to set proxies to use.
+    #[structopt(long, value_name = "PROTOCOL:PROXY_URL", number_of_values = 1)]
+    pub proxy: Option<Vec<Proxy>>,
```
blyxxyz, line 108: Could this just be a `Vec<Proxy>`?

src/cli.rs
otaconix: I like it. I had originally used blank lines, but decided to remove them, as _clap_ then decides on a different way to render the long help. Here's what it looks like now when using your suggestion:

```
xh 0.7.0

USAGE:
    xh [FLAGS] [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...

FLAGS:
        --offline         
            Construct HTTP requests without sending them anywhere

    -j, --json            
            (default) Data items from the command line are serialized as a JSON object

    -f, --form            
            Data items from the command line are serialized as form fields

    -m, --multipart       
            Similar to --form, but always sends a multipart/form-data request (i.e., even without files)

    -I, --ignore-stdin    
            Do not attempt to read stdin

    -F, --follow          
            Do follow redirects

    -d, --download        
            Download the body to a file instead of printing it

    -h, --headers         
            Print only the response headers, shortcut for --print=h

    -b, --body            
            Print only the response body, Shortcut for --print=b

    -c, --continue        
            Resume an interrupted download

    -v, --verbose         
            Print the whole request as well as the response

    -q, --quiet           
            Do not print to stdout or stderr

    -S, --stream          
            Always stream the response body

        --check-status    
            Exit with an error status code if the server replies with an error

        --help            
            Prints help information

    -V, --version         
            Prints version information


OPTIONS:
    -A, --auth-type <auth-type>              
            Specify the auth mechanism [possible values: Basic, Bearer]

    -a, --auth <USER[:PASS] | TOKEN>         
            

    -o, --output <output>                    
            Save output to FILE instead of stdout

        --max-redirects <max-redirects>      
            Number of redirects to follow, only respected if `follow` is set

    -p, --print <print>                      
            String specifying what the output should contain

        --pretty <pretty>                    
            Controls output processing [possible values: All, Colors, Format, None]

    -s, --style <theme>                      
            Output coloring style [possible values: Auto, Solarized]

        --proxy <PROTOCOL:PROXY_URL>...      
            Proxy to use for a specific protocol. The value passed to this option should take the form of
            <PROTOCOL>:<PROXY_URL>. For example: `--proxy https:http://proxy.host:8080`.
            
            PROTOCOL can be `http`, `https` or `all`.
            
            If your proxy requires credentials, put them in the URL, like so: `--proxy
            http:socks5://user:password@proxy.host:8000`.
            
            You can specify proxies for multiple protocols by repeating this option.
            
            When omitting this option, the environment variables `http_proxy` and `https_proxy` can also be used.
        --default-scheme <default-scheme>    
            The default scheme to use if not specified in the URL


ARGS:
    <[METHOD] URL>       
            The request URL, preceded by an optional HTTP method

    <REQUEST_ITEM>...    
            Optional key-value pairs to be included in the request
```

I don't mind, per se, but it was a choice I wasn't comfortable making.

src/cli.rs
otaconix: Good catch!

src/cli.rs
blyxxyz: Ah, I hadn't realized that. That might actually be a problem. Especially now that `-h` is short for `--header` instead of printing a shortened `--help`.

For now I guess I'd tend toward something compressed, then:
```rust
    /// Use a proxy for a protocol. For example: `--proxy https:http://proxy.host:8080`.
```
In the long term we should look for a better solution. Maybe we can convince `--help` to print the short help, and move the long help to `--long-help` or a man page (or both?) @ducaale, do you have any ideas?

blyxxyz (approved):
Aside from the documentation thing this looks ready to me

src/cli.rs
ducaale: >In the long term we should look for a better solution. Maybe we can convince --help to print the short help, and move the long help to --long-help or a man page (or both?)

Sounds good to me. I will create an issue for this as a reminder.

ducaale:
Thanks, @otaconix, and @blyxxyz 

I think we could fix the help message problem in a later PR. But for now, I going to merge this.

otaconix:
@ducaale Thanks. Just to be sure: the `README.md` hasn't been updated to reflect the latest (long) output of `--help`, so you'll want to update that if you plan on cutting a release.

ducaale:
This the resulting short-form doc after this PR was merged
```
xh 0.7.0

USAGE:
    xh [FLAGS] [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...

FLAGS:
        --offline         Construct HTTP requests without sending them anywhere
    -j, --json            (default) Data items from the command line are serialized as a JSON object
    -f, --form            Data items from the command line are serialized as form fields
    -m, --multipart       Similar to --form, but always sends a multipart/form-data request (i.e., even without files)
    -I, --ignore-stdin    Do not attempt to read stdin
    -F, --follow          Do follow redirects
    -d, --download        Download the body to a file instead of printing it
    -h, --headers         Print only the response headers, shortcut for --print=h
    -b, --body            Print only the response body, Shortcut for --print=b
    -c, --continue        Resume an interrupted download
    -v, --verbose         Print the whole request as well as the response
    -q, --quiet           Do not print to stdout or stderr
    -S, --stream          Always stream the response body
        --check-status    Exit with an error status code if the server replies with an error
        --help            Prints help information
    -V, --version         Prints version information

OPTIONS:
    -A, --auth-type <auth-type>              Specify the auth mechanism [possible values: Basic, Bearer]
    -a, --auth <USER[:PASS] | TOKEN>         
    -o, --output <output>                    Save output to FILE instead of stdout
        --max-redirects <max-redirects>      Number of redirects to follow, only respected if `follow` is set
    -p, --print <print>                      String specifying what the output should contain
        --pretty <pretty>                    Controls output processing [possible values: All, Colors, Format, None]
    -s, --style <theme>                      Output coloring style [possible values: Auto, Solarized]
        --proxy <PROTOCOL:PROXY_URL>...      Proxy to use for a specific protocol. The value passed to this option
                                             should take the form of <PROTOCOL>:<PROXY_URL>. For example: `--proxy
                                             https:http://proxy.host:8080`
        --default-scheme <default-scheme>    The default scheme to use if not specified in the URL

ARGS:
    <[METHOD] URL>       The request URL, preceded by an optional HTTP method
    <REQUEST_ITEM>...    Optional key-value pairs to be included in the request
```

I am planning to open a PR in which _--help_ outputs the short-form while the long-form version is accessible through man pages.

blyxxyz:
I nearly have a PR ready where `--help` prints the short form and `help` as a subcommand prints the long form.

ducaale:
Oh, thank you very much. I will let you then finish your PR.

PR #64, opened 2021-02-16
Rework help printing, split off --bearer from --auth

blyxxyz:
`--help` now prints short help again, while `help` as a subcommand prints long help (per discussion in #62). My implementation is pretty ugly, but I couldn't find a better way.

`--auth` now only does basic authentication, `--auth-type` is removed, and `--bearer` does bearer authentication (per discussion on #59).

I kept the short help in the README. The long help would be nice to have formatted on a separate page, but I think that's not a release blocker.


otaconix:
I can't quite put my finger on why, but I'm not a fan of referring to environment variables as `$ENV_VAR` in documentation. I guess that, to me, it implies that the `$` is part of the variable name (which happens when you run something like `env -i '$http_proxy=http://localhost:8000'`).

And I swear this isn't (just) because I wrote this, but I found it surprising that passing even a single `--proxy` option would completely override any proxies set through environment variables, even for other protocols. This is the reason the help said something about the environment variables only being used when `--proxy` is omitted altogether. 

Example:

```
env http_proxy=http://env-proxy xh --proxy https:http://env-proxy get http://example.test
```
results in a request going straight to `example.test`. In other words: proxy settings don't fallback to environment variables when `--proxy` is used (and this is due to [_reqwest_ behavior](https://github.com/seanmonstar/reqwest/blob/727903f69aa118c8202640cd05363fe35efcfa69/src/async_impl/client.rs#L578)).

blyxxyz:
> I can't quite put my finger on why, but I'm not a fan of referring to environment variables as `$ENV_VAR` in documentation. I guess that, to me, it implies that the `$` is part of the variable name (which happens when you run something like `env -i '$http_proxy=http://localhost:8000'`).

I get what you mean, it's using shell syntax even though the shell isn't involved. I figured it would be clearer this way but maybe it isn't necessary. I've changed it back.

> And I swear this isn't (just) because I wrote this, but I found it surprising that passing even a single `--proxy` option would completely override any proxies set through environment variables, even for other protocols. This is the reason the help said something about the environment variables only being used when `--proxy` is omitted altogether.

Aha, I missed that nuance. Updated.

blyxxyz:
After some fiddling with [`help2man`](https://www.gnu.org/software/help2man/) and `clap` and `roff` I managed to generate a decent manpage. You can view it without installing it with `man -l doc/xh.1`. `doc/generate.sh` generates it, so we'd have to remember to run that for each release.

As is it's probably only relevant for distro packaging. I'm not sure how to best render it to markdown or HTML, or if that's even a good idea.

doc/xh.1
```diff
@@ -0,0 +1,165 @@
...
+.TH XH "1" "February 2021" "xh 0.7.0" "User Commands"
+.SH NAME
+xh \- Yet another HTTPie clone
+.SH SYNOPSIS
+.B xh
+[\fI\,FLAGS\/\fR] [\fI\,OPTIONS\/\fR] \fI\,<\/\fR[\fI\,METHOD\/\fR] \fI\,URL> \/\fR[\fI\,--\/\fR] [\fI\,REQUEST_ITEM\/\fR]...
+.SH DESCRIPTION
+xh is a friendly and fast tool for sending HTTP requests.
+.PP
+It reimplements as much as possible of HTTPie's excellent design.
```
ducaale, line 11: Nice. I should add this to `README.md` just before the gif demo.

ducaale (approved):
Looks good. I noticed that there no space between subsequent long help blocks but it is a known bug which being tracked here https://github.com/clap-rs/clap/issues/1642

>After some fiddling with help2man and clap and roff I managed to generate a decent manpage. You can view it without installing it with man -l doc/xh.1. doc/generate.sh generates it, so we'd have to remember to run that for each release.

We will also need to modify `.github/workflows/CI.yaml` so that generated manpage is included in releases. Probably in a later PR.

Issue #65, opened 2021-02-17
Support escaping request item seperators


ducaale:
For example, if someone provides `foo\==bar`, it should be parsed as `DataField("foo=", "bar")` and not as `UrlParam("foo", "bar")`. Our current regex could be modified to support that through [negative lookbehind](https://stackoverflow.com/a/9306228/5915221) or via [some other trick](https://stackoverflow.com/a/7902016/5915221) but that comes with some drawbacks:

1. AFAIK, there is no way to discard the escape character. For example, if a regex like this was used `^(.+?)(?<!(\\))(==|:=|=|@|:)((?s).+)$` on `foo\==bar`, it would produce `("foo\=", "=", "bar")` instead of `("foo=", "=", "bar")`.
2. Some of the advanced features needed might not be supported in the `regex` crate we're currently using e.g lookaround.

I've been lately [playing](https://gist.github.com/ducaale/5d3a3df08a39d128eedab362aac34850) with https://github.com/Geal/nom to see how easy it is to support [nested JSON syntax](https://github.com/ducaale/xh/issues/3). Maybe that could help us here?

See https://httpie.io/docs#escaping-rules

ducaale:
Addressed by https://github.com/ducaale/xh/pull/66

PR #66, opened 2021-02-18
Support escaping characters in request items with backslash

blyxxyz:
See #65. A hand-written parser works pretty well.


ducaale (commented):
That was super quick, thanks!

Since `RequestItem` seems to be getting complex, can we move it to its own file?

blyxxyz:
Sure! We already had `request_items.rs`, I figure that's a reasonable place.

ducaale (approved):
LGTM

PR #67, opened 2021-02-19
Make structopt usage more consistent

blyxxyz:
- Always use `value_name` instead of `name` (to avoid conflicts, came up in #44)
- Always use a `value_name` (for more consistent, all-caps output)
- Always use explicit value for `short`
- Do not unnecessarily use explicit value for `long`
- Update README and man page

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,6 @@ cargo install xh
 
 ## Usage
 ```
-xh 0.7.0
-
 USAGE:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -34,8 +34,6 @@ cargo install xh
 
 ## Usage
 ```
-xh 0.7.0
```
ducaale, line 37: One less thing to maintain 👍

ducaale (approved):
LGTM

Issue #57, opened 2021-02-13
Option to convert into curl command (possibly from as well!)


yoursvivek:
I have always preferred `httpie` but as I have to often run same command on remote servers/docker for debugging. Sometimes I also pass these commands to colleagues, who might have liking for different tool (curl being most popular choice). It happens the other way too when I get curl commands.

I'm sure this would resonant with people who access multiple systems throughout the day.

It would be a great addition if this tool could have an option to print equivalent curl command and exit without making request. Even greater if it could do the reverse too.

And thanks for rewriting httpie in rust.

ducaale:
Sounds like an interesting feature to have in `xh`. Meanwhile, you can use https://curl2httpie.online/ to convert between HTTPie syntax and curl.

ducaale:
`xh` version 0.8.0 now supports a _--curl_ flag for converting `xh` commands to curl syntax 🎉. For the reverse, you can use https://curl2httpie.online/.

PR #69, opened 2021-02-21
Add --curl flag to print equivalent curl command

blyxxyz:
For example:
```
$ xh --curl -d -o foobar --continue example.org
curl -L -o foobar -C - 'http://example.org/'
$ xh --curl-long -d -o foobar --continue example.org
curl --location --output foobar --continue-at - 'http://example.org/'
```

See #57.


ducaale (approved):
LGTM

Issue #72, opened 2021-02-22
Replace --default-scheme


blyxxyz:
The only useful use of `--default-scheme` is `--default-scheme=https` (`http` is the default, other schemes don't work). Maybe it should be replaced by a `--https` flag?

We could keep `--default-scheme` as a deprecated [hidden](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.hidden) flag.

ducaale:
Sounds good to me. I wonder if `auth-type` should also have been made hidden instead of removing it.

PR #73, opened 2021-02-23
CLI-related improvements

blyxxyz:
- Replace `--default-scheme` by `--https`. `--default-scheme` is still kept as an undocumented flag. (#72)
- If `xh` is invoked as `xhs`, `https`, or `xhttps`, run as if `--https` was used. That matches HTTPie's `https` command.
  Currently undocumented, we should figure out if we want this and if it should be part of the packaging.
  HTTPie installs both a `http` and `https` executable if you get it with `pip`, but they're just thin wrappers. I expect Cargo only lets you compile two whole separate executables, which would be a bit costly for a convenience feature.
- Bring back `--auth-type` as a hidden flag. If we add more auth types we can document it again.
- Support `NO_COLOR` environment variable to turn colors off by default (https://no-color.org/).
- Make `--json`/`--form`/`--multipart` override each other and force content-type. If you use multiple of those flags, all but the last will be ignored. And if you use them without request items the appropriate headers will still be set. This brings behavior in line with HTTPie.
- Try to detect undeclared JSON response bodies: If the response is javascript or plain text, check if it's JSON.

### DIFF
--- a/src/request_items.rs
+++ b/src/request_items.rs
@@ -5,6 +5,12 @@ use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use reqwest::{blocking::multipart, Method};
 use structopt::clap;
 
+use crate::cli::RequestType;
+
+pub const FORM_CONTENT_TYPE: &str = "application/x-www-form-urlencoded";
+pub const JSON_CONTENT_TYPE: &str = "application/json";
+pub const JSON_ACCEPT: &str = "application/json, */*;q=0.5";
+
 #[derive(Debug, Clone, PartialEq)]
 pub enum RequestItem {
     HttpHeader(String, String),
@@ -127,6 +133,31 @@ pub enum Body {
     Raw(Vec<u8>),
 }
 
+impl Body {
+    pub fn is_empty(&self) -> bool {
+        match self {
+            Body::Json(map) => map.is_empty(),
+            Body::Form(items) => items.is_empty(),
+            Body::Raw(data) => data.is_empty(),
+            // A multipart form without items isn't empty, and we can't read
+            // a body from stdin because it has to match the header, so we
+            // should never consider this "empty"
+            // This is a slight divergence from HTTPie, which will simply
+            // discard stdin if it receives --multipart without request items,
+            // but that behavior is useless so there's no need to match it
+            Body::Multipart(_) => false,
+        }
+    }
+
+    pub fn pick_method(&self) -> Method {
+        if self.is_empty() {
+            Method::GET
+        } else {
+            Method::POST
+        }
+    }
+}
+
 impl RequestItems {
     pub fn new(request_items: Vec<RequestItem>) -> RequestItems {
         RequestItems(request_items)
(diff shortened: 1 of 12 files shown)


tests/cli.rs
```diff
@@ -660,3 +660,122 @@ fn cert_with_key() {
...
+    mock.assert();
+}
+
+#[test]
+fn noninferred_json_output() {
+    let server = MockServer::start();
+    let mock = server.mock(|_when, then| {
+        // Trailing comma makes it invalid JSON, though formatting would still work
+        then.header("content-type", "application/octet-stream")
+            .body(r#"{"":0}"#);
```
ducaale, line 771: I don't see a trailing comma here.

ducaale:
>HTTPie installs both a http and https executable if you get it with pip, but they're just thin wrappers. I expect Cargo only lets you compile two whole separate executables, which would be a bit costly for a convenience feature.

We could just copy and rename the binary, right? (I haven't checked yet if that is possible in Homebrew & Scoop)

src/request_items.rs
```diff
@@ -127,6 +133,31 @@ pub enum Body {
...
+            Body::Json(map) => map.is_empty(),
+            Body::Form(items) => items.is_empty(),
+            Body::Raw(data) => data.is_empty(),
+            // A multipart form without items isn't empty, and we can't read
+            // a body from stdin because it has to match the header, so we
+            // should never consider this "empty"
+            // This is a slight divergence from HTTPie, which will simply
+            // discard stdin if it receives --multipart without request items,
+            // but that behavior is useless so there's no need to match it
+            Body::Multipart(_) => false,
```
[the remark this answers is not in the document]
blyxxyz: Compare to HTTPie:
```sh
$ echo 'hello' | http -v --multipart httpbin.org/post
POST /post HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 38
Content-Type: multipart/form-data; boundary=9a61a9fc681b40b3a6491bc14010402e
Host: httpbin.org
User-Agent: HTTPie/2.5.0-dev

--9a61a9fc681b40b3a6491bc14010402e--


HTTP/1.1 200 OK
[...]
```
It just throws away the stdin, which is worse.

Supplying a raw stdin could maybe work if the user also supplies a `Content-Type` header with a custom boundary, but at that point they don't need the `--multipart` flag any more.

That error message is confusing though, so I'll fix that.

tests/cli.rs
blyxxyz: That's what I get for copy-pasting :stuck_out_tongue:

blyxxyz:
Even a symlink should be enough on Unix/Linux.

No clue about Windows, though. And would it be installed as `xh.exe` there? Should we check for that?

blyxxyz:
Ok, I cross-compiled and ran in Wine and `get_bin_name()` returned `"xh.exe"`. So I've changed it to ignore the extension.

ducaale (approved):
LGTM

PR #74, opened 2021-02-24
Use exit_status crate to generate inner_main function

ducaale:
This PR introduces a [proc-macro](https://github.com/ducaale/exit_status/tree/main) that abstracts away the redirection needed for returning `i32` from `main()`. I believe that a well-structured `main.rs` file could serve as a substitute for [`ARCHITECTURE.md`](https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html).

We can choose to remove it later once we decide to handle errors from `inner_main()` to `main()`.


blyxxyz (approved):
Nice solution!

I don't think using `main.rs` as a map would work for large applications, but at this scale the code is shallow, so it works.

PR #75, opened 2021-02-25
Introduce clippy and fmt in CI

### LINKED ISSUE #71: Include 'cargo fmt' and 'cargo clippy' in the CI pipeline

ducaale:
Resolves #71

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ jobs:
         job:
           # - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf }
           # - { os: ubuntu-latest  , target: x86_64-unknown-linux-musl }
-          - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnueabihf }
+          - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu }
           - { os: macos-latest   , target: x86_64-apple-darwin }
           - { os: windows-latest , target: x86_64-pc-windows-msvc }
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,67 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [ master, develop ]
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


jihchi:
@ducaale would it be better if we can run the test not only on Linux but also macOS and Windows?

ducaale:
It seems `httpmock` depends on OpenSSL which prevents tests from being run for `x86_64-unknown-linux-musl`. I will see what I can do about it.

blyxxyz:
Is using `cross` necessary? Seems like OpenSSL wouldn't be a problem if the builds ran natively.

ducaale:
It seems https://github.com/trishume/syntect is breaking if `cross` is not enabled 😬. Syntect has `fancy-regex` regex engine which is written in pure Rust but it is slower than the default configuration.

ducaale:
It seems that another C library binding is now failing. Maybe we should focus on OpenSSL problems instead?

re-enabling `onig` now...

blyxxyz:
All I see is `Failed to find tool. Is musl-gcc installed?`. Could that be fixed by installing `musl-tools`?

blyxxyz:
Hm, I think the `arm-unknown-linux-gnueabihf` is running on a `x86_64-unknown-linux-gnu` host.

ducaale:
From reading https://github.com/sfackler/rust-openssl/issues/603, it appears https://github.com/clux/muslrust could help us with OpenSSL for musl?

And here is the github action version https://github.com/zhxiaogg/cargo-static-build

ducaale:
Right, I was using https://github.com/sharkdp/bat/blob/master/.github/workflows/CICD.yml as a reference but somehow copied the wrong target 🤦🏻‍♂️.

blyxxyz:
Testing on musl is difficult, but maybe we could do a simple build on musl (without tests, so no OpenSSL)?

I guess Github Actions doesn't support running on ARM? That makes things difficult. But since it's complaining about OpenSSL, and running ARM binaries would be tricky even if we do compile them, maybe we could also do a testless build for that?

I had to do 5bc653c608cfc90da56fc7efec553a11db169577 to get the Windows pipeline to pass. I can look into a better fix for that later.

blyxxyz (approved):
Most of the commits should probably be squashed

Issue #70, opened 2021-02-22
Includes docs/xh.1, README.md, LICENSE.md in the release artifacts


ducaale:
I have looked at how [bat](https://github.com/sharkdp/bat) release artifact is structured ([ripgrep](https://github.com/BurntSushi/ripgrep) should have a similar structure)

```sh
$ wget https://github.com/sharkdp/bat/releases/download/v0.17.1/bat-v0.17.1-x86_64-apple-darwin.tar.gz
$ tar xvzf bat-v0.17.1-x86_64-apple-darwin.tar.gz
$ tree bat-v0.17.1-x86_64-apple-darwin
bat-v0.17.1-x86_64-apple-darwin
├── LICENSE-APACHE
├── LICENSE-MIT
├── README.md
├── autocomplete
│   ├── bat.fish
│   └── bat.zsh
├── bat
└── bat.1
```

The main difference they have with `xh` is:
* The release version is part of the downloaded tar's filename.
* the tar includes `README.md`, `LICENSE.md`, autocompletion, and a manpage file
* The tar contains a folder to avoid polluting the user's filesystem with random files from the release.

Apart from updating `.github\workflows\CI.yaml`, we would need to update places where `xh` is published i.e Homebrew, Scoop, etc.

blyxxyz:
Here's ripgrep's:
```
ripgrep-12.1.1-x86_64-unknown-linux-musl
├── complete
│   ├── _rg
│   ├── rg.bash
│   ├── rg.fish
│   └── _rg.ps1
├── COPYING
├── doc
│   ├── CHANGELOG.md
│   ├── FAQ.md
│   ├── GUIDE.md
│   └── rg.1
├── LICENSE-MIT
├── README.md
├── rg
└── UNLICENSE
```
The only difference is that the documentation files get their own directory, since there are multiple. For the moment we have just a man page, but we could put it in a directory to future-proof it.

[`clap` can generate shell completions.](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.gen_completions_to) I'll take a look to see how good they are, they might need nudging to know which options do and don't take filenames, but they could be nice to include.

blyxxyz:
It might also be useful to [strip debug symbols from the binary](https://github.com/rust-lang/cargo/issues/3483#issuecomment-431209957). That takes the size down from 15MB to 8MB (on Linux at least).

PR #76, opened 2021-02-25
Add shell autocompletion generation

blyxxyz:
See #70.

I've tested these and they seem to work ok, except for the elvish completions (which I've excluded for now).


ducaale (approved):
Approved 🎉🎉

PR #77, opened 2021-02-26
Adapt to_curl tests to Windows

blyxxyz:
This solves a problem with #75.

I'd like to get the input of Windows users on this. The escaping seems designed for cmd.exe, not Powershell, so it could be better to default to Unix-style escaping even on Windows. (But that can also be addressed at a later date.)


ducaale:
Yes, I would vote for defaulting to Unix escaping since it seems to work on PowerShell as well (at least the ones I tested now).

blyxxyz:
Ok, I've made it the default. I've kept the code and tests for cmd.exe escaping in case we add a setting for it later or something.

PR #78, opened 2021-02-27
Use static OpenSSL for tests

blyxxyz:
`curl-rust` turns out to have a feature for building with a bundled statically linked OpenSSL. That solves some of the problems we had in #75.

The integration tests don't work on ARM, possibly because of some QEMU limitation. So I've gated those behind a default feature. The other tests do work.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,12 +9,22 @@ jobs:
   test:
     name: Test
     runs-on: ${{ matrix.job.os }}
-    strategy: 
+    strategy:
       matrix:
         job:
-          - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu }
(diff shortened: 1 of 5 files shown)


ducaale (approved):
LGTM

ducaale:
I just remembered coming across this thread https://github.com/sagebind/isahc/issues/68#issuecomment-595870260 which recommends using `mesalink` feature in `curl` as an alternative solution. What do you think?

blyxxyz:
Aha, that looks even better. Gives us less C code to compile.

blyxxyz:
Apparently it still depends on OpenSSL if you enable the `mesalink` feature: https://github.com/blyxxyz/xh/runs/1995068209?check_suite_focus=true

`static-ssl` works, so let's stick with that.

PR #79, opened 2021-02-27
Include docs/xh.1, README.md, etc in the release artifacts

ducaale:
Also, strips debug symbols from the binary for Linux and macOS. Tested on a [temporary Github project](https://github.com/ducaale/git-deploy-test-1/tree/develop).

Resolves https://github.com/ducaale/xh/issues/70

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -88,7 +88,7 @@ jobs:
           if [ "${{ matrix.job.os }}" = "windows-latest" ]; then
             7z a "$staging.zip" $staging
           elif [ "${{ matrix.job.os }}" = "macos-latest" ]; then
-            zip -r "$staging.zip" $staging
+            gtar czvf "$staging.tar.gz" $staging
           else
             tar czvf "$staging.tar.gz" $staging
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -57,18 +57,42 @@ jobs:
...
+          mkdir -p "$staging"/{doc,completions}
+          cp LICENSE README.md $bin $staging
+          cp CHANGELOG.md doc/xh.1 "$staging"/doc
+          cp completions/* "$staging"/completions
+
+          if [ "${{ matrix.job.os }}" = "windows-latest" ]; then
+            7z a "$staging.zip" $staging
           elif [ "${{ matrix.job.os }}" = "macos-latest" ]; then
-            zip ../../../xh-${{ matrix.job.target }}.zip xh
+            zip -r "$staging.zip" $staging
```
ducaale, line 91: In case anyone wonders why not use the `tar` command for macOS, it is because it was producing [corrupted output](https://github.com/actions/upload-artifact/issues/151).

blyxxyz: GNU tar should be installed by default now, so maybe we can just run `gtar`?

ducaale: It seems `gtar` doesn't produce corrupted artifacts 👍

blyxxyz (approved):
LGTM

PR #81, opened 2021-02-28
Mention xhs in README.md

ducaale:
PRs for supporting `xhs` support in different package managers:
* Scoop: https://github.com/ScoopInstaller/Main/pull/1876
* Homebrew https://github.com/Homebrew/homebrew-core/pull/72178

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,19 @@ Run `xh help` for more detailed information.
 * `:` for adding or removing headers e.g `connection:keep-alive` or `connection:`.
 * `;` for including headers with empty values e.g `header-without-value;`.
 
+## xh and xhs
+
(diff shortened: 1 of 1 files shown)


blyxxyz:
@orhun, could you add an `xhs` symlink to the Arch package?

(I also noticed the fish completions are installed to `/usr/share/fish/completions/` at the moment—I think they're supposed to go in `/usr/share/fish/vendor_completions.d/`, so there's no conflict if they're ever also upstreamed into fish proper)

orhun:
> @orhun, could you add an `xhs` symlink to the Arch package?

Sure, it's available with `0.8.1-1`.

> (I also noticed the fish completions are installed to `/usr/share/fish/completions/` at the moment—I think they're supposed to go in `/usr/share/fish/vendor_completions.d/`, so there's no conflict if they're ever also upstreamed into fish proper)

Good catch!

PR #82, opened 2021-03-01
Content-Type handling improvements

blyxxyz:
- Default to a content type of `application/json` when reading a body from stdin. It already used to do this, but I broke it in #73.
  It might be worth a bugfix release, I'm not sure. 0e1c5a06f296a553078bfc08b9b13894da4846c6 is a minimal fix that could be cherrypicked.
- Highlight Javascript and CSS, since we're bundling the syntax files for them anyway.
- Check if text is actually JSON before formatting it.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -112,25 +112,19 @@ impl Printer {
         }
     }
 
-    fn print_javascript_text(&mut self, text: &str) -> io::Result<()> {
-        if valid_json(text) {
-            self.print_json_text(text, false)
-        } else {
-            self.print_syntax_text(text, "js")
-        }
-    }
-
     fn print_body_text(&mut self, content_type: Option<ContentType>, body: &str) -> io::Result<()> {
         match content_type {
             Some(ContentType::Json) => self.print_json_text(body, true),
             Some(ContentType::Xml) => self.print_syntax_text(body, "xml"),
             Some(ContentType::Html) => self.print_syntax_text(body, "html"),
             Some(ContentType::Css) => self.print_syntax_text(body, "css"),
-            Some(ContentType::JavaScript) => self.print_javascript_text(body),
             // In HTTPie part of this behavior is gated behind the --json flag
             // But it does JSON formatting even without that flag, so doing
             // this check unconditionally is fine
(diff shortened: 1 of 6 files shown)


src/printer.rs
```diff
@@ -103,15 +112,25 @@ impl Printer {
         }
     }
 
+    fn print_javascript_text(&mut self, text: &str) -> io::Result<()> {
+        if valid_json(text) {
+            self.print_json_text(text, false)
+        } else {
+            self.print_syntax_text(text, "js")
+        }
+    }
```
ducaale, line 121: if this is going to be only called from `print_body_text()`, then it is possible to move the `valid_json()` check to when pattern matching the content-type

```rust
 match content_type {
    /* code omitted */
    Some(ContentType::JavaScript) if valid_json(body) => self.print_javascript_text(body),
    Some(ContentType::JavaScript) => self.print_javascript_text(body),
    Some(ContentType::Text) if valid_json(body) => self.print_json_text(body, false),
     _ => self.buffer.print(body),
}
```

src/printer.rs
```diff
@@ -86,11 +86,20 @@ impl Printer {
...
-        } else if self.color {
+            return self.print_syntax_text(text, "json");
+        }
+
+        if check_valid && !valid_json(text) {
+            // JSONXF may mess up the text, e.g. by removing whitespace
+            // This is somewhat common as application/json is the default
+            // content type for requests
+            return self.print_syntax_text(text, "json");
+        }
```
ducaale, line 100: Can we move this check to when pattern-matching the content-type?

ducaale:
>It might be worth a bugfix release, I'm not sure. 0e1c5a0 is a minimal fix that could be cherry-picked.

I don't see a problem in including this whole PR in the next bugfix release.

src/printer.rs
blyxxyz: I thought about that, but this way we can skip the JSON check if formatting isn't necessary.

ducaale (approved):
LGTM

blyxxyz:
@ducaale, are you going to do a crates.io release too?

ducaale:
Oh, that step in the release escaped my mind. Thanks for reminder!

PR #86, opened 2021-03-03
Add --no-FLAG variants of flags

### LINKED ISSUE #84: "--no-" variants of flags
HTTPie lets you negate flags, e.g. `--no-check-status` cancels out `--check-status`. While [`argparse` has optional support for that](https://github.com/python/cpython/blob/3b4b2cf418707c79f96689e401e3c703c0fdd4d2/Lib/argparse.py#L856), HTTPie does it by [manually processing arguments `argparse` didn't recognize](https://github.com/httpie/httpie/blob/56c4ba17942fb6374cfa6f31954f3fab9061a21b/httpie/cli/argparser.py#L259) (plus a few [hardcoded cases](https://github.com/httpie/httpie/blob/56c4ba17942fb6374cfa6f31954f3fab9061a21b/httpie/cli/definition.py#L269)).

`clap` doesn't support this, but there's an issue for it (clap-rs/clap#815). It also has an [escape hatch](https://docs.rs/clap/2.33.3/clap/enum.AppSettings.html#variant.AllowExternalSubcommands) for looking at unknown arguments, but I don't know exactly how powerful it is, and I'm worried it would degrade the usual error message.

A downside of HTTPie's current approach (and probably of the escape hatch approach for `clap`) is that the order of the flags doesn't matter. Normally the last flag counts, so `--form --json` works like `--json` and `--json --form` like `--form`, but `--no-check-status` always overrides `--check-status`, even if it comes first. But that's acceptable, since the typical usage would be to add `--no-check-status` after an alias that includes `--check-status`, and the opposite wouldn't come up because there's little reason to use `--no-check-status` in an alias.

If this is added to `clap` it might take a long time before it's usable in `xh`. It'd land in version 3.0.0, and then after that we might have to wait for `structopt` to support it (or maybe switch to `clap_derive`).

blyxxyz:
Resolves #84.

I found a way of doing this that's fairly low on manual maintenance.

I noticed the man page includes the version number and the month, so `generate.sh` should be run for each release after editing `Cargo.toml`. (Maybe it's time for a checklist?)


ducaale:
>Maybe it's time for a checklist?

I agree 😅. These are the steps I usually do now:
- Update `CHANGELOG.md` (rename unreleased header to the current date, add any missing changes)
- Bump up the version in `Cargo.toml` and run `cargo check` to update `Cargo.lock`
- Run `git push`.
- Add git tag i.e `git tag v0.9.0`
- Push the local tags to remote i.e `git push --tags`  which will start the CI release action.
- publish to crates.io i.e `cargo publish`

ducaale (approved):
LGTM

Issue #52, opened 2021-02-11
.netrc support


dwink:
Would a PR for `.netrc` [support](https://httpie.io/docs#netrc) be accepted? If so I'll take a stab at it this weekend :)

ducaale:
Thanks, @dwink, That would be greatly appreciated.

PR #87, opened 2021-03-04
Add support for getting credentials from .netrc

dwink:
Addresses #52 .

Adds a `--ignore-netrc` cli option to disable the functionality (same as [httpie](https://httpie.io/docs#netrc)).

I'll update with some tests, but wanted to see if the approach overall was good.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -23,8 +23,8 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
 }
 
 fn netrc_path() -> Option<PathBuf> {
-    match env::var("NETRC") {
-        Ok(path) => {
+    match env::var_os("NETRC") {
+        Some(path) => {
             let pth = PathBuf::from(path);
             if pth.exists() {
                 Some(pth)
@@ -32,11 +32,12 @@ fn netrc_path() -> Option<PathBuf> {
                 None
             }
         }
-        Err(_) => {
+        None => {
             if let Some(hd_path) = home_dir() {
                 [".netrc", "_netrc"]
                     .iter()
                     .map(|f| hd_path.join(f))
+                    .inspect(|p| println!("{:?}", p))
                     .find(|p| p.exists())
             } else {
                 None
@@ -93,17 +94,24 @@ mod tests {
 
     #[test]
     fn netrc() {
-        let netrc = "machine example.com\nlogin user\npassword pass";
+        let good_netrc = "machine example.com\nlogin user\npassword pass";
+        let malformed_netrc = "I'm a malformed netrc!";
+        let missing_login = "machine example.com\npassword pass";
+        let missing_pass = "machine example.com\nlogin user\n";
 
         let expected = vec![
             (
                 "example.com",
+                good_netrc,
                 Some(("user".to_string(), Some("pass".to_string()))),
             ),
-            ("example.org", None),
+            ("example.org", good_netrc, None),
+            ("example.com", malformed_netrc, None),
+            ("example.com", missing_login, Some(("".to_string(), Some("pass".to_string())))),
+            ("example.com", missing_pass, Some(("user".to_string(), None))),
         ];
 
-        for (machine, output) in expected {
+        for (machine, netrc, output) in expected {
             assert_eq!(output, auth_from_netrc(machine, netrc));
         }
     }
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -37,7 +37,6 @@ fn netrc_path() -> Option<PathBuf> {
                 [".netrc", "_netrc"]
                     .iter()
                     .map(|f| hd_path.join(f))
-                    .inspect(|p| println!("{:?}", p))
                     .find(|p| p.exists())
             } else {
                 None
@@ -107,8 +106,16 @@ mod tests {
             ),
             ("example.org", good_netrc, None),
             ("example.com", malformed_netrc, None),
-            ("example.com", missing_login, Some(("".to_string(), Some("pass".to_string())))),
-            ("example.com", missing_pass, Some(("user".to_string(), None))),
+            (
+                "example.com",
+                missing_login,
+                Some(("".to_string(), Some("pass".to_string()))),
+            ),
+            (
+                "example.com",
+                missing_pass,
+                Some(("user".to_string(), None)),
+            ),
         ];
 
         for (machine, netrc, output) in expected {
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1,10 +1,11 @@
+use std::env;
 use std::io;
+use std::path::PathBuf;
 
 use crate::regex;
 use dirs::home_dir;
 use netrc_rs::Netrc;
-use std::fs::File;
-use std::io::Read;
+use std::fs;
 
 pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String>)> {
     if let Some(cap) = regex!(r"^([^:]*):$").captures(&auth) {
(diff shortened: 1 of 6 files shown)


src/auth.rs
```diff
@@ -17,6 +21,53 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
     }
 }
 
+pub fn read_netrc() -> Option<String> {
+    let mut netrc_buf = String::new();
+    if let Some(mut hd_path) = home_dir() {
+        hd_path.push(".netrc");
```
blyxxyz, line 27: HTTPie/`requests` [looks at](https://github.com/psf/requests/blob/913880c45a3a8c3bf6b298e9c38709cd95a9c97c/requests/utils.py#L172) `$NETRC` first and considers both `~/.netrc` and `~/_netrc`, maybe we should do that too?

Tests shouldn't look at the real netrc, or some of them will fail if you have one. I guess setting a `$HOME` environment variable in [`get_base_command()`](https://github.com/ducaale/xh/blob/5624e8e1e0ba5da6d38398735e6da7de810da0d9/tests/cli.rs#L16) would do the trick. Tests that need a fake homedir can then override it once more with a temporary directory.

src/auth.rs
```diff
@@ -17,6 +21,53 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
...
+}
+
+pub fn auth_from_netrc(machine: &str, netrc: String) -> Option<(String, Option<String>)> {
+    if let Ok(netrc) = Netrc::parse_borrow(&netrc, false) {
+        let mut auths: Vec<(String, Option<String>)> = netrc
+            .machines
+            .iter()
+            .filter_map(|mach| {
+                if let Some(name) = &mach.name {
+                    if name.ends_with(machine) {
```
blyxxyz, line 45: Why `ends_with`? Shouldn't it be an exact match?

src/main.rs
```diff
@@ -186,6 +186,14 @@ fn main() -> Result<i32> {
         if let Some(auth) = args.auth {
             let (username, password) = parse_auth(auth, url.host_str().unwrap_or("<host>"))?;
             request_builder = request_builder.basic_auth(username, password);
+        } else if !args.ignore_netrc {
+            if let Some(netrc) = read_netrc() {
+                if let Some((username, password)) =
+                    auth_from_netrc(url.host_str().unwrap_or("<host>"), netrc)
```
blyxxyz, line 192: The `"<host>"` up above is for prompting the user, so it's not essential, but here I think we should just skip the netrc if the URL doesn't have a host

src/auth.rs
```diff
@@ -17,6 +21,53 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
...
 
+pub fn read_netrc() -> Option<String> {
+    let mut netrc_buf = String::new();
+    if let Some(mut hd_path) = home_dir() {
+        hd_path.push(".netrc");
+        if let Ok(mut netrc_file) = File::open(hd_path) {
+            if netrc_file.read_to_string(&mut netrc_buf).is_ok() {
+                return Some(netrc_buf);
+            };
+        };
```
blyxxyz, line 32: This can be done in one go with [`std::fs::read_to_string`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html)

blyxxyz (changes requested):
Thank you! I think this is a solid approach.

src/auth.rs
dwink: Typically .netrc works on subdomains too, so `foo.org` works for `*.foo.org` but if your .netrc has `bar.foo.org` then it will only work on `*.bar.foo.org`.

But it might be better to be more explicit about this.

blyxxyz: HTTPie looks for an exact match, FWIW: [`netrc.py`](https://github.com/python/cpython/blob/374ee449331bc95d18c37f5032aaea1448462e58/Lib/netrc.py#L113)

(Which doesn't necessarily mean we should do the same, that implementation barely changed in two decades.)

Seems like `curl` doesn't match subdomains either, which software does?

If we do it, doing it right is a little tricky. `foo.org` should apply to `bar.foo.org`, but not to `barfoo.org`. And if the file lists `bar.foo.org` first and `foo.org` later then we would want to use the `bar.foo.org` configuration rather than `foo.org`.

I think I'd do a check like this:
```rust
match machine.strip_suffix(name) {
    Some(prefix) if prefix.is_empty() || prefix.ends_with('.') => {
        ...
    }
}
```
and then also compare to the length of the domain of the current best match to see if it's more or less specific.

dwink: If both curl and httpie are explicit, then I think I’ll make it explicit as well. It might be my own fuzzy memory that’s at fault here.

src/auth.rs
```diff
@@ -17,6 +21,53 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
...
+                    }
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        if !auths.is_empty() {
+            return auths.pop();
+        }
```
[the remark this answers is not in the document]
dwink: much better this way, thank you!

src/auth.rs
```diff
@@ -17,6 +22,57 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
     }
 }
 
+fn netrc_path() -> Option<PathBuf> {
+    match env::var("NETRC") {
```
blyxxyz, line 26: It's slightly safer to do it like this:
```suggestion
    match env::var_os("NETRC") {
```
That way we won't panic if there's invalid unicode in the variable. You'll get an `OsString` instead of a `String`, but that can be freely converted to `PathBuf`.

src/auth.rs
```diff
@@ -34,4 +90,21 @@ mod tests {
             assert_eq!(output, (user.as_str(), pass.as_deref()));
         }
     }
+
+    #[test]
+    fn netrc() {
+        let netrc = "machine example.com\nlogin user\npassword pass";
```
blyxxyz, line 96: Could you also add tests for files with missing fields and syntax errors?

src/cli.rs
```diff
@@ -60,6 +60,10 @@ pub struct Cli {
     #[structopt(long, value_name = "TOKEN")]
     pub bearer: Option<String>,
 
+    /// Do not use credentials from .netrc
+    #[structopt(long = "ignore-netrc")]
```
blyxxyz, line 64:
```suggestion
    #[structopt(long)]
```

tests/cli.rs
```diff
@@ -530,6 +532,28 @@ fn user_password_auth() {
...
+    let mut netrc = tempfile::NamedTempFile::new().unwrap();
+    writeln!(
+        netrc,
+        "machine {}\nlogin user\npassword pass",
+        server.host()
+    )
+    .unwrap();
+
+    get_command()
+        .env("NETRC", netrc.path().to_str().unwrap())
```
blyxxyz, line 551: `.env()` takes `AsRef<OsStr>`, so I think this would just work:
```suggestion
        .env("NETRC", netrc.path())
```

tests/cli.rs
```diff
@@ -530,6 +532,28 @@ fn user_password_auth() {
...
 }
 
+#[test]
+fn netrc_user_password_auth() {
+    let server = MockServer::start();
+    let mock = server.mock(|when, _then| {
+        when.header("Authorization", "Basic dXNlcjpwYXNz");
+    });
+
+    let mut netrc = tempfile::NamedTempFile::new().unwrap();
```
blyxxyz, line 542: Can you also test a case where it lives in `$HOME/.netrc`?

blyxxyz (changes requested):
Looks good! I'd like a few more tests, but aside from that I only have very small nitpicks.

src/auth.rs
```diff
@@ -17,6 +22,66 @@ pub fn parse_auth(auth: String, host: &str) -> io::Result<(String, Option<String
     }
 }
 
+fn get_home_dir() -> Option<PathBuf> {
+    #[cfg(all(target_os = "win", test))]
```
blyxxyz, line 26: The binary used for the integration tests unfortunately isn't compiled in test mode, that's also why the other `XH_TEST_MODE` variables are compiled unconditionally.

tests/cli.rs
```diff
@@ -530,6 +535,66 @@ fn user_password_auth() {
...
+        get_command()
+            .env("XH_TEST_MODE_WIN_HOME_DIR", homedir.path())
+            .arg(server.base_url())
+            .assert();
+
+        #[cfg(not(target_os = "win"))]
+        get_command()
+            .env("HOME", homedir.path())
+            .arg(server.base_url())
+            .assert();
```
blyxxyz, line 590: Not sure if it'll help, but it should be safe to roll these into one. `HOME` won't hurt Unix tests and `XH_TEST_MODE_WIN_HOME_DIR` won't hurt Windows tests.

blyxxyz (approved):

blyxxyz:
LGTM. Thank you!

PR #89, opened 2021-03-04
Support non-standard HTTP methods

blyxxyz:
Earlier (#69) I introduced a hardcoded list of valid HTTP methods, while HTTPie simply matches against `[a-zA-Z]+`. But weird HTTP methods are a bit more common than I thought, see [varnish](https://varnish-cache.org/docs/3.0/tutorial/purging.html) (httpie/httpie#693). The current system doesn't let you use weird HTTP methods at all, so HTTPie's way seems better.

That unfortunately means `xh localhost` no longer does what you want.

I've made the same change to the highlighting syntax file. I don't think that breaks anything.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::fs::read_dir;
 use std::path::Path;
 
 use syntect::dumps::*;
@@ -7,6 +8,12 @@ use syntect::parsing::SyntaxSetBuilder;
 
 fn main() {
     println!("cargo:rerun-if-changed=assets");
+    for entry in read_dir("assets").unwrap() {
+        println!(
+            "cargo:rerun-if-changed={}",
+            entry.unwrap().path().to_str().unwrap()
+        );
+    }
     let out_dir = env::var_os("OUT_DIR").unwrap();
 
     let mut builder = SyntaxSetBuilder::new();
--- a/build.rs
+++ b/build.rs
@@ -6,6 +6,7 @@ use syntect::highlighting::ThemeSet;
 use syntect::parsing::SyntaxSetBuilder;
 
 fn main() {
+    println!("cargo:rerun-if-changed=assets");
     let out_dir = env::var_os("OUT_DIR").unwrap();
 
     let mut builder = SyntaxSetBuilder::new();
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -421,13 +421,11 @@ impl Cli {
 }
 
 fn parse_method(method: &str) -> Option<Method> {
-    // These are defined as constants on Method
-    const METHODS: &[&str] = &[
-        "GET", "POST", "PUT", "DELETE", "HEAD", "OPTIONS", "CONNECT", "PATCH", "TRACE",
-    ];
-    let method = method.to_ascii_uppercase();
-    if METHODS.contains(&method.as_str()) {
-        Some(method.parse().unwrap())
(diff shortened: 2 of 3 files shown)


src/cli.rs
```diff
@@ -421,13 +421,11 @@ impl Cli {
...
-        "GET", "POST", "PUT", "DELETE", "HEAD", "OPTIONS", "CONNECT", "PATCH", "TRACE",
-    ];
-    let method = method.to_ascii_uppercase();
-    if METHODS.contains(&method.as_str()) {
-        Some(method.parse().unwrap())
+    // This unfortunately matches "localhost"
+    if !method.is_empty() && method.chars().all(|c| c.is_ascii_alphabetic()) {
+        // Method parsing seems to fail if the length is 0 or if there's a null byte
+        // Our checks rule those both out, so .unwrap() is safe
+        Some(method.to_ascii_uppercase().parse().unwrap())
```
ducaale, line 428: Is the uppercasing here necessary🤔?

src/cli.rs
```diff
@@ -767,6 +765,22 @@ mod tests {
         assert!(cli.request_items.is_empty());
     }
 
+    #[test]
+    fn method_edge_cases() {
+        // "localhost" is interpreted as method; this is undesirable, but expected
+        parse(&["localhost"]).unwrap_err();
```
ducaale, line 771: I was about to suggest adding a special case for `localhost`,  but then I remembered that custom domains defined in `/etc/hosts` would also be affected.

build.rs
```diff
@@ -6,6 +6,7 @@ use syntect::highlighting::ThemeSet;
 use syntect::parsing::SyntaxSetBuilder;
 
 fn main() {
+    println!("cargo:rerun-if-changed=assets");
```
ducaale, line 9: This wouldn't allow us to detect modified files inside the directory. Instead, we would need to print a line for each file in the directory.

See https://github.com/rust-lang/cargo/issues/2599

ducaale (commented):
Looks good. I have some minor comments

src/cli.rs
blyxxyz: Yes.

ducaale: It seems that is the case
```
➜  xh git:(blyxxyz-nonstandard-methods) ✗ cargo run -- -v get httpbin.org/get
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/xh -v get httpbin.org/get`
get /get HTTP/1.1
accept: */*
accept-encoding: gzip, deflate
connection: keep-alive
host: httpbin.org
user-agent: xh/0.8.1

HTTP/1.1 400 Bad Request
connection: close
content-length: 122
content-type: text/html
date: Thu, 04 Mar 2021 22:03:17 GMT
server: awselb/2.0

<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
</body>
</html>
```

build.rs
blyxxyz: Ah, yeah.

Better than nothing, though. I expect modifying files will be rare compared to adding and removing them.

ducaale: It could be annoying when someone tries to modify a theme file and nothing happens. We could add a comment or even better, iterate over syntax and theme files inside the assets directory.

ducaale (approved):
Approved 🎉

PR #90, opened 2021-03-04
Add rationale section, similar projects section, and RELEASE-CHECKLIST.md

ducaale:
Rationale section copied from https://github.com/ducaale/xh/issues/83#issuecomment-789851919 with slight modification.

PR #91, opened 2021-03-05
Add missing --no-auth-type, --no-default-scheme

blyxxyz:
#86 was not complete. I only looked at the output of `--help`, and we have hidden options that don't show up there.

I found a way to half-automate the list with [cargo expand](https://crates.io/crates/cargo-expand):
```sh
cargo expand cli | rg -o 'with_name\s*\("([^"]*)"\)' -r '    "--no-$1",' | rg -v 'raw-' | sort
```
If we make sure to do this before release (something for the #90 checklist, along with updating the usage section in the README?) then we don't have to stay on top of it for PRs. (Not ideal, but it's the best I can think of.)

(I sorted the list before replacing it to make sure only `--no-auth-type` and `--no-default-scheme` were added.)


ducaale (approved):
LGTM

ducaale:
>along with updating the usage section in the README?

Not sure if I understood what needs to be added to README

blyxxyz:
The code block with the copy/pasted `--help` output

ducaale:
Do you mean listing all negation flags in the usage section? We already have this line
```
Each option can be reset with a --no-OPTION argument.
```

blyxxyz:
No, I mean making sure it's up to date as a step in the release process. Same with the command I posted.

PR #92, opened 2021-03-06
Replace StructOpt with clap-derive

ducaale:
I found this [file](https://github.com/clap-rs/clap/blob/602c79fecdf9b12071b3040f04e81ffefe1d77f3/clap_up/src/lib.rs) helpful for the migration from StructOpt to clap-derive. Although, now I am realizing that all of this could have been done automatically by running `cargo up dep clap`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,15 +32,6 @@ dependencies = [
 
 [[package]]
 name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


blyxxyz:
I like the cleanup of clap's API.

But is it a good idea to depend on a beta? Are there functional improvements?

ducaale:
>But is it a good idea to depend on a beta? Are there functional improvements?

Maybe not. I just checked several popular rust projects and none of them depend on the beta version of `clap-rs`.

PR #94, opened 2021-03-07
Fix spacing in  generated docs

ducaale:
Also uses the ~undocumented~ [`verbatim_doc_comment`](https://github.com/TeXitoi/structopt/pull/296) attribute to preserve newlines in `REQUEST_ITEM` documentation.

`verbatim_doc_comment` is in the docs. See https://docs.rs/structopt/0.3.21/structopt/index.html#doc-comment-preprocessing-and-structoptverbatim_doc_comment

### Before:
```
      --verify <VERIFY>
            If "no", skip SSL verification. If a file path, use it as a CA bundle.

            Specifying a CA bundle will disable the system's built-in root certificates.

            "false" instead of "no" also works. The default is "yes" ("true").
        --cert <FILE>
            Use a client side certificate for SSL
```

### After:
```
      --verify <VERIFY>
            If "no", skip SSL verification. If a file path, use it as a CA bundle.

            Specifying a CA bundle will disable the system's built-in root certificates.

            "false" instead of "no" also works. The default is "yes" ("true").

        --cert <FILE>
            Use a client side certificate for SSL
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -173,21 +173,17 @@ pub struct Cli {
     raw_method_or_url: String,
 
     /// Optional key-value pairs to be included in the request.
-    #[structopt(
-        value_name = "REQUEST_ITEM",
-        long_help = r"Optional key-value pairs to be included in the request.
-
-- key==value to add a parameter to the URL
-- key=value to add a JSON field (--json) or form field (--form)
-- key:=value to add a complex JSON value (e.g. `numbers:=[1,2,3]`)
-- key@filename to upload a file from filename (with --form)
-- header:value to add a header
-- header: to unset a header
-- header; to add a header with an empty value
-
-A backslash can be used to escape special characters (e.g. weird\:key=value).
-"
-    )]
+    ///
+    ///   - key==value to add a parameter to the URL
+    ///   - key=value to add a JSON field (--json) or form field (--form)
+    ///   - key:=value to add a complex JSON value (e.g. `numbers:=[1,2,3]`)
+    ///   - key@filename to upload a file from filename (with --form)
+    ///   - header:value to add a header
+    ///   - header: to unset a header
+    ///   - header; to add a header with an empty value
+    /// 
+    /// A backslash can be used to escape special characters (e.g. weird\:key=value).
+    #[structopt(value_name = "REQUEST_ITEM", verbatim_doc_comment)]
     raw_rest_args: Vec<String>,
 
     /// The HTTP method, if supplied.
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -153,6 +157,7 @@ pub struct Cli {
     ///
     /// The environment variables `http_proxy` and `https_proxy` can also be used, but
     /// are completely ignored if --proxy is passed.
+    /// {n}{n}{n}
```
ducaale, line 160: By the way, this is not true at the moment, right?
```
/// The environment variables `http_proxy` and `https_proxy` can also be used, but
/// are completely ignored if --proxy is passed.
```

blyxxyz: It is true, but that's part of reqwest's behavior. Calling `.proxy()` stops it from looking at the environment variables later on: https://github.com/seanmonstar/reqwest/blob/a8566383168c0ef06c21f38cbc9213af6ff6db31/src/async_impl/client.rs#L582

blyxxyz:
Nice!

What does `{n}{n}{n}` actually mean?

ducaale:
>What does {n}{n}{n} actually mean?

Tbh, I haven't been able to find what `{n}` means in Rust. https://github.com/TeXitoi/structopt/issues/163 is where I found that it can be used for making newlines in structopt's docstring.

It is repeated 3 times is becuase I found it to be the minimum amount of newlines it takes for a trailing empty line to appear. I also found that something like this would also work
```
    /// Authenticate as USER with PASS. PASS will be prompted if missing.
    ///
    /// Use a trailing colon (i.e. `USER:`) to authenticate with just a username.
    ///
    /// {n}
    #[structopt(short = "a", long, value_name = "USER[:PASS]")]
    pub auth: Option<String>,
```

ducaale:
There is still some bit of inconsistency in doc spacing. For example, these are using one line as a separator
```
xh 0.8.1
xh is a friendly and fast tool for sending HTTP requests.

It reimplements as much as possible of HTTPie's excellent design.
USAGE:
    xh [FLAGS] [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...

FLAGS:
        --offline
            Construct HTTP requests without sending them anywhere
```

while in other cases two lines are being used
```
    -V, --version
            Prints version information


OPTIONS:
    -a, --auth <USER[:PASS]>
```

```
        --cert-key <FILE>
            A private key file to use with --cert.

            Only necessary if the private key is not contained in the cert file.


ARGS:
    <[METHOD] URL>
```

blyxxyz:
It's a clap thing that just inserts a newline, and it'll apparently be removed in clap 3: clap-rs/clap#1810

I think it's fine to use it for now, by the time we upgrade to clap 3 we can see if it's still needed.

Maybe it should be in the "structopt guidelines" list?

PR #96, opened 2021-03-09
Split small syntax definitions into separate file

blyxxyz:
xh can spend a large chunk of its running time just unpacking all the syntax highlighting definitions. The bulk of this is taken up by the HTML, CSS, XML and JavaScript highlighting, but a lot of requests only need the HTTP and JSON highlighting, which are really small.

Packing the HTTP and JSON definitions separately seems to make short small JSON requests about twice as fast, not much slower than with coloring disabled.


ducaale (approved):
LGTM

Issue #99, opened 2021-03-12
xh includes the --verify flag in usage even when it is not used


ducaale:
```
$ xh
error: The following required arguments were not provided:
    <[METHOD] URL>

USAGE:
    xh.exe <[METHOD] URL> --verify <VERIFY>

For more information try --help
```

blyxxyz:
I'm guessing [`default_value`](https://github.com/ducaale/xh/blob/58d19013a5b75c94e7b8510d01bf9cd59e1217e9/src/cli.rs#L218) is the culprit, and it's considered required in some way. Getting rid of that in favor of using `Option<Verify>` might solve it.

ducaale:
I was also wondering if we should use clap's `AppSettings::UnifiedHelpMessage` to simplify the `USAGE` section
#### Before
```
xh 0.9.0
xh is a friendly and fast tool for sending HTTP requests.

It reimplements as much as possible of HTTPie's excellent design.

USAGE:
    xh [FLAGS] [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...
```

#### After
```
xh 0.9.0
xh is a friendly and fast tool for sending HTTP requests.

It reimplements as much as possible of HTTPie's excellent design.

USAGE:
    xh [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...
```

blyxxyz:
That looks better, yeah.

PR #100, opened 2021-03-12
Remove default_value from --verify

ducaale:
Also unifies flags and options

Resolves https://github.com/ducaale/xh/issues/99

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,42 +36,40 @@ cargo install xh
 ## Usage
 ```
 USAGE:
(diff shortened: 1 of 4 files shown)


blyxxyz (approved):

blyxxyz:
Ah, I didn't realize the grouping in the help output would be affected too.

We might want to think about the order while we're at it. Output-related options should be next to each other, `--continue` and `--output` near `--download`, etcetera. Right now it's arbitrary.

Issue #98, opened 2021-03-11
Do not display special characters in dos prompt


anborg:
Expected: Not to display special characters when displaying json 

Actual: Displays special characters.

![image](https://user-images.githubusercontent.com/56970398/110866913-e5139000-8293-11eb-8319-9948e1decd56.png)

No special characters with curl

![image](https://user-images.githubusercontent.com/56970398/110866960-fbb9e700-8293-11eb-82d3-ce4e9bb7cea6.png)

blyxxyz:
Ah, those are the colors. We should look for a way to detect when the terminal doesn't support them.

As a workaround you can disable colors completely by adding `--pretty format`.

Which version of Windows are you using? In newer versions ANSI colors should be turned on by default, I think. Maybe the instructions under "Enabling ANSI colors in older versions of Windows" on https://ss64.com/nt/syntax-ansi.html can fix it?

ducaale:
Colors can also be disabled via the [`no_color`](https://no-color.org/) flag.
```sh
# temporarily set the env variable
set no_color=1   

# permenantly set the env variable
setx no_color=1  
```

You could also use [Windows Terminal](https://www.microsoft.com/en-gb/p/windows-terminal/9n0dx20hk701?rtc=1) which support ANSI colors

<img width="806" alt="Screenshot 2021-03-11 233110" src="https://user-images.githubusercontent.com/16341131/110869219-e9ef3880-82c1-11eb-8438-fea6529b6786.png">

anborg:
Interesting.. 

Windows Terminal, by default opened PoweShell tab which spewed error. Later I opened Dos inside it, no more special characters.  Looks like I need to find a better terminal :)


Power shell ...

![image](https://user-images.githubusercontent.com/56970398/110872972-cd8dd480-829e-11eb-9777-7fa5780524fd.png)


DOS... prompt?

![image](https://user-images.githubusercontent.com/56970398/110873015-dda5b400-829e-11eb-9fb2-c246d441dbdd.png)

blyxxyz:
I think https://github.com/BurntSushi/termcolor might do the trick.

[ripgrep](https://github.com/BurntSushi/ripgrep) uses it, so I'd be interested to know if it prints colors in the outdated DOS prompt that xh messes up. (If it just switches off colors then that would be progress too.)

ducaale:
It seems ripgrep is able to output colors correctly in the old Windows Console

[...]

PR #101, opened 2021-03-12
Enable ansi_support in windows

ducaale:
Resolves https://github.com/ducaale/xh/issues/98

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -522,17 +522,8 @@ arg_enum! {
 }
 
 impl Pretty {
-    #[cfg(not(target_os = "windows"))]
     pub fn color(self) -> bool {
         matches!(self, Pretty::colors | Pretty::all)
-    }
-
-    #[cfg(target_os = "windows")]
-    pub fn color(self) -> bool {
-        match self {
-            Pretty::colors | Pretty::all => ansi_term::enable_ansi_support().is_ok(),
-            _ => false,
-        }
     }
 
     pub fn format(self) -> bool {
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -522,10 +522,19 @@ arg_enum! {
...
 impl Pretty {
+    #[cfg(not(target_os = "windows"))]
     pub fn color(self) -> bool {
         matches!(self, Pretty::colors | Pretty::all)
     }
 
+    #[cfg(target_os = "windows")]
+    pub fn color(self) -> bool {
+        match self {
+            Pretty::colors | Pretty::all => ansi_term::enable_ansi_support().is_ok(),
```
blyxxyz, line 533: Hm, I envisioned `.color()` as a simple utility function without side effects. It might be better to put this in `Printer::new()` instead.

I'm also working on a version that uses `colorterm`. An advantage would be that it looks at `$TERM` (and `$NO_COLOR`, so we don't have to) and doesn't make changes that persist after the program is done. But I'm not sure that it does everything we want and it might end up complicating the code, so it may not be a good idea. This approach should be fine to merge regardless.

ducaale: `enable_ansi_support()` effect doesn't seem to persist between program runs
```
C:\Users\shara\Desktop\projects\xh>target\debug\xh.exe httpbin.org/json
# normal colored output (binary built from this PR that calls enable_ansi_support())

C:\Users\shara\Desktop\projects\xh>xh httpbin.org/json
# garbled output (v0.8.0)
```

However, I believe the flag `enable_ansi_support()` enables [is not supported in older Windows versions](https://stackoverflow.com/a/46044584/5915221) so we may migrate to https://github.com/BurntSushi/termcolor in the future.

PR #105, opened 2021-03-14
Replace ansi_term by termcolor

blyxxyz:
This required a refactoring which happened to simplify the code somewhat.

I think it's good as is, but I'd like to write some tests. Coloring in general is undertested.

See #98.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,15 +40,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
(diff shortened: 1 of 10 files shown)


blyxxyz:
I don't have a system to test it, but I'm worried that this breaks the solarized theme in some Windows terminals because `colorterm` prefers the console API even if ANSI formatting is available. (If so, I wonder if they also can't handle `rg --colors match:fg:0,128,255`.)

We might have to force ANSI formatting if a theme requires it. And perhaps bring back `ansi_term` just for `enable_ansi_support()`.

ducaale:
>If so, I wonder if they also can't handle rg --colors match:fg:0,128,255

Screenshot 2021-03-14 223527

blyxxyz:
Thank you, that's reassuring.

> The only defect I noticed so far is that foreground color is not reset when `xh` exits

Aha, I didn't add `buffer.reset()` everywhere that needs it.

ducaale (approved):
Approved 🎉🎉

PR #108, opened 2021-03-17
Escape backslash in JSON highlighting definition

blyxxyz:
This fixes the highlighting getting confused by `\":` inside a string, as seen when running `xh --offline : 'x=":'`.

### DIFF
--- a/assets/basic/json.sublime-syntax
+++ b/assets/basic/json.sublime-syntax
@@ -16,7 +16,7 @@ contexts:
         - meta_scope: comment.block.jsonkv
         - match: \*/
           pop: true
-    - match: '(")(?i)([^\"]+)(")\s*?:'
+    - match: '(")(?i)([^\\"]+)(")\s*?:'
       comment: Key names
       captures:
         1: keyword.other.name.jsonkv.start


assets/basic/json.sublime-syntax
```diff
@@ -16,7 +16,7 @@ contexts:
         - meta_scope: comment.block.jsonkv
         - match: \*/
           pop: true
-    - match: '(")(?i)([^\"]+)(")\s*?:'
+    - match: '(")(?i)([^\\"]+)(")\s*?:'
```
ducaale, line 19: This would break when keys contain a backslash e.g `xh --offline : "hel\lo=world"`.

#### Before:
Screenshot 2021-03-17 230840

#### After:
Screenshot 2021-03-17 230622



Interestingly, I have noticed that `xh --offline : 'hel"\lo=world'` confuses the highlighter both in develop and and this PR

Screenshot 2021-03-17 231002

blyxxyz: Yep, the highlighting for keys is a bit unreliable. It also fails on empty strings.

It's not a severe problem, and I can't think of a fix I'd be confident in. So I think it's ok as is.

Later on I'd like to see if I can integrate the JSON highlighting with jsonxf. That should make it faster and more reliable.

Another fun case where the key becomes yellow is when highlighting this JSON in streaming mode:
```json
{"foo"
:"bar"}
```
It highlights one line at once, so it doesn't see the colon in time.

ducaale: I agree that color spill is worse than keys that become yellow or blue so I will approve this.

Screenshot 2021-03-18 214226


By the way, when did we start sorting JSON keys?

ducaale (approved):

assets/basic/json.sublime-syntax
blyxxyz: serde_json does that. Seems that we could turn it off by using the `preserve_order` feature.

It only happens for requests, not responses.

ducaale: Although the behavior would be different than HTTPie, I don't think it is a good idea to sort JSON keys.

blyxxyz: Agreed, the order in which you specify them is the most intuitive.

HTTPie doesn't sort them when sending either, just when printing. So keeping insertion order would arguably be closer to HTTPie's behavior.

PR #111, opened 2021-03-19
Add more buffering and make piped output more consistent with HTTPie

blyxxyz:
- Wrapping a `LineWriter` around `File`s gives a nice speedup if the output is formatted or colored, because that avoids doing lots of small writes. That means all output is now line-buffered one way or another.
- Using a `BufWriter` in `print_json_text` gives another speedup. Writing formatted JSON to a file is now ~25 times faster than before.
- Writing to a redirect or a file now doesn't stream unless you use `--stream`, like HTTPie, and it properly decodes the response when it needs to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ xh -d httpbin.org/json -o res.json
 - `rustls` is used instead of the system's TLS library.
 - Headers are sent and printed in lowercase.
 - JSON keys are not sorted.
+- Formatted output is always UTF-8.
 
 ## Similar Projects
 
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,6 +1,6 @@
 use std::{
     fmt,
-    io::{self, stdout, Stdout, Write},
+    io::{self, stdout, LineWriter, Stdout, Write},
     path::Path,
 };
 
@@ -12,7 +12,11 @@ use crate::{
 };
 
 pub enum Buffer {
-    File(Ansi<std::fs::File>),
+    // These are all line-buffered (File explicitly, the others implicitly)
+    // Line buffering gives unsurprising behavior but can still be a lot
+    // faster than no buffering, especially with lots of small writes from
+    // coloring
+    File(Ansi<LineWriter<std::fs::File>>),
     Redirect(Ansi<Stdout>),
     Stdout(StandardStream),
     Stderr(StandardStream),
@@ -35,7 +39,7 @@ impl Buffer {
             Buffer::Stderr(StandardStream::stderr(color_choice))
         } else if let Some(output) = output {
             let file = std::fs::File::create(&output)?;
-            Buffer::File(Ansi::new(file))
+            Buffer::File(Ansi::new(LineWriter::new(file)))
         } else if is_stdout_tty {
             Buffer::Stdout(StandardStream::stdout(color_choice))
         } else {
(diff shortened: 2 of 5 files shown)


ducaale (approved):
Approved 🎉

ducaale:
This is not something related to this PR but I wondering if this was intentional?
```rust
 if check_valid && !valid_json(text) {
     // JSONXF may mess up the text, e.g. by removing whitespace
     // This is somewhat common as application/json is the default
     // content type for requests
     return self.print_syntax_text(text, "json");  // if text is not json, do we need to print it as such?
}
```
---

On a different note, I found `printer.rs` to be getting harder to follow 😔. So I was thinking if something along these lines could help us in taming the complexity:

1. Prefer `match-expression`  to nested `if-statement` when possible.
2. [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). This would apply to how we handle JSON validation inside `printer.rs`.

What do you think?

blyxxyz:
> if text is not json, do we need to print it as such?

It might be almost-JSON that still benefits from highlighting. If the only problem is that it has trailing commas then highlighting makes sense.

HTTPie also highlights even when it can't parse/format.

> Prefer `match-expression` to nested `if-statement` when possible.

I'm not sure. It might depend on taste.

> Parse, don't validate. This would apply to how we handle JSON validation inside `printer.rs`.

That's definitely something to consider. HTTPie does it that way. I think `serde_json` can do formatted output.

The biggest problem is that it would hide quirks like duplicate keys and escaped characters. But maybe dropping duplicate keys can be avoided by deserializing into a custom type, and changing escaped characters is not a big deal.

I've been working on overhauling `jsonxf` with coloring and loose validation, but I think that's a dead end. I'm going to look into building a formatter on top of `serde_json`. There are some limitations I thought it had that it turns out it doesn't.

Performance is also better than I thought. It takes maybe 50% longer than my souped up `jsonxf`, which is more than acceptable.

ducaale:
>I've been working on overhauling jsonxf with coloring and loose validation, but I think that's a dead end. I'm going to look into building a formatter on top of serde_json. There are some limitations I thought it had that it turns out it doesn't.

You can look into [colored_json](https://github.com/ctron/colored_json) crate which `xh` used before moving to syntect + jsonxf. One of the attractive points about jsonxf was that it enabled us to format ndjson without supporting the `--stream` flag.

My `parse, don't validate` comment was more about how `check_valid` bool could be replaced by a struct that wraps the JSON text.

```rust
// don't initialize this struct manually, instead use `.parse::<ValidJson>()`
struct ValidJson<'a>(json: &'a str)

/* TODO: implent TryFrom<&str> trait for ValidJson */
```

This would later allow you to have functions that only accept `ValidJson`.

```rust
fn print_json_text(&mut self, text: ValidJson<&str>) -> io::Result<()> {}
```

>>Prefer match-expression to nested if-statement when possible.
>
>I'm not sure. It might depend on taste.

There are two things I like about match-expressions:
1. It forces you to handle all possible states.
2. You can use match guards to signify special cases

For example, in `print_response_body`, we could ask ourselves which flags we need to handle all their combinations and which we should check in specials cases only
```rust
let should_format = (self.color || self.indent_json) && content_type.is_text();

// Here we chose `self.buffer.is_terminal()` and `self.stream` as the main flags
match (self.buffer.is_terminal(), self.stream) {
    (true, true) => { }
    (true, false) => { }
    (false, true) if should_format => { }
    (false, false) if should_format => { }
    (false, true) => { }
    (false, false) => { }
}

// or maybe `should_format` and `self.stream` are the main flags and `self.buffer.is_terminal` is a special case
```

PR #113, opened 2021-03-21
Preserve the order of JSON keys in requests

ducaale:


blyxxyz:
Can you add a test for this? Maybe a `MockServer` inspecting the request body.

blyxxyz (approved):

PR #114, opened 2021-03-24
Improve progress bar formatting

blyxxyz:
- Prevent mitsuhiko/indicatif#144 in narrow terminals
- Keep bar coloring consistent with other colored output (e.g. don't color it if `$NO_COLOR` is set)


ducaale (approved):
LGTM

PR #115, opened 2021-03-24
Use adaptive window for HTTP/2

### LINKED ISSUE #49: Downloading is slow
Hey, cool tool. :)

I noticed that downloading is pretty slow as compared to `wget`. For instance, on downloading a random large file, using `ht -d` I get 7-13MB/s on repeated tests while using `wget` I get a good 60MB/s (my full connection speed). Any idea?

Sorry I can't be more specific but literally all I do is a multi-run test of a 2-5GB-sized file between `wget` and `ht -d`.

blyxxyz:
Resolves #49.

Earlier I said:

> Enabling `http2_adaptive_window` raises the max speed (in my test, 15 MB/s vs wget's 18 MB/s), but it causes the speed to ramp up much more slowly initially. I'm not sure why. `http2_initial_stream_window_size()` and `http2_initial_connection_window_size()` don't seem to help.

This time I didn't see the slow ramp-up. My best guess is that it happened before because I was testing on a cheap VPS with a pretty fast connection but otherwise very low specs. I think this is an improvement on net.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1734,9 +1734,9 @@ dependencies = [
 
 [[package]]
 name = "reqwest"
-version = "0.11.1"
+version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0460542b551950620a3648c6aa23318ac6b3cd779114bd873209e6e8b5eb1c34"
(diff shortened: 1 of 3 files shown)


ducaale (approved):
LGTM

PR #118, opened 2021-03-29
Support reading DataField and JsonField value from a file

ducaale:
Ticks one more item from #4

### DIFF
--- a/src/request_items.rs
+++ b/src/request_items.rs
@@ -1,4 +1,4 @@
-use std::{fs::File, io, path::Path, str::FromStr};
+use std::{fs::File, io, io::Read, path::Path, str::FromStr};
 
 use anyhow::{anyhow, Result};
 use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
@@ -17,7 +17,9 @@ pub enum RequestItem {
     HttpHeaderToUnset(String),
     UrlParam(String, String),
     DataField(String, String),
+    DataFieldFromFile(String, String),
     JsonField(String, serde_json::Value),
+    JsonFieldFromFile(String, String),
     FormFile(String, String, Option<String>),
 }
 
@@ -25,7 +27,7 @@ impl FromStr for RequestItem {
     type Err = clap::Error;
     fn from_str(request_item: &str) -> clap::Result<RequestItem> {
         const SPECIAL_CHARS: &str = "=@:;\\";
-        const SEPS: &[&str] = &["==", ":=", "=", "@", ":"];
+        const SEPS: &[&str] = &["=@", ":=@", "==", ":=", "=", "@", ":"];
 
         fn unescape(text: &str) -> String {
             let mut out = String::new();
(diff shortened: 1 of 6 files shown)


ducaale:
I need to update `man.roff` to mention the new seperators i.e `=@` and `:=@`

src/request_items.rs
```diff
@@ -210,10 +214,17 @@ impl RequestItems {
                 RequestItem::JsonField(key, value) => {
                     body.insert(key, value);
                 }
+                RequestItem::JsonFieldFromFile(key, value)
+                | RequestItem::DataFieldFromFile(key, value) => {
+                    let mut file = File::open(value)?;
+                    let mut contents = String::new();
+                    file.read_to_string(&mut contents)?;
```
blyxxyz, line 221: You can use [`std::fs::read_to_string`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html) (here and elsewhere)

src/request_items.rs
```diff
@@ -210,10 +214,17 @@ impl RequestItems {
...
+                    body.insert(key, serde_json::from_str(&contents)?);
```
blyxxyz, line 222: I think this is wrong, if it's a `DataFieldFromFile` then it should be inserted as a string, not decoded as JSON

src/request_items.rs
```diff
@@ -228,10 +239,16 @@ impl RequestItems {
...
                 }
                 RequestItem::DataField(key, value) => text_fields.push((key, value)),
+                RequestItem::DataFieldFromFile(key, value) => {
+                    let mut file = File::open(value)?;
+                    let mut contents = String::new();
+                    file.read_to_string(&mut contents)?;
+                    text_fields.push((key, contents));
+                }
                 RequestItem::FormFile(..) => unreachable!(),
                 _ => {}
```
blyxxyz, line 253: I think it's better to not use wildcards for request items anywhere in the codebase any more, so we know we didn't forget anything.

[`to_curl.rs`](https://github.com/ducaale/xh/blob/90dbf65333746845d3c5f21f536e9386ef7e40aa/src/to_curl.rs#L271) is wrong right now. There might be others.

blyxxyz (changes requested):
Looks good overall. Maybe there should be a few integration tests?

src/to_curl.rs
```diff
@@ -253,13 +254,17 @@ pub fn translate(args: Cli) -> Result<Command> {
...
+                RequestItem::JsonField(..) | RequestItem::JsonFieldFromFile(..) => {
                     return Err(anyhow!("JSON values are not supported in multipart fields"));
                 }
                 RequestItem::DataField(key, value) => {
                     cmd.flag("-F", "--form");
                     cmd.push(format!("{}={}", key, value));
                 }
+                RequestItem::DataFieldFromFile(key, value) => {
+                    cmd.flag("-F", "--form");
+                    cmd.push(format!("{}={}", key, fs::read_to_string(value)?));
```
blyxxyz, line 266:
```suggestion
                    cmd.push(format!("{}=<{}", key, value));
```

blyxxyz (approved):
LGTM

PR #119, opened 2021-04-01
Improve the download progress bar

sorairolake:
* Add percentage of progress to progress bar
* Change to print the result when the download is completed


blyxxyz:
Thank you! I think having a percentage would be great.

I'm not sure about the labels. The speed already says `B/s`, so I don't think it needs an extra `DL:` label.

Here's what wget's progress bar looks like, for reference:
```
{filename}   0%[                             ]  15.32M  3.10MB/s    eta 10m 24s
```

blyxxyz (approved):
On second thought, the label does help break up the clutter. So I don't mind it.

ducaale:
I think it would be good to know what the `DL` stands for before merging this change. From what I have seen, multiple spaces are usually used to separate between different metrics and `ETA` is the only label that gets used (see attached screenshots).

## Snatch
Screenshot 2021-04-02 124738

## HTTPie
Screenshot 2021-04-02 125412

## wget 
Screenshot 2021-04-02 125705

## xh (develop)
Screenshot 2021-04-02 130157

## xh (sorairolake:improve-progress-bar)
Screenshot 2021-04-02 124305

ducaale:
Another thing you might notice from prev screenshots is that no program other than `xh` currently shows elapsed time. I was wondering if we could get rid of that and replace it with a percentage. We could then show the total elapsed time after the download finishes. What do you think @blyxxyz?

Screenshot 2021-04-02 131225

blyxxyz:
Hm, that makes sense.

So maybe use wget's bar layout (minus the filename) and add the same message as HTTPie?

We'd have to track the time ourselves for the duration and the speed (can't extract it from the bar), but we can at least reuse [indicatif's file size formatting](https://docs.rs/indicatif/0.15.0/indicatif/struct.HumanBytes.html) as for the starting message.

sorairolake:
I rethinked `DL` label isn't needed, so I don't add it.

src/download.rs
```diff
@@ -242,13 +245,26 @@ pub fn download_file(
...
+    let downloaded_length: u64;
+    if let Some(total_length) = total_length {
+        downloaded_length = total_length - starting_length;
+    } else {
+        downloaded_length = fs::metadata(dest_name).unwrap().len();
+    }
+    eprintln!(
+        "Done. {} in {} ({}/s)",
+        HumanBytes(downloaded_length),
+        FormattedDuration(starting_time.elapsed()),
```
blyxxyz, line 264: Would [`HumanDuration`](https://docs.rs/indicatif/0.15.0/indicatif/struct.HumanDuration.html) instead of `FormattedDuration` be better?

src/download.rs
```diff
@@ -242,13 +245,26 @@ pub fn download_file(
...
             copy_largebuf(&mut response, &mut buffer)?;
         }
     }
 
+    let downloaded_length: u64;
+    if let Some(total_length) = total_length {
+        downloaded_length = total_length - starting_length;
+    } else {
+        downloaded_length = fs::metadata(dest_name).unwrap().len();
+    }
```
blyxxyz, line 260: `fs::metadata()` won't work if we're printing to stdout, because then `dest_name` isn't a real file.

We shouldn't print anything if `quiet` is true (and then we also don't have a progress bar).

So maybe:
```suggestion
    let downloaded_length = pb.position() - starting_length;
```
and then move the whole thing into the `Some(ref pb)` block above?

[`.finish()` updates the position](https://docs.rs/indicatif/0.15.0/src/indicatif/progress.rs.html#631), so maybe it's best to compute `downloaded_length` before calling `.finish()`.

blyxxyz (changes requested):

src/download.rs
sorairolake: This is to print seconds even for more than a minute.

blyxxyz: I think `HumanDuration` is ok because if you want to know the exact time it's still next to the progress bar. `00:00:03` for "3 seconds" is a bit ugly. But you're right that it is unfortunate that 1:20 would just become `1 minute`.

@ducaale, what do you think?

src/download.rs
```diff
@@ -242,13 +245,26 @@ pub fn download_file(
...
+    eprintln!(
+        "Done. {} in {} ({}/s)",
+        HumanBytes(downloaded_length),
+        FormattedDuration(starting_time.elapsed()),
+        HumanBytes(downloaded_length / starting_time.elapsed().as_secs())
```
blyxxyz, line 265: This breaks if downloading takes less than a second:
```
thread 'main' panicked at 'attempt to divide by zero', src/download.rs:265:20
```
Maybe if it takes that little time we shouldn't print the speed at all, because it'll be inaccurate anyway.

src/download.rs
ducaale: It is possible to write a[ custom function](https://www.reddit.com/r/rust/comments/gju305/getting_chronos_duration_to_format_to_hhmmss/fqo9zbb?utm_source=share&utm_medium=web2x&context=3) for outputting human-readable duration. But what should human-readable duration look like?

* `1min 60s` or `1 minute 60 seconds`?
* should it use `and` for separating the minute and seconds? what about hours?

> if you want to know the exact time it's still next to the progress bar

I was hoping that we could replace `elapsed_time` with `percent` so it can look like something like this:

Screenshot 2021-04-02 220047

src/download.rs
sorairolake: Fix to calculate the speed only when the time taken is a second or more.

src/download.rs
sorairolake: @ducaale Changed the time taken to human-readable (Use [Human Time](https://crates.io/crates/humantime)).

blyxxyz (approved):

PR #122, opened 2021-04-02
Support installing xh via a shell script

### LINKED ISSUE #121: [Feature] One-line binary install improvement
I have a plethora of great statically linked binary installs in Linux (Ubuntu). These include `ripgrep`, `exa`, `bat`, `fd` etc. None shines as well as `chezmoi` when it comes to the one-line installer.

I recommend to implement a one-line binary installer to follow that of `chezmoi`. See this documentation: https://github.com/twpayne/chezmoi/blob/master/docs/INSTALL.md#one-line-binary-install

In effect, what it does, is to download the latest statically linked binary for your operating system and architecture into `./bin`
This means I can put myself in whatever dir I want and just call the installer and it will be installed into the bin path of my choosing. Be that `~/bin` or `.local/bin` or whatever I prefer.

ducaale:
Resolves #121

~~I need to test this on different computers hence the draft status of this pull request.~~


README.md
```diff
@@ -8,30 +8,26 @@ as possible of [HTTPie's](https://httpie.io/) excellent design.
...
-### On macOS and Linux via Homebrew
-```
-brew install xh
-```
+### via curl (Linux/macOS)
 
-### On windows via Scoop
 ```
-scoop install xh
+curl -sfL https://raw.githubusercontent.com/ducaale/xh/master/install.sh | sh
```
ducaale, line 14: I need to figure out how to get a nice URL using https://git.io/.

install.sh
```diff
@@ -0,0 +1,34 @@
...
+elif [ "$(uname -s)" = "Linux" ] && [ "$(uname -m)" = "amd64" ]; then
+    target="x86_64-unknown-linux-musl"
+elif [ "$(uname -s)" = "Linux" ] && [ "$(uname -m)" = "arm" ]; then
+    target="arm-unknown-linux-gnueabihf"
+else
+    echo "unsupported OS or architecture"
+    exit 1
+fi
+
+rm -rf xh*
```
blyxxyz, line 18: This is dangerous. Maybe you could use `mktemp` instead?

install.sh
```diff
@@ -0,0 +1,34 @@
...
+
+curl -s https://api.github.com/repos/ducaale/xh/releases/latest \
+| grep -wo "https.*$target.tar.gz" \
+| wget -qi -
```
blyxxyz, line 22: This uses both `wget` and `curl`. I think it's better to only use one of them, and ideal to detect which one is available.

install.sh
```diff
@@ -0,0 +1,34 @@
...
+
+tarball="$(find . -name "xh*$target.tar.gz")"
+tar -xzf $tarball --strip-components=1 
+
+chmod +x xh
+sudo mv xh /usr/local/bin/
```
blyxxyz, line 28: Can we install `xhs` too, as a symlink?

install.sh
```diff
@@ -0,0 +1,34 @@
...
+rm -rf xh*
+
+curl -s https://api.github.com/repos/ducaale/xh/releases/latest \
+| grep -wo "https.*$target.tar.gz" \
+| wget -qi -
+
+tarball="$(find . -name "xh*$target.tar.gz")"
+tar -xzf $tarball --strip-components=1 
+
+chmod +x xh
```
blyxxyz, line 27: This shouldn't be necessary for tarballs. It might be necessary for zip files.

install.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+rm -rf xh*
+
+curl -s https://api.github.com/repos/ducaale/xh/releases/latest \
+| grep -wo "https.*$target.tar.gz" \
+| wget -qi -
+
+tarball="$(find . -name "xh*$target.tar.gz")"
+tar -xzf $tarball --strip-components=1
```
blyxxyz, line 25: Is `--strip-components` available on macOS?

I think `cd`ing into the extracted directory is better. (Becomes easier if `mktemp` is used.)

install.sh
```diff
@@ -0,0 +1,34 @@
...
+    exit 1
+fi
+
+rm -rf xh*
+
+curl -s https://api.github.com/repos/ducaale/xh/releases/latest \
+| grep -wo "https.*$target.tar.gz" \
+| wget -qi -
+
+tarball="$(find . -name "xh*$target.tar.gz")"
```
blyxxyz, line 24: `find` is unnecessary, `tar xzf xh*$target.tar.gz` should work.

install.sh
ducaale: I wonder if we need to check the existence of `curl` since the script is going to be downloaded using curl.

README.md
```diff
@@ -8,30 +8,26 @@ as possible of [HTTPie's](https://httpie.io/) excellent design.
...
-### On macOS and Linux via Homebrew
-```
-brew install xh
-```
+### via curl (Linux/macOS)
 
-### On windows via Scoop
 ```
-scoop install xh
+curl -sfL https://git.io/xh/install.sh | sh
```
ducaale, line 14: Not sure why this doesn't work even when I added `install.sh` to master. The link was created using https://github.com/krrishd/gitio
```
$ gitio 'https://raw.githubusercontent.com/ducaale/xh/master/install.sh' 'xh/install.sh'

Congratulations! You can access your URL at https://git.io/xh/install.sh

```

install.sh
blyxxyz: That would be nice, yes. I don't think it's installed by default on all Linux distros.

If we want to also support `wget`, then `wget -nv -O` is roughly equivalent to `curl -sSL -o`. Both support `-` to write to stdout. So you could have something like this:

```sh
if which curl > /dev/null; then
    fetch='curl -sSL -o'
elif which wget > /dev/null; then
    fetch='wget -q -O'
else
    echo "Can't find curl or wget, can't download package"
    exit 1
fi
```

And then later on:

```sh
$fetch - https://api.github.com/repos/ducaale/xh/releases/latest
```
and
```sh
$fetch xh.tar.gz "$url"
```

install.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo "Could not download tarball"
+    exit 1
+fi
+
+tar xzf xh-*
+sudo mv xh-*/xh /usr/local/bin/
+sudo ln -sf /usr/local/bin/xh /usr/local/bin/xhs
+
+echo "$(xh --version) has been installed successfully"
+echo 'Please ensure that /usr/loca/bin is added to your $PATH'
```
blyxxyz, line 39: At least on Linux it's safe to assume it's in `$PATH`, I think. It's nice to print that that's the place where it's installed, though.

README.md
ducaale: Reverted to using `https://raw.githubusercontent.com/ducaale/xh/master/install.sh`

install.sh
ducaale: I have update the message to be something like this:
```
xh v0.9.2 has been installed to /usr/local/bin
```

install.sh
```diff
@@ -0,0 +1,38 @@
...
+if ! curl -SLO "$url"; then
+    echo "Could not download tarball"
+    exit 1
+fi
+
+tar xzf xh-*
+sudo mv xh-*/xh /usr/local/bin/
+sudo ln -sf /usr/local/bin/xh /usr/local/bin/xhs
+
+echo "xh v$(xh --version | cut -c4-8) has been installed to /usr/local/bin"
```
blyxxyz, line 38: This will break when we hit 0.10.0 and the version grows another character :stuck_out_tongue:

`cut -c4-` should work. But maybe this is simplest:
```suggestion
echo "$(/usr/local/bin/xh --version) has been installed to /usr/local/bin"
```
(the `/usr/local/bin/` makes sure we take the version that's installed, even if another one takes precedence)

blyxxyz (commented):
I think `set -e` should be at the top of the file to make sure the script aborts when an error happens.

The `grep` call should then also become `grep -wo -m1 "https://.*$target.tar.gz" || true` so the proper error message still gets shown.

install.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+echo "Downloading xh..."
+
+if ! $fetch xh.tar.gz "$url"; then
+    echo "Could not download tarball"
+    exit 1
+fi
+
+bindir="${bindir:-/usr/local/bin}"
```
ducaale, line 48: would it be a good idea to create `bindir` if it doesn't exist?

blyxxyz:
Asking the user interactively might be nicer.

Maybe this?

```sh
user_bin=~/.local/bin
case $PATH in
    *:"$user_bin":* | "$user_bin":* | *:"$user_bin")
        default_bin=$user_bin
        ;;
    *)
        default_bin=/usr/local/bin
        ;;
esac

printf "Install location [default: %s]: " "$default_bin"
read -r bindir
bindir=${bindir:-$default_bin}

if ! test -d "$bindir"; then
    echo "Directory $bindir does not exist"
    exit 1
fi
```

If `~/.local/bin` is in `$PATH`, use that as the default, otherwise, use `/usr/local/bin`. But give the user an option to provide an alternative either way.

(I believe `~/.local/bin` is more standard than `~/bin`.)

We should also use `test -w "$bindir"` to determine whether `sudo` is needed.

I feel like it wouldn't be common that we'd have to create the directory, but I might be wrong.

ducaale:
Since `read` doesn't work in non-interactive mode. I had to update the instruction to `bash <(curl -sfL $url)`.

blyxxyz:
Is there a reason you changed it to bash? It's not quite available everywhere, so sticking to sh would be ideal.

`<()` doesn't work in all shells (e.g. fish). I found another solution in Rust's installer:
```sh
read -r bindir < /dev/tty
```

It additionally uses `test -t 1` to make sure the *output* is a TTY, to detect whether the script is run interactively. I don't think we need to bother with that for now.

install.sh
blyxxyz: I think this line is obsolete now

install.sh
```diff
@@ -0,0 +1,79 @@
...
+
+if test -w "$bindir"; then
+    mv xh-*/xh "$bindir/"
+    ln -sf "$bindir/xh" "$bindir/xhs"
+else
+    sudo mv xh-*/xh "$bindir/"
+    sudo ln -sf "$bindir/xh" "$bindir/xhs"
+fi
+
+echo "$($bindir/xh --version) has been installed to $bindir"
```
blyxxyz, line 79:
```suggestion
echo "$("$bindir"/xh --version) has been installed to $bindir"
```
(caught by [shellcheck](https://www.shellcheck.net/), a very useful tool)

install.sh
```diff
@@ -0,0 +1,79 @@
...
+esac
+
+printf "Install location [default: %s]: " "$default_bin"
+read -r bindir < /dev/tty
+bindir=${bindir:-$default_bin}
+
+if ! test -d "$bindir"; then
+    echo "Directory $bindir does not exist"
+    exit 1
+fi
```
blyxxyz, line 65: By this time we've already done a bunch of (potentially slow) downloads, so it might be best to let the user retry.
```suggestion
while ! test -d "$bindir"; then
    echo "Directory $bindir does not exist"
    printf "Install location [default: %s]: " "$default_bin"
    read -r bindir < /dev/tty
    bindir=${bindir:-$default_bin}
fi
```

install.sh
ducaale: Thanks, I will make sure to bookmark this.

install.sh
```diff
@@ -0,0 +1,79 @@
...
+
+echo "Downloading xh..."
+
+if ! $fetch xh.tar.gz "$url"; then
+    echo "Could not download tarball"
+    exit 1
+fi
+
+user_bin=~/.local/bin
```
ducaale, line 48: Should I use `$HOME` here instead of `~` 🤔?

blyxxyz: It should be equivalent. It's expanded before assignment either way.

install.sh
```diff
@@ -0,0 +1,79 @@
...
+    echo "Can't find curl or wget, can't download package"
+    exit 1
+fi
+
+echo "Detected target: $target"
+
+url=$(
+    $fetch - https://api.github.com/repos/ducaale/xh/releases/latest |
+    grep -wo -m1 "https://.*$target.tar.gz" || true
+)
```
ducaale, line 35: On Raspberry Pi 4, this emits the warning `(23) Failed writing body`. I am guessing it is related to this issue https://stackoverflow.com/questions/16703647/why-does-curl-return-error-23-failed-writing-body

ducaale: Fixed

ducaale:
This script has now been tested on Ubuntu (x86), macOS (x86), and Rasberry Pi OS (arm).

@blyxxyz Is there anything else you think should be added? We could support a `non-interactive` mode via a flag but I am not sure if it is worth adding.

install.sh
```diff
@@ -0,0 +1,79 @@
...
+set -e
+
+temp_dir=$(mktemp -d /tmp/xh.XXXXXXXX)
+cd "$temp_dir"
+
+if [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "x86_64" ]; then
+    target="x86_64-apple-darwin"
+elif [ "$(uname -s)" = "Linux" ] && [ "$(uname -m)" = "x86_64" ]; then
+    target="x86_64-unknown-linux-musl"
+elif [ "$(uname -s)" = "Linux" ] && [ "$(uname -m)" = "amd64" ]; then
```
blyxxyz, line 12: Can't hurt to leave this in, but is `uname -m` ever `amd64` on Linux? `x86_64` seems standard.

install.sh
```diff
@@ -0,0 +1,79 @@
...
+    target="x86_64-unknown-linux-musl"
+elif [ "$(uname -s)" = "Linux" ] && ( uname -m | grep -q "arm" ); then
```
blyxxyz, line 14: `uname -m` on my phone (in [Termux](https://termux.com/)) gives `aarch64`, so I don't think this covers it. What is it on a Raspberry Pi?

(There are also [ARM processors without AArch32 support](https://stackoverflow.com/questions/59379848/running-32-bit-arm-binary-on-aarch64-not-working-despite-config-compat) that wouldn't be able to run our binaries, but I don't think it's worth going out of our way to detect those.)

(Termux is a case where `/usr/local/bin` doesn't exist, but we check whether the directory exists either way so that's not a big deal.)

install.sh
```diff
@@ -0,0 +1,79 @@
...
+
+if test -w "$bindir"; then
+    mv xh-*/xh "$bindir/"
+    ln -sf "$bindir/xh" "$bindir/xhs"
+else
+    sudo mv xh-*/xh "$bindir/"
+    sudo ln -sf "$bindir/xh" "$bindir/xhs"
+fi
+
+echo "$("$bindir"/xh --version) has been installed to $bindir"
```
blyxxyz, line 79:
```suggestion
echo "$("$bindir"/xh --version) has been installed to $bindir/xh"
```
I think this sounds a bit better. Or alternatively, `in $bindir`.

blyxxyz (commented):
I think we're nearly there.

install.sh
ducaale: It was something I originally copied from zoxide's [install script](https://github.com/ajeetdsouza/zoxide/blob/main/install.sh#L32) but I agree that it is not needed 👍

install.sh
ducaale: >name -m on my phone (in Termux) gives aarch64, so I don't think this covers it. What is it on a Raspberry Pi?
On Raspberry Pi 4, it was `armv7l`.

`chezmoi`'s [install script](https://github.com/twpayne/chezmoi/blob/master/assets/scripts/install.sh#L122) matches `aarch64` and `armv*` to `arm` so maybe we should copy that?

blyxxyz: Debian has its own arch called `armhf`. I don't know if it shows up in `uname -m`.

Maybe this is best:
```suggestion
elif [ "$(uname -s)" = "Linux" ] && ( uname -m | grep -q -e '^arm' -e '^aarch' ); then
```

tests/cli.rs
```diff
@@ -509,16 +509,18 @@ fn request_binary_detection() {
...
-fn timeout() {
-    get_command()
-        .arg("--timeout=0.01")
-        .arg("https://httpbin.org/delay/0.05")
-        .assert()
-        .failure()
-        .stderr(predicates::str::contains("operation timed out"));
-}
-
+// TODO: figure out a way to avoid non-deterministic test failure
```
blyxxyz, line 512: Why does this fail? Does it manage to connect before the timeout anyway?

reqwest tests this by [connecting to an IP address on the local network with a port that's unlikely to be used](https://github.com/seanmonstar/reqwest/blob/8d3e27966c80a18e4ae71fd40a564d9cd096708a/tests/timeouts.rs#L76). Not airtight, but maybe that would be a better approach?

blyxxyz (changes requested):

tests/cli.rs
ducaale: >Why does this fail? Does it manage to connect before the timeout anyway?

Yes, I think we should update `xh` to use reqwest's `timeout` API instead of `connect_timeout`.

### Timeout behavior in HTTPie
```
$ http --timeout=1 httpbin.org/delay/5

http: error: Request timed out (1.0s).
```

### Timeout behavior in xh with `.connect_timeout()`
```
$ xh httpbin.org/delay/5 --timeout=1

HTTP/1.1 200 OK
connection: keep-alive
content-length: 338
content-type: application/json
date: Wed, 28 Apr 2021 17:35:37 GMT
server: gunicorn/19.9.0

{
    "args": {},
    "data": "",
    "files": {},
    "form": {},
    "headers": {
        "Accept": "*/*",
        "Accept-Encoding": "gzip, br",
        "Host": "httpbin.org",
        "User-Agent": "xh/0.9.2",
        "X-Amzn-Trace-Id": "Root=1-60899ce4-51fea36a5bff4e311d8e0d71"
    },
    "origin": "86.148.212.173",
    "url": "http://httpbin.org/delay/5"
}
```

### Timeout behavior in xh with `.timeout()`
```
$ xh httpbin.org/delay/5 --timeout=1

Error: error sending request for url (http://httpbin.org/delay/5): operation timed out

Caused by:
    operation timed out
```

It would also be nice to match HTTPie error message i.e include the timeout value

blyxxyz: Yep. I suggested `.connect_timeout()` but I was mistaken about what HTTPie did, `.timeout()` gives exactly the right behavior as far as I can tell.

install.sh
```diff
@@ -0,0 +1,78 @@
...
+    printf "Install location [default: %s]: " "$default_bin"
+    read -r bindir < /dev/tty
+    bindir=${bindir:-$default_bin}
+done
+
+tar xzf xh.tar.gz
+
+if test -w "$bindir"; then
+    mv xh-*/xh "$bindir/"
+    ln -sf "$bindir/xh" "$bindir/xhs"
```
blyxxyz, line 72: I feel like we should tell the user that we're doing this, but I can't think of a way that isn't noisy.

blyxxyz (changes requested):

install.sh
ducaale: Would listing each install location in its own line reduce the clutter?
```
Detected target: x86_64-unknown-linux-musl
Downloading xh...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   650  100   650    0     0   9420      0 --:--:-- --:--:-- --:--:--  9420
100 3357k  100 3357k    0     0  5381k      0 --:--:-- --:--:-- --:--:-- 8491k
Install location [default: /home/ducaale/.local/bin]:
xh 0.9.2 has been installed to:
 • /home/ducaale/.local/bin/xh
 • /home/ducaale/.local/bin/xhs
```

blyxxyz: Good idea, I like the way it looks.

install.sh
```diff
@@ -0,0 +1,81 @@
...
+    target="arm-unknown-linux-gnueabihf"
+else
+    echo "Unsupported OS or architecture"
+    exit 1
+fi
+
+fetch()
+{
+    if which curl > /dev/null; then
+        if [ "$#" -eq 2 ]; then curl -SL -o "$1" "$2"; else curl -sSL "$1"; fi
```
blyxxyz, line 19: Nice solution!
`-S` only has an effect if `-s` is used, so:
```suggestion
        if [ "$#" -eq 2 ]; then curl -L -o "$1" "$2"; else curl -sSL "$1"; fi
```

blyxxyz (approved):

PR #123, opened 2021-04-03
Add config file for clippy

sorairolake:
See [clippy's README](https://github.com/rust-lang/rust-clippy#specifying-the-minimum-supported-rust-version) for details.


blyxxyz:
Ah, I didn't know clippy can enforce the MSRV. That's very nice. Seems that we could also put `#![clippy::msrv = "1.45.0"]` in `main.rs`, not sure what's better.

Does the `rustfmt.toml` config add anything? Isn't the edition in `Cargo.toml` enough?

ducaale:
I think clippy's `msrv` is only for disabling lints that don't make sense in old rust versions. see https://github.com/rust-lang/rust-clippy#specifying-the-minimum-supported-rust-version


>Does the rustfmt.toml config add anything? Isn't the edition in Cargo.toml enough?

Yep, from https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#edition

>Rustfmt is able to pick up the edition used by reading the Cargo.toml file if executed through the Cargo's formatting tool `cargo fmt`

sorairolake:
@blyxxyz This applies the same edition as defined in `Cargo.toml` when `rustfmt` is execute directly. The default edition when execute directly is `2015` (see: [config website](https://rust-lang.github.io/rustfmt/)).

ducaale:
It seems that `#![clippy::msrv = "1.45.0"]` requires `custom_inner_attributes` which is not available in stable rust. Personally, I would have preferred if we could limit the number of files we have at the root directory. `Cargo.toml` has `[package.metadata]` that could be used to [store extra information for tools](https://internals.rust-lang.org/t/tool-configs-in-cargo-toml-e-g-rustfmt-clippy-and-alternatives-brainstorming/9055) but that isn't supported by clippy or rustfmt.

ducaale:
> The default edition when execute directly is 2015.

@sorairolake Since `rustfmt` is normally used through cargo, it might not be necessary to have a `rustfmt.toml` file unless we decide to customize `rustfmt`'s default configuration. Can you remove that so we can merge this PR? thanks

sorairolake:
@ducaale I understand. I'll remove `rustfmt.toml`.

ducaale (approved):

PR #124, opened 2021-04-09
Unvendor jsonxf

blyxxyz:


Cargo.toml
```diff
@@ -33,6 +33,7 @@ serde_urlencoded = "0.7.0"
 shell-escape = "0.1.5"
 structopt = "0.3"
 termcolor = "1.1.2"
+jsonxf = "1.0.2"
```
ducaale, line 36: The latest version of `jsonxf` now is v1.1.0

blyxxyz: Ah, I ran `cargo add` and didn't pay close attention to what it added.

Seems that `Cargo.lock` got it right, but we should have it accurate in `Cargo.toml` as well.

ducaale (approved):

PR #125, opened 2021-04-11
Add support for Sessions

ducaale:
This is an attempt to finish the work started by @plombard in #25. I chose to not reuse the commits from the mentioned PR since it would require a bit of work to rebase it on the `develop` branch.


src/session.rs
```diff
@@ -0,0 +1,137 @@
...
+        Ok(path)
+    } else {
+        fs::create_dir_all(path.parent().unwrap())?;
+        fs::File::create(&path)?;
+        Ok(path)
+    }
+}
+
+impl Session {
+    pub fn load_session(host: &str, name_or_path: &str, read_only: bool) -> Result<Self> {
```
blyxxyz, line 51: I think HTTPie's behavior might be a bit better. As far as I can tell:
1. If there's no file, it doesn't create it until it successfully completed the request.
2. An empty file is rejected. A session file should always have valid JSON in it, if it exists.
3. If it didn't exist yet, a file is always created after a successful request, even if `--session-read-only` is used.

Doing it like that means we don't have to handle empty files, and there's no risk of leaving empty files behind. I think it's best not to modify the filesystem at all in `load_session()`.

So the new way would be to try to open the file, and
- If it's not found, make a default session and only try to create it when persisting.
- If there's some other error, bubble it up.
- If decoding fails, bubble it up.

src/session.rs
```diff
@@ -0,0 +1,137 @@
...
+            self.content
+                .cookies
+                .insert(parsed_cookie.name().into(), raw_cookie.into());
+        }
+
+        Ok(())
+    }
+
+    pub fn persist(&self) -> Result<()> {
+        fs::write(&self.path, serde_json::to_string_pretty(&self.content)?)?;
```
blyxxyz, line 134: It's best to put a newline at the end of the file, as is normal for text files.

And increasing the indentation to four spaces would match HTTPie.

src/session.rs
```diff
@@ -0,0 +1,137 @@
...
+use reqwest::header::{HeaderMap, HeaderValue, AUTHORIZATION, COOKIE, SET_COOKIE};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Content {
+    #[serde(rename = "__meta__")]
+    meta: HashMap<String, String>,
+    // TODO: Replace String with Cookie
+    // (need to use Serde's serialize_with and deserialize_with)
+    cookies: HashMap<String, String>,
```
blyxxyz, line 17: Might be covered by your TODO, but HTTPie generates an object with `expires`, `path`, `secure` and `value`, not a string. It crashes when trying to load cookies from this format. (It's otherwise willing to load xh's session files, which is great.)

src/session.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl Session {
+    pub fn load_session(host: &str, name_or_path: &str, read_only: bool) -> Result<Self> {
+        let path = if name_or_path.contains(std::path::is_separator) {
+            ensure_file_exists(PathBuf::from(name_or_path))?
+        } else {
+            let mut path = dirs::config_dir()
+                .unwrap()
+                .join::<PathBuf>(["xh", "sessions", host, name_or_path].iter().collect());
+            path.set_extension("json");
```
blyxxyz, line 58: This turns `foo.bar.baz` into `foo.bar.json`, where HTTPie turns it into `foo.bar.baz.json`. So a dumb append would be better.

src/session.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+impl Session {
+    pub fn load_session(host: &str, name_or_path: &str, read_only: bool) -> Result<Self> {
+        let path = if name_or_path.contains(std::path::is_separator) {
+            ensure_file_exists(PathBuf::from(name_or_path))?
+        } else {
+            let mut path = dirs::config_dir()
+                .unwrap()
```
blyxxyz, line 56: I'd turn this into an error, or at least an `.expect()`. It's possible to trigger it if you unset enough environment variables or if you're running on WASM.

src/main.rs
```diff
@@ -137,6 +139,23 @@ fn main() -> Result<i32> {
...
+                Session::load_session(host, name_or_path, read_only)
+                    .context(format!("couldn't load session {}", name_or_path))?,
+            )
+        }
+        (_, _, _) => None,
+    };
+
+    if let Some(ref s) = session {
+        for (key, value) in s.headers()?.iter() {
+            headers.entry(key).or_insert(value.clone());
```
blyxxyz, line 155: Nitpick: if the key already exists, this clones unnecessarily, so `.or_insert_with()` is more efficient.

blyxxyz (changes requested):
Nice to see this continued!

src/session.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+pub struct Session {
+    pub path: PathBuf,
+    pub read_only: bool,
+    content: Content,
+}
+
+fn ensure_file_exists(path: PathBuf) -> Result<PathBuf> {
```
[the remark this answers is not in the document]
ducaale: This function is now removed as we are no longer creating file when loading the session

src/session.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl Default for Content {
+    fn default() -> Self {
+        Content {
+            meta: Meta::default(),
+            cookies: HashMap::new(),
+            headers: HashMap::new(),
+        }
+    }
+}
```
blyxxyz, line 45: I think `Default` can be derived because all the fields implement it.

src/cli.rs
```diff
@@ -93,6 +93,12 @@ pub struct Cli {
     #[structopt(short = "c", long = "continue", name = "continue")]
     pub resume: bool,
 
+    #[structopt(long)]
+    pub session: Option<String>,
```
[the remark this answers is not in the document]
blyxxyz: On second thought, an `OsString` is more appropriate in case it's a name instead of a path. It can be converted later.

src/session.rs
```diff
@@ -0,0 +1,140 @@
...
+        if let Some(parent_path) = self.path.parent() {
+            fs::create_dir_all(parent_path)?;
+        }
+        let mut session_file = fs::File::create(&self.path)?;
+        if !self.read_only {
+            let formatter = serde_json::ser::PrettyFormatter::with_indent(b"    ");
+            let mut ser = serde_json::Serializer::with_formatter(&mut session_file, formatter);
+            self.content.serialize(&mut ser)?;
+            session_file.write_all(b"\n")?;
+        }
```
blyxxyz, line 137: I don't think this is right. It can leave blank files.

HTTPie creates a file if `read_only` is false *or* if the file doesn't exist yet. Otherwise it does nothing.

blyxxyz (changes requested):

src/session.rs
ducaale: oh, I must have misread your earlier comment. So I am assuming we would want something like this:
```rust
if !args.is_session_read_only {
    if let Some(ref mut s) = session {
        s.persist()
            .context(format!("couldn't persist session {:?}", s.path))?;
    }
}
```

blyxxyz: HTTPie also persists if the session is new, i.e. if the file doesn't exist yet. See [`client.py`](https://github.com/httpie/httpie/blob/fc45bf0fe32711cbbdda2eb5f71ef7cd25b08312/httpie/client.py#L125), [`config.py`](https://github.com/httpie/httpie/blob/fc45bf0fe32711cbbdda2eb5f71ef7cd25b08312/httpie/config.py#L81).

ducaale:
I will need to write some tests for this feature

src/session.rs
```diff
@@ -0,0 +1,168 @@
...
+            .content
+            .cookies
+            .values()
+            .filter(|c| c.has_expired())
+            .map(|c| format!("{}={}", c.name, c.value))
+            .collect::<Vec<_>>()
+            .join("; ");
+        if !cookies.is_empty() {
+            headers.insert(COOKIE, HeaderValue::from_str(&cookies)?);
+        }
```
blyxxyz, line 120: I think this is too simple. HTTPie merges cookies, so if your session file has a `foo=bar` cookie and you have `cookie:baz=qux` as a request item then it sends `Cookie: foo=bar; baz=qux`.

blyxxyz (changes requested):
We should take a close look at what happens with redirects. I assume that right now this happens when using a session with a redirect:
- The final response gets to set cookies, even if it's another domain
- Intermediate responses are ignored entirely
- Requests after the first one don't use session information

I think the first is a vulnerability, or at least a bug. We should check what HTTPie does.

The second seems wrong, not sure how HTTPie handles it.

For the third one, HTTPie does send session information even after a redirect. (I assume it does not after a cross-domain redirect, but I can't find an easy way to test that with httpbin.)

We need to rework redirects in general, redirecting with `-v` doesn't print intermediate requests and responses. I don't think we can let reqwest handle it any more.

I think for this PR we should simply make sure it isn't actually insecure, and then in a later PR (preferably before the next release) we can rework redirects in general. That'll require restructuring `main`, it's not a simple fix.

src/session.rs
```diff
@@ -0,0 +1,193 @@
...
+        ))
+        .unwrap();
+        headers.insert(COOKIE, cookie);
+    } else {
+        headers.insert(COOKIE, cookie);
+    }
+}
+
+pub fn merge_headers(mut headers1: HeaderMap, headers2: HeaderMap) -> HeaderMap {
+    for (key, value) in headers2.iter() {
```
blyxxyz, line 184: Could this be `into_iter()` to avoid the clones?

blyxxyz: (or maybe simply `in headers2 {` would work for that)

src/session.rs
```diff
@@ -0,0 +1,198 @@
...
+        headers.insert(COOKIE, cookie);
+    } else {
+        headers.insert(COOKIE, cookie);
+    }
+}
+
+pub fn merge_headers(mut headers1: HeaderMap, headers2: HeaderMap) -> HeaderMap {
+    let mut current_key = None;
+    for (key, value) in headers2 {
+        current_key = key.or(current_key);
```
blyxxyz, line 186: Huh, why this? Would `let key = key.unwrap();` work?

I'm surprised `key` is an `Option`, in the [docs](https://docs.rs/http/0.2.4/http/header/struct.HeaderMap.html#impl-IntoIterator) it seems that it would be a plain `HeaderName`. Do you know why it does that?

ducaale: I think it is because each key can be associated with multiple values.

From https://docs.rs/http/0.2.4/src/http/header/map.rs.html#1758-1764
```
    /// Creates a consuming iterator, that is, one that moves keys and values
    /// out of the map in arbitrary order. The map cannot be used after calling
    /// this.
    ///
    /// For each yielded item that has `None` provided for the `HeaderName`,
    /// then the associated header name is the same as that of the previously
    /// yielded item. The first yielded item will have `HeaderName` set.
```

Also see https://docs.rs/http/0.2.4/http/header/struct.HeaderMap.html#method.drain

blyxxyz: Ahh, thank you. That makes sense.

src/main.rs
```diff
@@ -253,6 +274,16 @@ fn main() -> Result<i32> {
         if is_redirect && exit_code != 0 {
             eprintln!("\n{}: warning: HTTP {}\n", env!("CARGO_PKG_NAME"), status);
         }
+
+        if let Some(ref mut s) = session {
+            if orig_url.host() == response.url().host() {
```
blyxxyz, line 279: This is not quite enough. We also need to check the scheme and the port. [`remove_sensitive_headers`](https://github.com/seanmonstar/reqwest/blob/18dfac4fe2590f2e3a27e808c72fb26e6d9af98b/src/redirect.rs#L234) is a good example.

HTTPie makes `requests`'s session handling do the hard work. That even handles the authorization header. But I don't think that exists for `reqwest`.

While digging this up I did some more thinking about cookies. Ideally we would capture cookies that were set in intermediate responses, even if the final response is from another host. 

If we enable the `cookies` feature we can use [`cookie_provider()`](https://docs.rs/reqwest/0.11.3/reqwest/blocking/struct.ClientBuilder.html#method.cookie_provider) to set a cookie jar. It doesn't have a way to get the jar out, but it takes an `Arc` so we can keep using it after passing it into the ClientBuilder.

I assume it does safe domain handling. I think that's the sanest option for now. (Unfortunately I think it'd require rewriting some of the other code in this PR.)

Cargo.toml
```diff
@@ -13,6 +13,8 @@ readme = "README.md"
 [dependencies]
 anyhow = "1.0.38"
 atty = "0.2"
+cookie_crate = { version = "0.15", package = "cookie" }
+cookie_store = "0.13"
```
ducaale, line 17: cookie_store depends on reqwest which in turn enables `native-tls` 😕

blyxxyz: Setting `default-features = false` in `cookie_store`'s `Cargo.toml` does the trick. Should I make a PR?

ducaale: Yeah, I think that could work, but we would have to reimplement the `reqwest::cookie::CookieStore` trait for `cookie_str`, right?

>Should I make a PR?

Yes, please

blyxxyz: If we set `default-features = false` for `cookie_store` in our `Cargo.toml` then we have to implement it ourselves, yeah.

You can't implement another crate's trait for another crate's type without a wrapper, but `CookieStoreMutex` is a wrapper anyway so I think we could just copy it over wholesale.

> Yes, please

Done: pfernie/cookie_store#15

For now this works:
```suggestion
cookie_store = { git = "https://github.com/blyxxyz/cookie_store", branch = "no-reqwest-default-features" }
```
It should be replaced before the next release, either by a new upstream release with the fix or by the manual implementation workaround.

ducaale: Thank you very much!

src/request_items.rs
```diff
@@ -177,19 +177,21 @@ impl RequestItems {
             .any(|item| matches!(item, RequestItem::FormFile(..)))
     }
 
-    pub fn headers(&self) -> Result<(HeaderMap<HeaderValue>, Vec<HeaderName>)> {
+    pub fn headers(&self) -> Result<(HeaderMap<HeaderValue>, HashSet<HeaderName>)> {
         let mut headers = HeaderMap::new();
-        let mut headers_to_unset = vec![];
+        let mut headers_to_unset = HashSet::new();
```
ducaale, line 182: I wonder what would be the downside if I disable `#[deny(clippy::mutable_key_type)]` for this line?

blyxxyz: That should be ok. The lint's [documentation](https://rust-lang.github.io/rust-clippy/master/#mutable_key_type) specifically calls out the `bytes` crate as causing false positives, and I think that's what's causing `HeaderName` to be flagged as possibly having interior mutability. But there's no actual way to modify a `HeaderName` while it's inside the `HashSet`.

src/session.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn save_auth(&mut self, request_headers: &HeaderMap) -> Result<()> {
+        if let Some(value) = request_headers.get(AUTHORIZATION) {
+            self.content
+                .headers
+                .insert("authorization".into(), value.to_str()?.into());
+        }
```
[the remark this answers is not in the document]
ducaale: I have added the ability to deserialize auth from session files (no serialization at the moment).

src/session.rs
```diff
@@ -0,0 +1,253 @@
...
+
+use anyhow::{anyhow, Context, Result};
+use reqwest::header::{HeaderMap, AUTHORIZATION};
+use reqwest::Url;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Meta {
+    about: String,
+    xh: Option<String>, // needed to maintain compatibility with HTTPie's session files
```
blyxxyz, line 16:
```suggestion
    xh: Option<String>, // optional to be able to load HTTPie's session files
```

src/main.rs
ducaale: >HTTPie makes requests's session handling do the hard work. That even handles the authorization header. But I don't think that exists for reqwest.

I think reqwest does remove sensitive headers on cross-domain redirect. See https://github.com/seanmonstar/reqwest/blob/c6eb2c4fcbc3934b119772e0997d91671ff68dc7/src/redirect.rs#L234

ducaale:
>We need to rework redirects in general, redirecting with -v doesn't print intermediate requests and responses. I don't think we can let reqwest handle it any more.

I am really hoping that we won't have to go that route given the complexity of reqwests's [redirection logic](https://github.com/seanmonstar/reqwest/blob/b88f3093397095eacb4dcec6eceb8d9d5d2cccfc/src/async_impl/client.rs#L1472). If this was to be something that reqwest natively supports, what do you think it would have looked like?

ducaale:
I need to tidy up `session.rs` a little bit.

src/main.rs
```diff
@@ -315,11 +315,7 @@ fn main() -> Result<i32> {
 
     if let Some(ref mut s) = session {
         let cookie_jar = cookie_jar.lock().unwrap();
-        // TODO: investigate why expires is set to None when get_request_cookies() is used
-        // but not for iter_any()
-        let response_cookies = cookie_jar.get_request_cookies(&url).collect::<Vec<_>>();
-        s.save_cookies(response_cookies);
-
+        s.save_cookies(cookie_jar.matches(&url));
```
ducaale: See https://github.com/pfernie/cookie_store/issues/16 for why `get_request_cookies()` cannot be used at the moment.

src/session.rs
```diff
@@ -0,0 +1,342 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    secure: Option<bool>,
+}
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+struct Content {
+    #[serde(rename = "__meta__")]
+    meta: Meta,
+    #[serde(skip_serializing)]
+    auth: Option<Auth>,
```
ducaale, line 55: To prevent `auth` from being an option, `Content` could be changed to an enum. However, that seems a bit over-engineering so I will skip it. 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6ae6c75db2f2eb73f262ac08f6c65d58

src/session.rs
```diff
@@ -0,0 +1,342 @@
...
+        "0.0.0".into()
+    } else {
+        env!("CARGO_PKG_VERSION").into()
+    }
+}
+
+fn path_from_url(url: &Url) -> Result<String> {
+    match (url.host_str(), url.port()) {
+        (Some(host), Some(port)) => Ok(format!("{}_{}", host, port)),
+        (Some(host), None) => Ok(host.into()),
```
blyxxyz, line 216: This is *technically* not quite safe because host might be `.` or `..` and then you're looking in the wrong directory.

But it seems very hard to run into, and HTTPie actually has the same issue:
```
$ echo ok > ~/.config/httpie/sessions/foo.json
$ http --session foo http://.
http: error: ConfigFileError: invalid session file: Expecting value: line 1 column 1 (char 0) [~/.config/httpie/sessions/foo.json]
```
So I'm also ok with ignoring it.

blyxxyz (changes requested):
Can you test `--session-read-only`?

src/printer.rs
```diff
@@ -445,8 +446,9 @@ mod tests {
         Printer::new(pretty, args.style, false, buffer)
     }
 
-    fn temp_path(filename: &str) -> String {
+    fn temp_path() -> String {
         let mut dir = std::env::temp_dir();
+        let filename = random_string();
```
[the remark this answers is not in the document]
ducaale: I think that would create an empty file that would error when `xh` attempts to deserialize it as a JSON file.

src/main.rs
ducaale: This is now documented as of version v0.14.1. See https://github.com/pfernie/cookie_store/issues/16#issuecomment-839981352

Cargo.toml
```diff
@@ -13,6 +13,10 @@ readme = "README.md"
 [dependencies]
 anyhow = "1.0.38"
 atty = "0.2"
+base64 = "0.13"
+cookie_crate = { version = "0.15", package = "cookie" }
+cookie_store = { version = "0.14.1" }
+reqwest_cookie_store = { version = "0.1.4" }
```
ducaale, line 19: `reqwest_impl` feature now lives in its own crate.

src/session.rs
ducaale: This is now being handled by an early error

src/session.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    pub fn headers(&self) -> Result<HeaderMap> {
+        Ok(HeaderMap::try_from(&self.content.headers)?)
+    }
+
+    pub fn save_headers(&mut self, request_headers: &HeaderMap) -> Result<()> {
+        for (key, value) in request_headers.iter() {
+            let key = key.as_str();
+            if key != "cookie" && !key.starts_with("content-") && !key.starts_with("if-") {
```
blyxxyz, line 119: Can you put an explanatory comment about this?

src/session.rs
```diff
@@ -0,0 +1,340 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn save_auth(&mut self, request_headers: &HeaderMap) -> Result<()> {
+        if let Some(value) = request_headers.get(AUTHORIZATION) {
+            self.content
+                .headers
+                .insert("authorization".into(), value.to_str()?.into());
+        }
```
blyxxyz, line 133: Isn't this code as it is now redundant because of `save_headers()`?

src/session.rs
```diff
@@ -0,0 +1,340 @@
...
+                }
+            "#},
+        )?;
+
+        Session::load_session(
+            &Url::parse("http://localhost")?,
+            path_to_session.into(),
+            false,
+        )?;
+        Ok(())
```
blyxxyz, line 264: Can you test that it actually loads the cookie and header?

Idem for `can_read_xh_session_file`.

tests/cli.rs
```diff
@@ -15,6 +15,16 @@ use predicates::prelude::*;
 use serde_json::json;
 use tempfile::{tempdir, tempfile};
 
+pub fn random_string() -> String {
```
blyxxyz, line 18: Nit:
```suggestion
fn random_string() -> String {
```

blyxxyz (changes requested):
Session cookies aren't printed with the other request headers. We might have to copy some of reqwest's logic to build it ourselves.

tests/cli.rs
```diff
@@ -1329,6 +1330,25 @@ fn can_set_unset_header() {
...
+                .unwrap()
+                .iter()
+                .find(|(key, actual_value)| {
+                    key == "cookie" && {
+                        let expected = $expected_value.split("; ").collect::<HashSet<_>>();
+                        let actual = actual_value.split("; ").collect::<HashSet<_>>();
+                        actual == expected
+                    }
+                })
+                .is_some()
```
blyxxyz, line 1347: You can use [`any()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) instead

tests/cli.rs
```diff
@@ -1329,6 +1330,25 @@ fn can_set_unset_header() {
...
+        });
+    }};
+}
```
blyxxyz, line 1350: Could this be a function instead of a macro?

ducaale: Unfortunately not. See https://github.com/alexliesenfeld/httpmock/issues/44#issuecomment-840797442

src/session.rs
ducaale: Not at the moment (that is, unless we stop using `RequestBuilder::basic_auth()`)

src/session.rs
```diff
@@ -0,0 +1,340 @@
...
+        if let Some(Auth {
+            auth_type: Some(ref auth_type),
+            raw_auth: Some(ref raw_auth),
+        }) = auth
+        {
+            if auth_type.as_str() == "basic" {
+                self.headers
+                    .entry("authorization".into())
+                    .or_insert_with(|| format!("Basic {}", base64::encode(raw_auth)));
+            }
```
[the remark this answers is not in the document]
ducaale: Until we can support serializing the auth section, do we want to default to `{ "raw_auth": null, "type": null }`?

src/session.rs
ducaale: `save_auth()` has been replaced with `save_basic_auth()` and `save_bearer_auth()`

src/session.rs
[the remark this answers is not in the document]
ducaale: I have now made some changes so that data from `--bearer` and `--auth` are saved into the auth section.

src/session.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+}
+
+fn is_path(value: &OsString) -> bool {
+    value.to_string_lossy().contains(std::path::is_separator)
+}
+
+fn config_dir() -> Option<PathBuf> {
+    if test_mode() {
+        std::env::var_os("XH_TEST_CONFIG_DIR").map(Into::into)
```
ducaale, line 206: It might be good to let users customize their config directory. Should we rename this to `XH_CONFIG_DIR` so it similar to HTTPie's `HTTPIE_CONFIG_DIR`? or maybe we should have both i.e `XH_CONFIG_DIR` and `XH_TEST_CONFIG_DIR`?

src/session.rs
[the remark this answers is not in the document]
blyxxyz: I think it's better to move back to the HTTPie model and deprecate `--bearer` and undeprecate `--auth-type`. We might add more auth types later.

src/session.rs
blyxxyz: I wasn't aware of `$HTTPIE_CONFIG_DIR`, having our own version of that makes sense.

I can't offhand think of a reason to also separately have `$XH_TEST_CONFIG_DIR`.

src/session.rs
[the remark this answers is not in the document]
ducaale: I was quite fond of the `--bearer` flag, but I also think compatibility with HTTPie should come first.

ducaale:
>Session cookies aren't printed with the other request headers. We might have to copy some of reqwest's logic to build it ourselves.

Fortunately for us, `CookieStoreMutex` already implements the CookieStore trait 🙂.

src/session.rs
```diff
@@ -0,0 +1,374 @@
...
+        if let Auth {
+            auth_type: Some(ref auth_type),
+            raw_auth: Some(ref raw_auth),
+        } = self.content.auth
+        {
+            if auth_type.as_str() == "basic" {
+                return Some(format!("Basic {}", base64::encode(raw_auth)));
+            } else if auth_type.as_str() == "bearer" {
+                return Some(format!("Bearer {}", raw_auth));
+            }
```
blyxxyz, line 128: I think an unrecognized auth type should be an error.

HTTPie crashes with `http: error: KeyError: 'notarealtype'`, we can be more informative.

blyxxyz (changes requested):
Almost there. There's one issue for which we'll have to hear back from HTTPie, but we can fix that after this is merged if necessary.

src/session.rs
```diff
@@ -0,0 +1,374 @@
...
+        }
+    }
+
+    pub fn save_basic_auth(&mut self, username: String, password: Option<String>) {
+        let password = password.unwrap_or_else(|| "".into());
+        self.content.auth = Auth {
+            auth_type: Some("basic".into()),
+            raw_auth: Some(format!("{}:{}", username, password)),
+        }
+    }
```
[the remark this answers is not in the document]
blyxxyz: Let's just keep it like this for the time being. If you're using `--session` it's understood that sensitive info may end up on the disk.

blyxxyz (approved):
I think this is ready.

(#167 will break some of the tests, so watch out for that.)

PR #128, opened 2021-04-20
Change value of Accept-Encoding

sorairolake:
* Remove `deflate`, because xh does not support Deflate decoding yet.
* Add `br`, because xh supports Brotli decoding.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -142,7 +142,7 @@ fn main() -> Result<i32> {
     let request = {
         let mut request_builder = client
             .request(method, url.clone())
-            .header(ACCEPT_ENCODING, HeaderValue::from_static("gzip, deflate"))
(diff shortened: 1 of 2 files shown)


blyxxyz (approved):
Good catch!

Do you know a good website to test this? I confirmed it by comparing with curl (and discovered that `xh https://example.org accept-encoding:deflate` prints garbage) but it was a bit tedious.

ducaale:
httpbin.org supports returning responses that are encoded in different formats e.g httpbin.org/deflate.

By the way, I have been trying to [enable deflate in reqwest](https://github.com/ducaale/reqwest/commit/b8570d7757daa5e0061edb63e5bc1fc861f87fa7) (copy-pasting mostly) but that hasn't been successful yet.

PR #129, opened 2021-04-20
Change to use Trust-DNS as DNS resolver

sorairolake:
[Trust-DNS](https://github.com/bluejekyll/trust-dns) is based on Rust.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -553,6 +553,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "data-encoding"
+version = "2.3.2"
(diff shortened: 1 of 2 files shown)


blyxxyz:
What advantages would this give?

sorairolake:
@blyxxyz I think it is safer than the default DNS resolver using `getaddrinfo()`.

blyxxyz:
If it behaves differently from the normal resolver then that could be confusing. And this adds almost a megabyte to the binary size, even after stripping, so I don't think it's worth it.

sorairolake:
@blyxxyz I understand. Stop changing DNS resolver.

PR #130, opened 2021-04-21
Change Accept-Encoding in download mode

sorairolake:
This is the same as HTTPie (see: [HTTPie documentation](https://httpie.io/docs#other-notes)).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -139,10 +139,16 @@ fn main() -> Result<i32> {
 
     let client = client.build()?;
 
+    let compression_scheme = if args.download {
+        HeaderValue::from_static("identity")
+    } else {
+        HeaderValue::from_static("gzip, br")
+    };
+
     let request = {
         let mut request_builder = client
             .request(method, url.clone())
-            .header(ACCEPT_ENCODING, HeaderValue::from_static("gzip, br"))
+            .header(ACCEPT_ENCODING, compression_scheme)
             .header(CONNECTION, HeaderValue::from_static("keep-alive"))
             .header(USER_AGENT, get_user_agent());
 


blyxxyz:
Hmm. A [comment](https://github.com/httpie/httpie/blob/611bcdaab18fb4d5062da7470dd2d9c6af5d44be/httpie/downloads.py#L222) says the content shouldn't be encoded so that resuming is possible. And then it was tweaked to `identity` because a download in httpie/httpie#423 still failed (but that download no longer works).

I'm ok with merging this, "do what HTTPie does" is a safe bet. But do you have any examples of when this helps?

sorairolake:
@blyxxyz For example, when downloading an already compressed file (e.g. image file), this prevents recompression (except when `Accept-Encoding` is ignored).

See: [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding)

ducaale:
@sorairolake I don't think it is a good idea to default to disabling compression as that can increase the bytes sent over the wire by 10x. Also, setting the `accept-encoding` to `gzip, br` doesn't preclude the server from sending uncompressed responses. From MDN docs:

>Even if both the client and the server supports the same compression algorithms, the server may choose not to compress the body of a response, if the identity value is also acceptable.
>
>As long as the identity value, meaning no encoding, is not explicitly forbidden, by an identity;q=0 or a *;q=0 without another explicitly set value for identity, the server must never send back a 406 Not Acceptable error.

sorairolake:
@ducaale I understood it. Close this issue.

blyxxyz:
@ducaale But HTTPie *does* do this, so I suspect there are websites where this improves things.

ducaale:
wget also sets accept-encoding to identity so I was probably wrong in my comment.

ducaale:
I've stumbled on https://github.com/whatwg/fetch/issues/747 which links to a [site](https://jakearchibald.github.io/accept-encoding-range-test/) that lists all of the existing mp4/mp3 URLs from the HTTP archive and checks their behaviour with different `accept-encoding` values. I can see that one of them doesn't return the `content-length` unless `accept-encoding` is set to `identity` 
```
$ xh -d -c -o test.mp4 https://www.dmsg.de/fileadmin/videoarchiv/2017/TV-Spots/Interview_Kim_bearbeitet.mp4
HTTP/2.0 200 OK
accept-ranges: bytes
cache-control: max-age=2592000
content-type: video/mp4
date: Thu, 29 Apr 2021 18:40:45 GMT
expires: Sat, 29 May 2021 18:40:45 GMT
last-modified: Thu, 20 Jul 2017 08:10:28 GMT
server: Apache
vary: Accept-Encoding
x-content-type-options: nosniff


$ xh -d -c -o test.mp4 https://www.dmsg.de/fileadmin/videoarchiv/2017/TV-Spots/Interview_Kim_bearbeitet.mp4 accept-encoding:identity
HTTP/2.0 200 OK
accept-ranges: bytes
cache-control: max-age=2592000
content-length: 138155170
content-type: video/mp4
date: Thu, 29 Apr 2021 18:40:07 GMT
expires: Sat, 29 May 2021 18:40:07 GMT
last-modified: Thu, 20 Jul 2017 08:10:28 GMT
server: Apache
vary: Accept-Encoding
x-content-type-options: nosniff
```

ducaale (approved):

PR #131, opened 2021-04-22
Add the timeout flag

sorairolake:
See: #4


src/main.rs
```diff
@@ -74,10 +75,35 @@ fn main() -> Result<i32> {
...
+            if !(nanos.is_finite() && nanos.is_sign_positive()) {
+                return Err(anyhow!("Invalid seconds as connection timeout"));
+            }
+            const MAX_TIMEOUT_NANOS: f64 =
+                ((u64::MAX as u128 + 1) * (NANOS_PER_SEC as u128)) as f64;
+            if nanos >= MAX_TIMEOUT_NANOS {
+                return Err(anyhow!(
+                    "Out of range of specifiable seconds as connection timeout"
+                ));
+            }
```
ducaale, line 92: We usually do any input validation in `cli.rs`. For this particular case, it would be by defining a newtype that wraps `std::time::Duration` and implementing the `FromStr` trait for it.

```rust
struct Timeout(std::time::Duration);

impl FromStr for Timeout { /* .. */ }

struct Cli {
    #[structopt(long, value_name = "SEC")]
    pub timeout: Option<Timeout>,
}
```

However, It might not be worth the effort of handling all [possible panics](https://doc.rust-lang.org/src/core/time.rs.html#681) from `from_secs_f64` and we could get away by just making sure the value of timeout is not negative.

ducaale (commented):
Thanks for taking the time to implement this 😀. I have one comment regarding the validation

src/main.rs
sorairolake: Change to only check that the value has a positive sign. Checking finite value and overflow are disabled.

src/main.rs
```diff
@@ -74,10 +75,21 @@ fn main() -> Result<i32> {
...
+        }
+    } else {
+        None
+    };
 
     let mut client = Client::builder()
         .http2_adaptive_window(true)
-        .redirect(redirect);
+        .redirect(redirect)
+        .timeout(timeout);
```
blyxxyz, line 92: I think this is the timeout we need:
```suggestion
        .connect_timeout(timeout);
```

tests/cli.rs
```diff
@@ -509,6 +509,40 @@ fn request_binary_detection() {
         "#});
 }
 
+#[test]
+fn timeout() {
+    let mut cmd = get_command();
+    cmd.arg("--timeout=0.01")
+        .arg("GET")
+        .arg("https://httpbin.org/delay/0.5");
```
blyxxyz, line 517: Could you use a `MockServer` with the [`delay`](https://docs.rs/httpmock/0.5.8/httpmock/struct.Then.html#method.delay) method instead? Then we don't need httpbin for these tests.

And maybe 0.05 instead of 0.5 would be enough, that would make the tests a bit faster.

blyxxyz (changes requested):
Nice work!

tests/cli.rs
sorairolake: I tried it, but it seems difficult for me. I'm sorry.

blyxxyz: I can't make it work either, I guess httpmock doesn't delay in the right way.

src/main.rs
```diff
@@ -74,10 +75,21 @@ fn main() -> Result<i32> {
...
+    let timeout = if let Some(timeout) = args.timeout {
+        let timeout = timeout.as_duration();
+        if timeout == Duration::default() {
+            None
+        } else {
+            Some(timeout)
+        }
+    } else {
+        None
+    };
```
blyxxyz, line 87: Maybe this logic could go inside `as_duration()`, and that could return an `Option<Duration>` instead?

src/main.rs
```diff
@@ -74,10 +75,15 @@ fn main() -> Result<i32> {
         true => Policy::limited(args.max_redirects.unwrap_or(10)),
         false => Policy::none(),
     };
+    let timeout = args
+        .timeout
+        .map(|t| t.as_duration())
+        .filter(|t| t != &Duration::from_nanos(0));
```
blyxxyz, line 81: Nice improvement!

I still think it would make a bit more sense to have this inside `as_duration()` but it doesn't really matter.

blyxxyz (approved):

sorairolake:
Change the newline of `README.md` and `assets/basic/*` to LF. Because only these are different from others.

ducaale:
@sorairolake would you like to do the optional changes requested by @blyxxyz before I merge this?

Issue #4, opened 2020-12-11
HTTPie feature parity checklist


ducaale:
- [x] Support reading request body from stdin.
- [x] Support piping response body to stdout (including binary).
- [ ] Support unbuffered [streamed responses](https://httpie.io/docs#streamed-responses).
- [ ] Support [Download mode](https://httpie.io/docs#download-mode).
- [ ] Support [Sessions](https://httpie.io/docs#sessions).
- [ ] Decode responses compressed in deflate format.
- [ ] Add Monokai theme.

abner:
httpie uses the value of environment variable `REQUESTS_CA_BUNDLE` to verify if it is present.
Would be nice if xh uses it too.

ducaale:
It doesn't seem to be mentioned in the [docs](https://httpie.io/docs) but it is, in fact, something that [HTTPie supports](https://github.com/httpie/httpie/blob/fc45bf0fe32711cbbdda2eb5f71ef7cd25b08312/httpie/cli/definition.py#L686).

The library we use for parsing command-line arguments supports [falling back to an env variable](https://docs.rs/structopt/0.3.21/structopt/#environment-variable-fallback) so this should be an easy one to implement. Would like to create a PR for it?

blyxxyz:
requests checks both `REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE`, so perhaps we should also look for both.

ducaale:
So this means we can no longer use the nice API that structopt offers. In that case, this should be done in the `from_iter_safe()` function inside `src/cli.rs`.

vlcinsky:
`xh` does not process query parameter read from file as `httpie` does:

## Prepare a file with value for query parameter

First we create a file with a value to use:
```
$ echo "filed-value" > arg.value
```

## Use `http`, reading query parameter from the file

Then using `http` we ask to use this value read from the file for query parameter `arg`:

```
$ http https://httpbin.org/get arg==@arg.value
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 342
Content-Type: application/json
Date: Wed, 05 Oct 2022 23:11:13 GMT
Server: gunicorn/19.9.0

{
    "args": {
        "arg": "filed-value"
    },
    "headers": {
        "Accept": "*/*",
        "Accept-Encoding": "gzip, deflate",
        "Host": "httpbin.org",
[...]

PR #137, opened 2021-05-05
Add --all flag for printing intermediate requests and responses

ducaale:
Also adds the `history_print` flag for customizing printed sections for intermediate requests only.

Ticks another item from https://github.com/ducaale/xh/issues/4.


src/redirect.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+fn clone_request(request: &Request) -> ClonedRequest {
+    let mut cloned_request = Request::new(request.method().clone(), request.url().clone());
+    *cloned_request.timeout_mut() = request.timeout().cloned();
+    *cloned_request.headers_mut() = request.headers().clone();
+
+    match request.body().map(|b| b.as_bytes()) {
+        Some(Some(body)) => {
```
ducaale, line 30: I think the code in this file could be drastically simplified if we chose to buffer the body here which is an approach that something that some crates take e.g https://docs.rs/follow-redirects/0.1.3/follow_redirects/

ducaale: One problem I see with the current approach is that it's a bit unintuitive. For example, this following will successfully reach the last destination because the request body has been buffered by the `-v` flag
```
cargo run  -- --follow --all -v post pie.dev/redirect-to url==https://pie.dev/post status_code==307 @Cargo.toml
```

But this will stop midway
```
cargo run  -- --follow --all post pie.dev/redirect-to url==https://pie.dev/post status_code==307 @Cargo.toml
```

ducaale: We are now buffering the request body before following HTTP redirects.

src/redirect.rs
```diff
@@ -0,0 +1,125 @@
...
+    };
+
+    match response.status() {
+        StatusCode::MOVED_PERMANENTLY | StatusCode::FOUND | StatusCode::SEE_OTHER => {
+            let next_url = get_next_url(&request)?;
+            let prev_url = request.url().clone();
+            if is_cross_domain_redirect(&next_url, &prev_url) {
+                remove_sensitive_headers(request.headers_mut());
+            }
+            remove_content_headers(request.headers_mut());
```
ducaale, line 84: I wonder if there can be credentials embedded in Url that needs to be removed in cross-domain-redirect 🤔?

ducaale: Credentials seem to be always moved from the URL to headers, so nothing to do here.

src/redirect.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    pub fn on_redirect(&mut self, callback: T) {
+        self.callback = Some(callback);
+    }
+
+    pub fn execute(&mut self, mut first_request: Request) -> Result<Response> {
+        let mut request = clone_request(&mut first_request);
+        let mut response = self.client.execute(first_request)?;
```
blyxxyz, line 37: I think the whole buffering conundrum is worth a comment:
```suggestion
        // This buffers the body in case we need it again later
        // reqwest does *not* do this, it ignores 307/308 with a streaming body
        let mut request = clone_request(&mut first_request);
        let mut response = self.client.execute(first_request)?;
```

blyxxyz (changes requested):
I had only a few nitpicks, the code looks good.

But it doesn't seem to match HTTPie yet. `http --follow --print=HhBb httpbin.org/status/302` prints all the request bodies and only the final response body.

src/main.rs
```diff
@@ -286,34 +282,45 @@ fn main() -> Result<i32> {
...
+                redirect::RedirectFollower::new(&client, args.max_redirects.unwrap_or(10));
+            if args.all {
+                client.on_redirect(|response, request| {
+                    printer.print_response_headers(&response)?;
+                    printer.print_response_body(response)?;
+                    printer.print_seperator()?;
+                    printer.print_request_headers(request)?;
+                    printer.print_request_body(request)?;
+                    Ok(())
+                });
```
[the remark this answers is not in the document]
ducaale: I wanted to pretend as though reqwest supported `on_redirect` listener. Plus, this file is now copy-paste friendly so anyone interested in https://github.com/seanmonstar/reqwest/issues/1251 can use this.

ducaale:
>But it doesn't seem to match HTTPie yet. `http --follow --print=HhBb httpbin.org/status/302` prints all the request bodies and only the final response body.

That sounds really unintuitive, do you know why it ignores intermediate responses in this case?

src/cli.rs
```diff
@@ -74,6 +74,10 @@ pub struct Cli {
     #[structopt(short = "v", long)]
     pub verbose: bool,
 
+    /// show any intermediary requests/responses.
```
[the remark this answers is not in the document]
ducaale: >(Though I don't think what it says is actually accurate, it shows more requests than that)

Can you expand on that point a little bit?

src/main.rs
```diff
@@ -286,34 +282,45 @@ fn main() -> Result<i32> {
...
+        let status = response.status();
+        if args.check_status {
+            exit_code = match status.as_u16() {
+                300..=399 if !args.follow => 3,
+                400..=499 => 4,
+                500..=599 => 5,
+                _ => 0,
+            }
+        }
         if is_redirect && exit_code != 0 {
```
ducaale, line 319: `is_redirect` is now a bit ambiguous and could be confused with HTTP redirect. Maybe `is_redirected_output` or a similar thing would help?

src/cli.rs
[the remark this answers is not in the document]
blyxxyz: It shows all intermediary requests (if it shows requests). See the top-level discussion.

blyxxyz:
> Do you know why it ignores intermediate responses in this case?

That's what `--all` is for. The weird thing is that it does print all requests even without `--all`.

I think we should file an issue with HTTPie for clarification to fix either the implementation or the documentation.

src/main.rs
```diff
@@ -315,34 +312,48 @@ fn main() -> Result<i32> {
...
-            300..=399 if !args.follow => 3,
-            400..=499 => 4,
-            500..=599 => 5,
-            _ => 0,
+        let response = if args.follow {
+            let mut client =
+                redirect::RedirectFollower::new(&client, args.max_redirects.unwrap_or(10));
+            if let Some(history_print) = args.history_print {
+                printer.print = history_print;
+            }
```
blyxxyz, line 326: Does this mean `--history-print` applies to the final response?

blyxxyz (changes requested):
From experimenting and digging around HTTPie's source code I think its `--history-print` is unused. Or maybe I'm missing something.

-----

Printing the first request and last response by default seems reasonable to go with for now.

src/main.rs
ducaale: No, we are resetting `printer.print` just before printing the final response.

```rust
        printer.print = print;
        printer.print_response_headers(&response)?;
```

Maybe I should move that line to line 341?

blyxxyz: Ah, I missed that, even though I looked at that spot specifically. I think it's ok as-is.

blyxxyz (approved):
Two tiny nitpicks, but I'm also ok with ignoring them.

src/main.rs
```diff
@@ -316,34 +313,48 @@ fn main() -> Result<i32> {
...
+            if args.all {
+                client.on_redirect(|prev_response, next_request| {
+                    printer.print_response_headers(&prev_response)?;
+                    printer.print_response_body(prev_response)?;
+                    printer.print_seperator()?;
```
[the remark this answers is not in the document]
ducaale: I will look into this in a separate PR as I have a feeling that the fix will be a little bit involved.

PR #138, opened 2021-05-07
Replace HTTP/2 adaptive window by fixed window to prevent crashes

### LINKED ISSUE #135: Download fails with "broken pipe" error
When downloading this particular file `xh` tends to fail.

```sh
❯ xh --download --output lid.176.bin https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin
HTTP/2.0 200 OK
accept-ranges: bytes
cf-cache-status: DYNAMIC
cf-ray: 647ae6814e4ae80d-LAX
cf-request-id: 09c0ae64d00000e80d05a72000000001
content-length: 131266198
content-type: application/octet-stream
date: Thu, 29 Apr 2021 19:22:40 GMT
etag: "be9a581239b678aafe80d1d4a46d86d8-16"
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
last-modified: Fri, 18 Jan 2019 15:01:14 GMT
server: cloudflare
set-cookie: __cfduid=dfd5b87e416587032a11e8c6f286918f91619724160; expires=Sat, 29-May-21 19:22:40 GMT; path=/; domain=.fbaipublicfiles.com; HttpOnly; SameSite=Lax
x-amz-id-2: umZpKluJeIllqq/c6zhTmEKE2/dYK0mJ7QBI6xtRpEMj1sko6eU1W2m970vyu07+/Ih7QglfQPA=
x-amz-request-id: C4P9T3MK474MWB9J
x-amz-server-side-encryption: AES256
x-amz-version-id: i0bvGpDpJiHbCgE2pdfmiT6e8rhcCJdD

Downloading 125.19MB to "lid.176.bin"
Error: request or response body error: error reading a body from connection: broken pipe

Caused by:
    0: error reading a body from connection: broken pipe
    1: broken pipe
```

On my macOS laptop it seems most consistently to fail. It also intermittently fails on two Arch Linux machines. Both `curl` and `httpie` consistently download the file without issue.

blyxxyz:
The HTTP/2 adaptive window feature sends a lot of pings. Cloudflare doesn't like that and eventually breaks off the connection.

We can just set a large initial window instead. A window of 4 MiB appears to give the same speed (tested on Linux with a local HTTP/2 server and an artificial delay).

This way we also reach maximum speed earlier, the adaptive window takes a while to grow.

I'm unsure how this affects memory usage, particularly on Windows, which doesn't overcommit memory.

Resolves #135.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -77,7 +77,8 @@ fn main() -> Result<i32> {
     };
 
     let mut client = Client::builder()
-        .http2_adaptive_window(true)
+        .http2_initial_stream_window_size(4_194_304)
+        .http2_initial_connection_window_size(4_194_304)
         .timeout(timeout)
         .redirect(redirect);
 


ducaale (approved):
Looks good.

After testing this on Windows by using a local [http2 server](https://github.com/GoogleChromeLabs/simplehttp2server), I can see that this change tends to be 2x faster than `develop` with 2x-1.5x of memory usage (8MB vs 4-6MB).

PR #141, opened 2021-05-14
Enforce accept-encoding to identity in download mode

ducaale:
HTTPie behaviour is to ignore `accept-encoding` altogether in download mode.

Also, see https://httpie.io/docs#other-notes


blyxxyz (approved):
I figured there's no harm in allowing users to override it explicitly. But if HTTPie actually documents it then maybe it's best to stick to that.

PR #146, opened 2021-05-29
Support {REQUESTS, CURL}_CA_BUNDLE env variables

ducaale:
This PR adds support for HTTPie's undocumented `REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE` environment variables that behave similarly to the `--verify PATH` flag.

See https://github.com/ducaale/xh/issues/4#issuecomment-827942103

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -385,6 +385,14 @@ impl Cli {
             cli.https = true;
         }
 
+        if cli.verify.is_none() {
+            if let Some(path) = env::var_os("REQUESTS_CA_BUNDLE") {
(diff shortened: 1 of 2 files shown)


blyxxyz:
These variables will be picked up by `to_curl`, I don't think we want that. Maybe they should go in `main.rs`?

I'd mention in a comment as precedent that `requests` also looks at `CURL_CA_BUNDLE`, as looking at other libraries' environment variables is a bit weird.

blyxxyz (approved):

PR #151, opened 2021-06-04
Beautify help messages

QuarticCat:
1. Add `wrap_help` feature to clap
2. Enable `AppSettings::ColoredHelp`

Before:

image

After:

image

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -29,8 +29,11 @@ use crate::{buffer::Buffer, request_items::RequestItem};
 #[derive(StructOpt, Debug)]
 #[structopt(
     name = "xh",
-    settings = &[AppSettings::DeriveDisplayOrder, AppSettings::UnifiedHelpMessage],
-    global_setting = AppSettings::ColoredHelp,
+    settings = &[
+        AppSettings::DeriveDisplayOrder,
+        AppSettings::UnifiedHelpMessage,
+        AppSettings::ColoredHelp,
+    ],
 )]
 pub struct Cli {
     /// (default) Serialize data items from the command line as a JSON object.
@@ -300,6 +303,11 @@ impl Cli {
         match Self::from_iter_safe(iter) {
             Ok(cli) => cli,
             Err(err) if err.kind == ErrorKind::HelpDisplayed => {
+                let mut app = if env::var("NO_COLOR").is_ok() {
+                    Cli::clap().setting(AppSettings::ColorNever)
+                } else {
+                    Cli::clap()
+                };
                 // The logic here is a little tricky.
                 //
                 // Normally with structopt/clap, -h prints short help while --help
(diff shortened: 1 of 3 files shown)


src/cli.rs
```diff
@@ -27,7 +27,11 @@ use crate::{buffer::Buffer, request_items::RequestItem};
 ///
 /// It reimplements as much as possible of HTTPie's excellent design.
 #[derive(StructOpt, Debug)]
-#[structopt(name = "xh", settings = &[AppSettings::DeriveDisplayOrder, AppSettings::UnifiedHelpMessage])]
+#[structopt(
+    name = "xh",
+    settings = &[AppSettings::DeriveDisplayOrder, AppSettings::UnifiedHelpMessage],
+    global_setting = AppSettings::ColoredHelp,
```
ducaale, line 33: Can we include `AppSettings::ColoredHelp` in the settings array instead?

ducaale (changes requested):
The help looks much nicer now, thanks.

blyxxyz:
Very pretty!

Can you check if `generate.sh` still works? I'm worried `help2man` will get confused.

src/cli.rs
QuarticCat: Ah, yes. If there are subcommands, `global_setting` will also apply to them. But it seems that there is not.

QuarticCat:
@blyxxyz It still works. By default [`AppSettings::ColorAuto`](https://docs.rs/clap/2.31.2/clap/enum.AppSettings.html#variant.ColorAuto) is enabled.

ducaale:
Generated manpages seem to be okay. This is because coloring gets disabled when input is being redirected.

However, It seems that `NO_COLOR` doesn't disable colored help.

src/cli.rs
ducaale: Okay, let's use `settings` to be consistent for now.

QuarticCat:
@ducaale That's weird. Because clap also uses termcolor which respects `NO_COLOR`, and `NO_COLOR` take effect on my small cli project. I'm trying to figure out what went wrong.

src/cli.rs
QuarticCat: Will fix it in the next commit.

QuarticCat:
OK, I found why. In clap 2.3, it implements color output through ansi_term instead of termcolor, which does not support `NO_COLOR` (see https://github.com/ogham/rust-ansi-term/issues/68) and has no other global toggle (see https://github.com/ogham/rust-ansi-term/issues/10).

I'm wondering if there is a way to inject `AppSettings::ColorNever` but I have to go to sleep now.

src/cli.rs
```diff
@@ -296,6 +303,11 @@ impl Cli {
         match Self::from_iter_safe(iter) {
             Ok(cli) => cli,
             Err(err) if err.kind == ErrorKind::HelpDisplayed => {
+                let mut app = if env::var("NO_COLOR").is_ok() {
+                    Cli::clap().setting(AppSettings::ColorNever)
+                } else {
+                    Cli::clap()
+                };
```
ducaale, line 310: Can we move this logic to the `clap()` function defined on L441? That would allow us to also suppress coloring in clap's `MissingRequiredArgument` error.

src/cli.rs
```diff
@@ -296,6 +303,11 @@ impl Cli {
         match Self::from_iter_safe(iter) {
             Ok(cli) => cli,
             Err(err) if err.kind == ErrorKind::HelpDisplayed => {
+                let mut app = if env::var("NO_COLOR").is_ok() {
```
blyxxyz, line 306: After experimenting a bit I think this is the best way out. It's possible to [look at](https://docs.rs/termcolor/1.1.2/termcolor/trait.WriteColor.html#tymethod.supports_color) `termcolor`'s detection, but `NO_COLOR` is really the only piece that's missing.

One nitpick, so unicode won't matter:
```suggestion
                let mut app = if env::var_os("NO_COLOR").is_some() {
```

QuarticCat:
@blyxxyz @ducaale Done. But I find a wrong output case:

image

The last case seems not handled by clap.

blyxxyz:
Ah, that's unfortunate. It already did that, so you can ignore it for this PR if you want, and I'll look at it later. The problem seems to be that [`Error::with_description` uses automatic color detection](https://github.com/clap-rs/clap/blob/v2-master/src/errors.rs#L883).

Can you target `develop` instead of `master`?

blyxxyz (approved):

ducaale (approved):

blyxxyz:
One way this does affect `generate.sh` is that the terminal size now determines when long lines in the help output are split, and that's visible in `xh.1`.

But that doesn't affect the rendered man page so it's fine.

Issue #136, opened 2021-05-04
Allow for optionally using OpenSSL


svenstaro:
Hey, I think it'd be neat if the user could choose between using rustls or OpenSSL. For instance, I do sometimes have the requirement to use either one or the other and if xh supported that it'd be pretty good. I'm not sure whether it'd be a compile-time switch or a runtime one as linking to OpenSSL at all times would decrease the nice portability of xh which I do value.

Perhaps allow the user to _add_ OpenSSL support during compile time? Then distro packagers could choose to ship a xh supporting both versions while you still keep building the static builds with only rustls.

I imagine building this would enable a flag `--native-tls` which would then internally switch TLS calls to the linked OpenSSL.

Use cases:
- Sometimes I want it to use my system certificate store instead of the shipped certs.
- Sometimes rustls introduces weird behavior or performes badly (for instance #135)
- Sometimes I want to make sure a given server I poke is compliant in more ways than one.
 
What do you think?

blyxxyz:
That sounds useful and doable. It's possible to choose between native-tls and rustls [at runtime](https://docs.rs/reqwest/0.11.3/reqwest/blocking/struct.ClientBuilder.html#method.use_native_tls).

- Do we compile with native-tls by default? On Linux you'd need libssl-dev (or equivalent) installed. I think that random people running `cargo install` would want it disabled (for convenience) while people packaging it for distros would want it enabled. So maybe disable it by default but add a notice for packagers (that could also point out the man page and completions) or contact them directly.

- If both backends are available, which one do we use by default? Apparently reqwest can't do ALPN using the native-tls backend, so it's not a straight upgrade even when it's available.

- Do we ever want to use native-tls-vendored? That uses a statically linked OpenSSL on Linux (and matches native-tls on other platforms).

- Do we support compiling without rustls? That'd have to remove the `--cert` flag but it would cut down the binary size by a megabyte or so compared to compiling with both.

svenstaro:
> That sounds useful and doable. It's possible to choose between native-tls and rustls [at runtime](https://docs.rs/reqwest/0.11.3/reqwest/blocking/struct.ClientBuilder.html#method.use_native_tls).
> 
>     * Do we compile with native-tls by default? On Linux you'd need libssl-dev (or equivalent) installed. I think that random people running `cargo install` would want it disabled (for convenience) while people packaging it for distros would want it enabled. So maybe disable it by default but add a notice for packagers (that could also point out the man page and completions) or contact them directly.

[...]

PR #154, opened 2021-06-09
Add --native-tls flag

blyxxyz:
- [x] Enable feature by default on macOS and Windows (?)
  - [ ] Test to make sure releases still work
- [ ] Support DER certificates
- [x] Add tests
  - [x] Run native-tls tests in CI on platforms where that doesn't give issues
- [ ] Add `--debug` flag
- [x] Automatically switch the backend in certain cases?
  - Do we offer a way to still force rustls? A `--rustls` flag?
- [ ] Support for compiling without rustls, if the `--cert` flag is already partially conditional?
  - Support for compiling without TLS support?
- [x] Add warning for (or fix) seanmonstar/reqwest#1260
  - [x] Check for similar issues

See #136


ducaale:
>Enable feature by default on macOS and Windows (?)

Is this because `native-tls` doesn't use OpenSSL in these platforms? my only worry is that it could become harder to document when `native-tls` is used vs `rustls`. Or maybe this is all an implementation detail and users shouldn't worry about which tls backend is being used behind the scenes?

By the way, is there a way to mark these tls flags as, for the lack of a better word, "unstable"? because reqwest could hypothetically decide to deprecate one of the backends.

blyxxyz:
I meant the cargo feature, not the runtime feature. It'd still use rustls by default, but it'd also compile with native-tls support since (AFAIK) it's stable enough to not hurt portability.

> reqwest could hypothetically decide to deprecate one of the backends

In principle they can do anything, but is there a reason to expect that?

ducaale:
>I meant the cargo feature, not the runtime feature. It'd still use rustls by default, but it'd also compile with native-tls support since (AFAIK) it's stable enough to not hurt portability.

Ah, I see. So you mean something like this?

```toml
[target.'cfg(unix)'.dependencies.reqwest]
version = "0.11.1"
default-features = false
features = ["rustls-tls"]

[target.'cfg(any(windows, macos))'.dependencies.reqwest]
version = "0.11.1"
default-features = false
features = ["native-tls", "rustls-tls"]

# How can I detect inside the code if native-tls is enabled 🤔?
```

>In principle they can do anything, but is there a reason to expect that?

It was a hypothetical scenario. I don't have any reason to expect it.

blyxxyz:
I hoped to enable xh's own `native-tls` feature by default on those platforms, but it seems there's no way to do that. I think the best option is to enable it explicitly for our releases.

blyxxyz:
I can't find a good way to get the library versions for `--debug`. The only solution I've seen is to parse `Cargo.lock`, but that's by default ignored when running `cargo install` so it would be inaccurate.

I'll have to think a bit about what to include.

Cargo.toml
```diff
@@ -61,5 +61,7 @@ default = ["integration-tests"]
 
 integration-tests = []
 
+native-tls = ["reqwest/native-tls"]
```
ducaale, line 64: **nit**: Can we remove the newline between `integration-tests` and `native-tls`?

src/request_items.rs
```diff
@@ -141,7 +141,11 @@ impl FromStr for RequestItem {
     }
 }
 
-pub struct RequestItems(pub Vec<RequestItem>);
+#[derive(Default, Debug)]
+pub struct RequestItems {
+    pub items: Vec<RequestItem>,
+    pub request_type: RequestType,
```
ducaale, line 147: `request_type` is primarily about the request body, so maybe something like `body_type` or `request_body_type` would be more appropriate?

src/utils.rs
```diff
@@ -125,3 +126,7 @@ pub fn copy_largebuf(reader: &mut impl io::Read, writer: &mut impl Write) -> io:
 pub fn valid_json(text: &str) -> bool {
     serde_json::from_str::<serde::de::IgnoredAny>(text).is_ok()
 }
+
+pub(crate) fn requires_native_tls(url: &Url) -> bool {
```
ducaale, line 130: This name is a little bit too generic, but I can't think of a better one.

src/cli.rs
```diff
@@ -244,12 +247,16 @@ pub struct Cli {
     pub method: Option<Method>,
 
     /// The request URL.
-    #[structopt(skip)]
-    pub url: String,
+    #[structopt(skip = ("http://placeholder".parse::<Url>().unwrap()))]
```
ducaale, line 250: Is this needed because Url doesn't implement the default trait?

ducaale (commented):
Should we update this line in the `README.md`?

```diff
+ * rustls is used instead of the system's TLS library by default.
- * rustls is used instead of the system's TLS library.
```

src/cli.rs
blyxxyz: Yes.

Cargo.toml
ducaale, line 64: let's also enable reqwest's [`native-tls-alpn`](https://github.com/seanmonstar/reqwest/releases/tag/v0.11.4) feature.

src/main.rs
```diff
@@ -40,22 +39,56 @@ fn get_user_agent() -> &'static str {
...
     }
 
-    let request_items = RequestItems::new(args.request_items);
-    let query = request_items.query();
-    let (headers, headers_to_unset) = request_items.headers()?;
-    let url = construct_url(&args.url, args.default_scheme.as_deref(), query)?;
+    let warn = {
+        let bin_name = &args.bin_name;
+        move |msg| eprintln!("{}: warning: {}", bin_name, msg)
+    };
```
ducaale, line 85: Would it suffice to just use `CARGO_PKG_NAME` instead of `args.bin_name`? This would allow to turn this into a macro that is similar to `println()`.

```rust
warn(&format!("HTTP {}", status));
// vs
warn!("HTTP {}", status)
```

src/main.rs
```diff
@@ -40,22 +39,56 @@ fn get_user_agent() -> &'static str {
...
+                } else {
+                    eprintln!("Consider building with the `native-tls` feature enabled.");
+                }
+            }
+            process::exit(1);
+        }
+    }
+}
+
+fn inner_main(args: Cli) -> Result<i32> {
```
ducaale, line 76: Should we rename `inner_main()` to `run()`?

src/main.rs
```diff
@@ -80,13 +113,28 @@ fn main() -> Result<i32> {
...
+        client = client.use_native_tls();
+    } else if utils::requires_native_tls(&args.url) {
+        // We should be loud about this to prevent confusion
+        warn("rustls does not support HTTPS for IP addresses. native-tls will be enabled. Use --native-tls to silence this warning.");
+        client = client.use_native_tls();
+    }
+
+    #[cfg(not(feature = "native-tls"))]
+    if args.native_tls {
+        return Err(anyhow!("This binary was built without native-tls support"));
```
ducaale, line 132: Although we don't have to use it, Anyhow has the `bail!("error")` macro as a shorthand for `return Err(anyhow!("error"))`.

src/main.rs
```diff
@@ -40,22 +39,56 @@ fn get_user_agent() -> &'static str {
...
+            process::exit(exit_code);
+        }
+        Err(err) => {
+            eprintln!("{}: error: {:?}", bin_name, err);
+            if !native_tls && err.root_cause().to_string() == "invalid dnsname" {
+                eprintln!();
+                if utils::requires_native_tls(&url) {
+                    eprintln!("rustls does not support HTTPS for IP addresses.");
+                } else {
+                    // Maybe we went to https://<IP> after a redirect?
```
ducaale, line 60: We will hopefully be able to handle this case once we start using our own HTTP redirection logic https://github.com/ducaale/xh/pull/137

blyxxyz: Is it possible/reasonable to switch the TLS backend midway through? It'd get especially tricky if it was configured.

I think this is obscure enough that I'd rather just crash and let the user know.

src/main.rs
blyxxyz: My thinking was that you want to see which command caused the message, and the name it was invoked with is more informative.

It actually could be made a macro but I feel macros should be a tool of last resort. If we have a lot of warnings that require string formatting then it could be worth it but a closure is good enough for now.

blyxxyz:
I can't find anyone asking for DER support in HTTPie, so let's not bother. (I also don't have a good understanding of it.)

I think it would make sense for the `--debug` flag to enable logging, and at that point it deserves its own PR, so I'll leave that out.

Releasing still [works](https://github.com/blyxxyz/xh/releases/tag/untagged-57cc14293a42cbe1381a), with the usual caveat that I can't run the Windows and macOS binaries.

src/main.rs
blyxxyz: I think I'd rather not hide the control flow like that.

ducaale (approved):
Approved 🎉

Issue #144, opened 2021-05-29
Make `--check-status` default


Livven:
IMO this is a major issue for both curl and httpie that should be fixed in a Rust rewrite. Specifically:

- For interactive use, exit codes don't matter that much. Personally, I use a prompt that indicates the last exit code, but many people don't. In any case, switching to error codes shouldn't cause any issues here.
- For scripting, the current behavior is an unsafe default. Sure, there are valid uses of accepting non-200 statuses, but that should be opt-in. One of the selling points of Rust rewrites is using sensible defaults, such as ripgrep or fd excluding hidden and gitignored files. There are already way too many shell scripts out there not using `curl --fail` or `http --check-status` (or even `set -e`).

Thoughts?

blyxxyz:
I think that makes sense. Do you know if there's previous discussion of this default for HTTPie or other tools?

If we do this we should perhaps also have a strict compatibility mode to change the default back, like some GNU tools have for POSIX.

Livven:
Only found https://github.com/httpie/httpie/issues/611 but that is not about changing the default. In general, I guess there is little hope of such a change in existing tools as it would be breaking a lot of existing users. This being a newer, pre-1.0, Rust project I had more hope :)

There definitely should be a `--no-check-status` or similar flag as well. Apart from that, not sure whether full httpie compatibility is a priority for this project. I can see arguments for both sides; on the one hand, httpie itself is already quite modern, so there is less (but not zero) pressure to diverge, but conversely, it's not a "standard" Unix tool anyway, so why even bother.

ducaale:
>There definitely should be a --no-check-status or similar flag as well.

Similar to HTTPie, every flag can be prefixed with `--no-` to unset it.

>Apart from that, not sure whether full httpie compatibility is a priority for this project.

It is hard to say. Breaking compatibility with HTTPie means we can supporting new features like [nested JSON syntax](https://github.com/ducaale/xh/issues/3) without too much headache. However, it would also mean we are no longer a drop-in replacement for HTTPie and any hopes of this project being [integrated into HTTPie](https://crates.io/crates/httpie) might fade away.

But @blyxxyz's idea of a compatibility mode seems the most sensible approach. Maybe we could use the binary name to achieve that so if it is either `http` or `https`, we turn on strict compatibility mode?

ducaale:
[...]

PR #155, opened 2021-06-11
Enable check-status by default

ducaale:
- [x] Introduce strict compatibility mode
- [x] Enable `--check-status` by default if strict compatibility mode is not on.

Resolves https://github.com/ducaale/xh/issues/144

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,10 @@ xh httpbin.org/get  # resolves to http://httpbin.org/get
 xhs httpbin.org/get # resolves to https://httpbin.org/get
 ```
 
+### Strict compatibility mode
+
+TODO: explain how renaming binary to either `http` or `https` turns on strict compatibility mode.
+
 ## Examples
 
 ```sh
@@ -161,11 +165,12 @@ xh -d httpbin.org/json -o res.json
 
 ### Other differences
 
+- `--check-status` is enabled unless `xh` is being used in
+  [strict compatibility mode](https://github.com/ducaale/xh#strict-compatibility-mode).
 - `rustls` is used instead of the system's TLS library.
 - Headers are sent and printed in lowercase.
 - JSON keys are not sorted.
 - Formatted output is always UTF-8.
-- `--check-status` is enabled by default.
 
 ## Similar or related Projects
 
--- a/README.md
+++ b/README.md
@@ -177,4 +177,4 @@ xh -d httpbin.org/json -o res.json
 
 - [curlie](https://github.com/rs/curlie) - frontend to cURL that adds the ease of use of httpie
 - [httpie-go](https://github.com/nojima/httpie-go) - httpie-like HTTP client written in Go
-- [curl2httpie](https://github.com/dcb9/curl2httpie) - covert command arguments between cURL and HTTPie
+- [curl2httpie](https://github.com/dcb9/curl2httpie) - convert command arguments between cURL and HTTPie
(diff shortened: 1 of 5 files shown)


src/cli.rs
```diff
@@ -422,6 +422,12 @@ impl Cli {
         if self.auth_type == AuthType::bearer && self.auth.is_some() {
             self.bearer = self.auth.take();
         }
+        // We would like for --check-status to default to true but that is not
+        // something structopt supports.
+        self.check_status = true;
```
ducaale, line 427: What do we think about defaulting to cURL's `--fail` flag in translated commands?

README.md
```diff
@@ -165,6 +165,7 @@ xh -d httpbin.org/json -o res.json
 - Headers are sent and printed in lowercase.
 - JSON keys are not sorted.
 - Formatted output is always UTF-8.
+- `--check-status` is enabled by default.
```
blyxxyz, line 168: I'd place this higher in the list, maybe even at the top. It's more significant.

blyxxyz (changes requested):
Now that the name matters, it might be nicer to replace `clap::Arg::with_name(flag)` by `clap::Arg::with_name(&flag[2..])`. That way the arguments will be named `check-status` and `no-check-status`, nicely symmetrical.

Can you implement a strict compatibility mode?

src/cli.rs
blyxxyz: It adds noise and it's not even strictly equivalent (curl doesn't fail on 3XX status codes).

I think that ideally we'd add it only if `--check-status` was set explicitly.

We could have a `no_check_status` field that we populate manually, so other code can distinguish between explicitly check/explicitly don't check/default. (I'm not sure that's the cleanest solution.)

src/cli.rs
```diff
@@ -385,19 +391,20 @@ impl Cli {
...
-            app.get_bin_name().and_then(|name| name.split('.').next()),
-            Some("https") | Some("xhs") | Some("xhttps")
-        ) {
+        let bin_name = app.get_bin_name().and_then(|name| name.split('.').next());
+        if matches!(bin_name, Some("https") | Some("xhs") | Some("xhttps")) {
             cli.https = true;
         }
+        if matches!(bin_name, Some("http") | Some("https")) {
+            cli.strict_compat_mode = true;
+        }
```
ducaale, line 400: Maybe I should also add an env for enabling `strict_compat_mode`? This could be helpful for testing purposes and for users who have installed `xh` via a package manager.

ducaale: Update: I have now added `XH_HTTPIE_COMPAT_MODE` env variable.

blyxxyz (changes requested):
Looks good! `Option<bool>` is better than my idea about two fields :slightly_smiling_face:

README.md
```diff
@@ -170,4 +177,4 @@ xh -d httpbin.org/json -o res.json
 
 - [curlie](https://github.com/rs/curlie) - frontend to cURL that adds the ease of use of httpie
 - [httpie-go](https://github.com/nojima/httpie-go) - httpie-like HTTP client written in Go
-- [curl2httpie](https://github.com/dcb9/curl2httpie) - covert command arguments between cURL and HTTPie
\ No newline at end of file
+- [curl2httpie](https://github.com/dcb9/curl2httpie) - covert command arguments between cURL and HTTPie
```
blyxxyz, line 180: While we're at it:
```suggestion
- [curl2httpie](https://github.com/dcb9/curl2httpie) - convert command arguments between cURL and HTTPie
```

blyxxyz (approved):

PR #156, opened 2021-06-14
Replace static curl with vendored openssl

ducaale:
This is another attempt at making integration tests work everywhere. Also see the following:
* https://stackoverflow.com/a/65990512/5915221
* https://github.com/rust-embedded/cross/issues/229#issuecomment-597898074

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,6 @@ jobs:
           - target: arm-unknown-linux-gnueabihf
             os: ubuntu-latest
             use-cross: true
-            flags: --no-default-features # Integration tests don't work with cross
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


blyxxyz:
This might also be nice for the FreeBSD port, which currently [patches Cargo.toml to not use static-ssl](https://github.com/freebsd/freebsd-ports/blob/main/www/xh/files/patch-Cargo.toml).

ducaale:
That is nice, I didn't know `xh` is available in the official FreeBSD ports. I will update the `README.md` to include it.

As for this PR, all tests using httpmock are panicking with `No request has been received by the mock server`, so this doesn't seem to be the correct approach. Hopefully, once [curl is able to use `rustls` as tls backend](https://github.com/alexcrichton/curl-rust/pull/374), we can re-enable integration tests for every target.

PR #157, opened 2021-06-19
Add monokai theme

ducaale:
Ticks another item from #4

Screenshot 2021-06-19 210243


blyxxyz (commented):
Almost unreadable in a light terminal, but that's faithful to HTTPie :slightly_smiling_face: 

Can you add the origin to `assets/README.md`?

blyxxyz (approved):

ducaale:
Just out of curiosity, is the solarized theme usable in a light terminal?

blyxxyz:
It's alright:

image

But you can't always expect these themes to work with terminal themes they weren't designed for. Unless they just use the basic color codes that are defined by the terminal theme, like the `ansi` theme does.

HTTPie has a `paraiso-dark` and a `paraiso-light` theme, and one is only readable in dark terminals while the other is only readable in light terminals.

PR #158, opened 2021-06-21
Enable reqwest's deflate feature

ducaale:
~Also updates accept-encoding to match HTTPie's~


blyxxyz:
Supporting deflate is good, but Brotli is often better.

Firefox and Chromium both send `Accept-Encoding: gzip, deflate, br`, so maybe we should do the same.

blyxxyz (approved):

PR #159, opened 2021-06-25
Mark automated releases as draft

ducaale:
This can come in handy when a release fails for some targets but not for others. Draft releases should also allow us to add more information to the release before it starts notifying everyone who starred the project 😬.


blyxxyz (approved):
That sounds very sensible.

PR #161, opened 2021-06-30
Add a flag for forcing a specific http version

### LINKED ISSUE #68: Allow forcing HTTP version
Hey, I really like this tool and use it quite a bit for debugging. As such, I sometimes need to specifically request a certain version of HTTP from a server. Could you add a mode to force the HTTP version? Curl has `--http1.0`, `--http1.1`, `--http2`, `--http3` and I think that makes sense.

Please consider adding those flags.

ducaale:
Resolves #68

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -200,6 +200,9 @@ pub struct Cli {
     #[structopt(long)]
     pub https: bool,
 
+    #[structopt(long)]
+    pub http_version: Option<Version>,
+
     /// Do not attempt to read stdin.
     #[structopt(short = "I", long)]
     pub ignore_stdin: bool,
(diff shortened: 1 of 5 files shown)


svenstaro:
What's left to do here? Why is it a draft?

ducaale:
@svenstaro we are waiting for the next version of reqwest so we can make use of the newly added [`http1_only()`](https://github.com/seanmonstar/reqwest/pull/1292) option.

ducaale:
**Note to self** Remove the following from README.md

>HTTP/2 cannot be disabled. (#68)

ducaale:
**Note to self:** Remove connection-related headers when `--http-version=2` is used. See https://github.com/ducaale/xh/pull/132

src/cli.rs
```diff
@@ -797,6 +800,28 @@ impl Default for RequestType {
...
+    Http11,
+    Http2,
+}
+
+impl FromStr for Version {
+    type Err = Error;
+    fn from_str(version: &str) -> Result<Version> {
+        match version {
+            "1.0" => Ok(Version::Http10),
+            "1.1" => Ok(Version::Http11),
```
blyxxyz, line 815:
```suggestion
            "1" | "1.1" => Ok(Version::Http11),
```

src/cli.rs
```diff
@@ -797,6 +800,28 @@ impl Default for RequestType {
...
+            "2" => Ok(Version::Http2),
+            _ => Err(Error::with_description(
+                &format!("{:?} is not a valid http version", version),
+                ErrorKind::InvalidValue,
+            )),
```
blyxxyz, line 820: `0.9` and `3` exist, so maybe they should get a different error message?

ducaale:
Three tests are currently failing due to [expired certs](https://github.com/chromium/badssl.com/issues/477). I will disable them until https://self-signed.badssl.com/ starts working again.

PR #164, opened 2021-07-19
Add ;filename= parameter for multipart file uploads

### LINKED ISSUE #162: Option to set filename for multipart uploads
See header "Pinning a Directory Example": https://pinata.cloud/documentation#PinFileToIPFS

Essentially, I need to be able to set the filename of a multipart form part to be, e.g. `./foo/bar.jpg` and not just `bar.jpg` if I upload `foo/bar.jpg`.

Looks like the request item parsing already supports overriding the `Content-Type` of the file part with `;type=`: https://github.com/ducaale/xh/blob/master/src/request_items.rs#L105

So maybe `file@./foo/bar.jpg;filename=./foo/bar.jpg`? It's a bit verbose but I can work with it.

blyxxyz:
Resolves #162.

- [x] Handle for body from file
- [x] Integration tests
- [x] Documentation

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::bool_assert_comparison)]
 mod auth;
 mod buffer;
 mod cli;
--- a/src/main.rs
+++ b/src/main.rs
@@ -200,7 +200,9 @@ fn main() -> Result<i32> {
             Body::File {
                 file_name,
                 file_type,
+                file_name_header,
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::bool_assert_comparison)]
```
ducaale, line 1: Not a blocking issue, but I don't think this line is needed at the moment.

ducaale (approved):
LGTM

PR #165, opened 2021-08-02
Support reading args from a config file

ducaale:
Ticks another item from #4.

Also, see https://httpie.io/docs#configurable-options

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,6 +2,7 @@ use std::convert::TryFrom;
 use std::env;
 use std::ffi::OsString;
 use std::fmt;
+use std::fs;
 use std::io::Write;
 use std::mem;
 use std::path::PathBuf;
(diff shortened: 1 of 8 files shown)


src/cli.rs
```diff
@@ -33,6 +35,7 @@ use crate::{buffer::Buffer, request_items::RequestItem};
         AppSettings::DeriveDisplayOrder,
         AppSettings::UnifiedHelpMessage,
         AppSettings::ColoredHelp,
+        AppSettings::AllArgsOverrideSelf,
```
blyxxyz, line 38: Good catch, I didn't realize we didn't already have this behavior.

blyxxyz (changes requested):

src/cli.rs
ducaale: I also didn't know a flag like this existed until I read https://github.com/clap-rs/clap/discussions/2627

src/main.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(clippy::bool_assert_comparison)]
+#![allow(clippy::needless_borrow)]
```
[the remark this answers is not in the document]
ducaale: You are right. At first, I didn't properly understand what the warnings were about, so I disabled the `bool_assert_comparison` rule. Now I've re-enabled it and fixed the corresponding lint errors.

blyxxyz (approved):
LGTM

blyxxyz:
Should this have gone into `master`?

ducaale:
Nope, I will revert the recent changes on master and merge to develop instead 👍

PR #167, opened 2021-08-04
Send and display headers as title case

### LINKED ISSUE #166: xh lowercases all headers
## What I expect
Xh copies case of user-specified headers.

## What actually happens
Xh treats all headers as case-insensitive and lowercases them.

## Why is this a problem?
1. It's a difference from httpie.
2. It has a negative net effect when interacting with non-compliant servers.

To explain 2...

Some services treat headers as case-sensitive (e.g with `X-*` headers, especially for auth).
The HTTP specification does specify that headers are case-insensitive, however, consider the following matrix.
If headers are treated as case insensitive, this will break servers that (incorrectly) treat them as case-sensitive, while not having a negative effect on compliant servers.
If headers are treated as case sensitive, this will work with the non-compliant servers, while not having a negative effect on compliant servers.

It would also be ok to auto-capitalize headers (e.g `foO-baR-baZ` could become `Foo-Bar-Baz`), since that's what standard industry practice anyway, but this is not what HTTPie does.

## How to test compliance
`xh --print=H localhost My-OdDlY-Capitalized-Header:Here`
compare to
`http --print=H localhost My-OdDlY-Capitalized-Header:Here`

blyxxyz:
Resolves #166.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ xh -d httpbin.org/json -o res.json
 
 - Not all of HTTPie's features are implemented. ([#4](https://github.com/ducaale/xh/issues/4))
 - HTTP/2 cannot be disabled. ([#68](https://github.com/ducaale/xh/issues/68))
+- Headers are not case-sensitive.
 - No plugin system.
 - General immaturity. HTTPie is old and well-tested.
 - Worse documentation.
@@ -171,7 +172,6 @@ xh -d httpbin.org/json -o res.json
 - `--check-status` is enabled unless `xh` is being used in
   [strict compatibility mode](https://github.com/ducaale/xh#strict-compatibility-mode).
 - `rustls` is used instead of the system's TLS library.
-- Headers are sent and printed in lowercase.
 - JSON keys are not sorted.
 - Formatted output is always UTF-8.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -162,6 +162,7 @@ xh -d httpbin.org/json -o res.json
 
 - Not all of HTTPie's features are implemented. ([#4](https://github.com/ducaale/xh/issues/4))
 - HTTP/2 cannot be disabled. ([#68](https://github.com/ducaale/xh/issues/68))
+- Headers are not case-sensitive.
```
ducaale, line 165: It might be good to clarify that header values are case-sensitive

src/printer.rs
```diff
@@ -564,4 +598,40 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            p.headers_to_string(&headers, reqwest::Version::HTTP_2),
+            "ab-cd: 0
+-cd: 0
+-: 0
+ab-%c: 0
+a-b--c: 0"
+        );
```
ducaale, line 635: We can use the `indoc!` macro here just like we do in `tests/cli.rs`.

ducaale (commented):
Looks good. I have two minor comments

ducaale (approved):
LGTM

rburgst:
IMHO sending headers as title case by default is somewhat unexpected (and certainly different from how `httpie` does it. 
Note that by RFC headers should be case insensitive therefore, on the server you should usually only expect lowercase headers. 
Now if I send a request
```
xh http://localhost:3000 foo:bar
```

`xh`  will actually send `Foo: bar` to the server which is unexpected IMHO.

PR #168, opened 2021-08-07
Add --ssl option for forcing a TLS version

### LINKED ISSUE #120: Option to force TLS version
HTTPie has an option to force TLS version. I think it would be useful if xh also has this.

see also: [HTTPie documentation](https://httpie.io/docs#ssl-version)

blyxxyz:
Resolves #120.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -245,10 +245,10 @@ pub struct Cli {
     /// "auto" or "ssl2.3" gives the default behavior of negotiating a version
     /// with the server.
     #[structopt(long, value_name = "VERSION", parse(from_str = parse_tls_version),
-      possible_values = &["auto", "ssl2.3", "tls1", "tls1.1", "tls1.2", "tls1.3"],
-      default_value = "auto")]
-    // An explicit path to Option disables structopt magic we don't want here
-    pub ssl: std::option::Option<tls::Version>,
+      possible_values = &["auto", "ssl2.3", "tls1", "tls1.1", "tls1.2", "tls1.3"])]
+    // The nested option is weird, but parse_tls_version can return None.
+    // If the inner option doesn't use a qualified path structopt gets confused.
+    pub ssl: Option<std::option::Option<tls::Version>>,
 
     /// Use the system TLS library instead of rustls (if enabled at compile time).
     #[structopt(long)]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -11,7 +11,7 @@ use std::time::Duration;
 
 use anyhow::anyhow;
 use encoding_rs::Encoding;
-use reqwest::{Method, Url};
+use reqwest::{tls, Method, Url};
 use serde::{Deserialize, Serialize};
 use structopt::clap::{self, arg_enum, AppSettings, Error, ErrorKind, Result};
 use structopt::StructOpt;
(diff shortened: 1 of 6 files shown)


blyxxyz:
```
---- bad_tls_version_nativetls stdout ----
thread 'bad_tls_version_nativetls' panicked at 'Unexpected success
command=`"D:\\a\\xh\\xh\\target\\x86_64-pc-windows-msvc\\debug\\xh.exe" "--ssl=tls1.1" "--native-tls" "https://tls-v1-2.badssl.com:1012/"`
code=0
stdout=```"HTTP/1.1 200 OK [...]
```

I'm guessing this is a bug in native-tls's `max_protocol_version()` on Windows. Might end up disabling the test and printing a warning (on Windows).

src/main.rs
```diff
@@ -72,6 +74,10 @@ fn main() {
                     eprintln!("Consider building with the `native-tls` feature enabled.");
                 }
             }
+            if native_tls && msg == "invalid minimum TLS version for backend" {
+                eprintln!();
+                eprintln!("Try running without the --native-tls flag.");
+            }
```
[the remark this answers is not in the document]
blyxxyz: It happens if you combine `--ssl tls1.3` with `--native-tls`. Reqwest creates the error because native-tls doesn't let you specify it. I'll add a test.

src/cli.rs
```diff
@@ -215,6 +215,16 @@ pub struct Cli {
...
     pub cert_key: Option<PathBuf>,
 
+    /// Force a particular TLS version.
+    ///
+    /// "auto" or "ssl2.3" gives the default behavior of negotiating a version
+    /// with the server.
+    #[structopt(long, value_name = "VERSION", parse(from_str = parse_tls_version),
+      possible_values = &["auto", "ssl2.3", "tls1", "tls1.1", "tls1.2", "tls1.3"],
+      default_value = "auto")]
+    // An explicit path to Option disables structopt magic we don't want here
```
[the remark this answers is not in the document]
blyxxyz: If you write `Option<...>` structopt makes it optional and expects `parse_tls_version` to return `tls::Version` instead of `Option<tls::Version>`.

But I'm going to change this anyway because of the `default_value` thing.

blyxxyz:
> If possible, I think it's better to add TLS 1.1 or less to `possible_values` only when `native-tls` is enabled.

You get a more confusing error message that way, so I'd rather not.

src/to_curl.rs
```diff
@@ -170,6 +170,31 @@ pub fn translate(args: Cli) -> Result<Command> {
         cmd.push("--key");
         cmd.push(keyfile.to_string_lossy());
     }
+    if let Some(Some(tls_version)) = args.ssl {
+        match tls_version {
+            tls::Version::TLS_1_0 => {
+                cmd.push("--tlsv1");
```
ducaale, line 176: nit: can we use `--tlsv1.0` here instead of `--tlsv1`?

src/to_curl.rs
```diff
@@ -170,6 +170,31 @@ pub fn translate(args: Cli) -> Result<Command> {
...
+                cmd.push("--tlsv1.2");
+                cmd.push("--tls-max");
+                cmd.push("1.2");
+            }
+            tls::Version::TLS_1_3 => {
+                cmd.push("--tlsv1.3");
+                cmd.push("--tls-max");
+                cmd.push("1.3");
+            }
+            _ => unreachable!(),
```
ducaale, line 195: I think If we used our own enum for representing TLS versions, we could have avoided the catch-all arm here plus the `Option<std::option::Option>` hack. What do you think?

blyxxyz: If we do it like that then we need extra effort to ignore `TlsVersion::Auto`. So I'm not sure it'd end up cleaner overall.

src/to_curl.rs
blyxxyz: I think I did that because `--tlsv1` has existed for much longer. But `--tls-max` postdates both and `--tlsv1` on its own is wrong. So `--tlsv1.0` is better.

ducaale (approved):
LGTM

PR #174, opened 2021-09-10
Avoid printing unnecessary line separators

ducaale:
Addresses https://github.com/ducaale/xh/pull/137#discussion_r684471944

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -295,7 +295,7 @@ impl Printer {
     // except for print_response_body. We are using this function when we have
     // something to print after the response body.
     pub fn print_separator(&mut self) -> io::Result<()> {
-        if self.print.request_body {
+        if self.print.response_body {
             self.buffer.print("\n")?;
         }
         Ok(())
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -294,7 +294,7 @@ impl Printer {
     // Each of the print_* functions adds an extra line separator at the end
     // except for print_response_body. We are using this function when we have
     // something to print after the response body.
-    pub fn print_seperator(&mut self) -> io::Result<()> {
+    pub fn print_separator(&mut self) -> io::Result<()> {
         if self.print.request_body {
             self.buffer.print("\n")?;
         }
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -291,8 +291,13 @@ impl Printer {
         header_string
     }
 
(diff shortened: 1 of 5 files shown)


src/printer.rs
```diff
@@ -291,8 +291,13 @@ impl Printer {
...
 
-    pub fn print_seperator(&mut self) -> io::Result<()> {
-        self.buffer.print("\n")?;
+    // Each of the print_* functions adds an extra line separator at the end
+    // except for print_response_body. We are using this function when we have
+    // something to print after the response body.
+    pub fn print_separator(&mut self) -> io::Result<()> {
+        if self.print.request_body {
+            self.buffer.print("\n")?;
+        }
```
ducaale, line 300: I am not a fan of this change so I will revisit this in a later PR while implementing a middleware system + Digest authentication.

src/printer.rs
```diff
@@ -291,8 +291,13 @@ impl Printer {
         header_string
     }
 
-    pub fn print_seperator(&mut self) -> io::Result<()> {
-        self.buffer.print("\n")?;
+    // Each of the print_* functions adds an extra line separator at the end
+    // except for print_response_body. We are using this function when we have
+    // something to print after the response body.
+    pub fn print_separator(&mut self) -> io::Result<()> {
+        if self.print.request_body {
```
blyxxyz, line 298: I'm not sure I understand this. It looks at the `request_body` setting, but it comes right after the `response_body`, right?

Would it make more sense to add the newline into whichever of `print_request_body` and `print_response_body` needs it?

(I'm also ok with ignoring this altogether, I don't know if matching the exact number of newlines is important as long as there are no unreasonably short or long gaps.)

ducaale: >I'm not sure I understand this. It looks at the request_body setting, but it comes right after the response_body, right?

Aargh, I meant to check `self.print.response_body` instead of `self.print.request_body`.

>Would it make more sense to add the newline into whichever of print_request_body and print_response_body needs it?

`print_request_body` does already print a newline. In contrast, `print_response_body` doesn't add a newline because it is usually the last thing that would be printed (except if it is an intermediate response).

blyxxyz: Ahh, now everything makes sense.

It might be clearer to inline `print_separator` at this point but I think it all works out.

blyxxyz (approved):

PR #176, opened 2021-09-12
Add support for Digest authentication

ducaale:
Also introduces a middleware system "inspired" by https://truelayer.com/blog/adding-middleware-support-to-rust-reqwest.

Ticks another item from #4.


ducaale:
Now that `--auth` can take either a `user[:pass]` or a `token`, how its docs should be updated? and do we want to deprecate the `--bearer` flag?

src/main.rs
```diff
@@ -373,31 +376,46 @@ fn run(args: Cli) -> Result<i32> {
...
-                client.on_redirect(|prev_response, next_request| {
-                    printer.print_response_headers(&prev_response)?;
-                    printer.print_response_body(prev_response)?;
-                    printer.print_separator()?;
-                    printer.print_request_headers(next_request, &*cookie_jar)?;
-                    printer.print_request_body(next_request)?;
-                    Ok(())
-                });
+            if let Some(Auth::Digest(username, password)) = &auth {
+                // TODO: print intermediary requests/responses if --all flag is used
```
ducaale, line 415: This is a little bit tricky since `Printer` cannot be easily shared between multiple middlewares. I have thought about cloning `Printer` but that would mean I will have to clone `Buffer` which can be a file handle.

src/session.rs
```diff
@@ -115,36 +116,53 @@ impl Session {
...
+            Ok(None)
         }
     }
 
-    pub fn save_basic_auth(&mut self, username: String, password: Option<String>) {
-        let password = password.unwrap_or_else(|| "".into());
-        self.content.auth = Auth {
-            auth_type: Some("basic".into()),
-            raw_auth: Some(format!("{}:{}", username, password)),
+    pub fn save_auth(&mut self, auth: &auth::Auth) {
```
ducaale, line 145: Some auth credentials such as those from `.netrc` or passwords from prompt shouldn't be saved to session files. Maybe I should introduce a flag on `auth::Auth` to mark those?

src/main.rs
ducaale: I have opted to store the printer callback in `ClientWithMiddleware` instead

src/auth.rs
```diff
@@ -1,85 +1,139 @@
...
+                let context = digest_auth::AuthContext::new(
+                    self.username,
+                    self.password,
+                    request.url().path(),
+                );
+                let mut prompt = digest_auth::parse(wwwauth.to_str()?)?;
+                let answer = prompt.respond(&context)?.to_header_string();
+                request
+                    .headers_mut()
+                    .insert(AUTHORIZATION, HeaderValue::from_str(&answer)?);
```
ducaale, line 128: I don't think it is possible to create a [mock server](https://github.com/alexliesenfeld/httpmock/issues/49) to test this code. That is unless I can modify the second request's URL during the test.

src/middleware.rs
```diff
@@ -0,0 +1,91 @@
...
+        self.middlewares.push(Box::new(middleware));
+        self
+    }
+
+    pub fn execute(&mut self, request: Request) -> Result<Response> {
+        let mut ctx = Context::new(
+            self.client,
+            self.printer
+                .as_mut()
+                .map(|p| p as &mut dyn FnMut(Response, &mut Request) -> Result<()>),
```
blyxxyz, line 86: This seems to work:
```suggestion
            self.printer.as_mut().map(|p| p as _),
```

blyxxyz (changes requested):
This looks great! I like that unlike HTTPie it can actually print the intermediary requests.

Sorry for taking so long to review it.

src/middleware.rs
ducaale: 😮

tests/cli.rs
```diff
@@ -2197,6 +2244,122 @@ fn warns_if_config_is_invalid() {
         .success();
 }
 
+#[test]
```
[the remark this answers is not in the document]
ducaale: Done

blyxxyz (approved):

src/cli.rs
```diff
@@ -152,21 +152,22 @@ pub struct Cli {
...
+    /// Authenticate as USER with PASS or with token.
     ///
-    /// Use a trailing colon (i.e. `USER:`) to authenticate with just a username.
+    /// PASS will be prompted if missing. Use a trailing colon (i.e. `USER:`)
+    /// to authenticate with just a username.
+    ///
+    /// if --auth-type=bearer then --auth expects a token
     /// {n}{n}{n}
-    #[structopt(short = "a", long, value_name = "USER[:PASS]")]
+    #[structopt(short = "a", long, value_name = "USER[:PASS] | token")]
```
blyxxyz, line 166: I'd capitalize `TOKEN` here, and write the last line as ``TOKEN is expected if `--auth-type=bearer`.``.

ducaale: Done

PR #190, opened 2021-11-11
Replace httpmock with hyper

### LINKED ISSUE #186: Consider replacing httpmock with hyper+helpers
### My reasons against Httpmock
1. It has a dependency on `curl+openssl` which can be problematic when running tests in CI.
2. It is not possible to have a custom logic that returns a [different response depending on the request](https://github.com/alexliesenfeld/httpmock/issues/49).
3. `Httpmock::When` [cannot be currently extended]( https://github.com/alexliesenfeld/httpmock/issues/44#issuecomment-840797442). This limitation can sometimes force us to either duplicate code or [use macros](https://github.com/ducaale/xh/pull/125#discussion_r658814597).

On the other hand, `hyper`+[helpers](https://github.com/seanmonstar/reqwest/tree/master/tests/support) can do everything `httpmock` does without the mentioned limitations. The only problem I can think of is that hyper can be a bit verbose but that verbosity can be easily abstracted away (I think).

### A typical Httpmock test
```rust
#[test]
fn basic_json_post() {
    let server = MockServer::start();
    let mock = server.mock(|when, then| {
        when.method(POST)
            .header("Content-Type", "application/json")
            .json_body(json!({"name": "ali"}));
        then.header("Content-Type", "application/json")
            .json_body(json!({"got": "name", "status": "ok"}));
    });

    get_command()
        .arg("--print=b")
        .arg("--pretty=format")
        .arg("post")
        .arg(server.base_url())
        .arg("name=ali")
        .assert();

    mock.assert();
}
```

### The same test but in hyper
```rust
// Inspired by reqwest's tests. See https://github.com/seanmonstar/reqwest/tree/master/tests
#[tokio::test]
async fn basic_json_post() {
    let server = server::http(move |req| async move {
        assert_eq!(req.method(), "POST");
        assert_eq!(req.headers()["Content-Type"], "application/json");
        let mut full: Vec<u8> = Vec::new();
        while let Some(item) = req.body_mut().next().await {
            full.extend(&*item.unwrap());
        }
        assert_eq!(full, b"{\"name\": \"ali\"}");

        http::Response::builder()
            .header(http::header::CONTENT_TYPE, "application/json")
            .body("{\"got\": \"name\", \"status\": \"ok\"}".into())
            .unwrap()   
    });

    get_command()
        .arg("--print=b")
        .arg("--pretty=format")
        .arg("post")
        .arg(server.addr())
        .arg("name=ali")
        .assert();
}
```

ducaale:
Also introduces a feature flag named `online-tests` that can be used to filter out tests that require an internet connection.

Resolves #186


tests/server.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use hyper::service::{make_service_fn, service_fn};
+use tokio::runtime;
+use tokio::sync::oneshot;
+
+pub struct Server {
+    addr: net::SocketAddr,
+    hits: Arc<Mutex<u8>>,
+    // The panic_rx channel is to make sure the test fails if the server panics.
+    // If the server panics, the message is not sent and the recv call panics.
```
ducaale, line 19: https://discord.com/channels/500028886025895936/670880858630258689/906845871021383700

blyxxyz:
The problem with `x86_64-unknown-linux-gnu` seems to be that Ubuntu 20.04 by default disables the ciphers TLS1.1 needs to function, even if you enable that TLS version. HTTPie also suffers from this but it has a `--ciphers` flag that can presumably help. IIRC reqwest and native-tls don't let you set the cipher suite.

An easy fix is to enable the `vendored` feature on the openssl crate. That's not something we'd want to do by default but it could make sense to offer outside of testing.

An even easier fix is to change the `good_tls_version_nativetls` test to test TLS1.2 instead. I think that's the best option for now.

blyxxyz:
I managed to get the integration tests working on ARM:
```rust
/// Cargo-cross for ARM runs tests using qemu.
///
/// It sets an environment variable like this:
/// CARGO_TARGET_ARM_UNKNOWN_LINUX_GNUEABIHF_RUNNER=qemu-arm
fn find_runner() -> Option<String> {
    for (key, value) in std::env::vars() {
        if key.starts_with("CARGO_TARGET_") && key.ends_with("_RUNNER") && !value.is_empty() {
            return Some(value);
        }
    }
    None
}

fn get_base_command() -> Command {
    let mut cmd;
    let path = assert_cmd::cargo::cargo_bin("xh");
    if let Some(runner) = find_runner() {
        cmd = Command::new(runner);
        cmd.arg(path);
    } else {
        cmd = Command::new(path);
    }
    cmd.env("HOME", "");
    #[cfg(target_os = "windows")]
    cmd.env("XH_TEST_MODE_WIN_HOME_DIR", "");
    cmd
}
```

tests/cli.rs
```diff
@@ -986,13 +1133,13 @@ fn good_tls_version() {
...
-#[cfg(feature = "native-tls")]
+#[cfg(all(feature = "native-tls", feature = "online-tests"))]
 #[test]
 fn good_tls_version_nativetls() {
     get_command()
-        .arg("--ssl=tls1.1")
+        .arg("--ssl=tls1.2")
         .arg("--native-tls")
-        .arg("https://tls-v1-1.badssl.com:1011/")
+        .arg("https://tls-v1-2.badssl.com:1011/")
```
blyxxyz, line 1142:
```suggestion
        .arg("https://tls-v1-2.badssl.com:1012/")
```

ducaale:
Woohoo! I can't believe that integration tests are finally working for ARM. How does the solution work? are there any docs I can refer to?

Oh, and should we get rid of the `integration-tests` feature flag?

blyxxyz:
I don't know if it's documented anywhere (and how safely we can rely on it).

Cross uses qemu to run binaries for other architectures. You prefix a command with e.g. `qemu-arm` and then qemu translates it for your host system.

Cross fiddles with cargo to make sure your test binaries are run through qemu. But we build our own commands and they aren't fixed automatically.

I had a suspicion it was something like this, and discovered the details and this solution by panicking in the tests with the output of `pgrep -af .` and `env`.

I think this would've worked before rust-embedded/cross#122, when it still used [binfmt](https://en.wikipedia.org/wiki/Binfmt_misc) and didn't need to make the runner explicit.

> Oh, and should we get rid of the `integration-tests` feature flag?

Yeah, I think we no longer have a need for it.

tests/server.rs
blyxxyz: I'm not in that server (and can't tell which server it is) so I can't view that message

blyxxyz (approved):
This looks great! I also like the `online-tests` feature.

I noticed some inconsistency between writing `http::Response` and `hyper::Response`, `use http::Response;` could clean things up. But that doesn't really matter.

tests/server.rs
ducaale: It is in [tokio's discord server](https://discord.com/invite/tokio)

blyxxyz:
I'm ok with merging now and re-enabling the test after chromium/badssl.com#482 is fixed.

PR #191, opened 2021-11-12
Change to output the feature status only in `--version`

sorairolake:
Remove the feature status from output of `-V` and `--help` (same as `rg`).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -36,7 +36,7 @@ use crate::utils::config_dir;
 #[derive(StructOpt, Debug)]
 #[structopt(
     name = "xh",
-    version = concat!(env!("CARGO_PKG_VERSION"), "\n", env!("XH_FEATURES")),
+    long_version = long_version(),
     settings = &[
         AppSettings::DeriveDisplayOrder,
         AppSettings::UnifiedHelpMessage,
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -62,6 +62,13 @@ macro_rules! regex {
     }};
 }
 
+#[macro_export]
+macro_rules! long_version {
+    () => {
+        concat!(env!("CARGO_PKG_VERSION"), "\n", env!("XH_FEATURES"))
+    };
+}
+
```
ducaale, line 71: This could have been a normal function instead of a macro rule. However, I am in favour of inlining this logic since it is only used once.

sorairolake: Fixed.

ducaale (approved):
LGTM

src/cli.rs
```diff
@@ -1009,6 +1009,11 @@ fn safe_exit() -> ! {
     std::process::exit(0);
 }
 
+#[inline]
```
blyxxyz, line 1012: There's no need to add this, "obvious" things are inlined automatically. Compilers really like inlining things.

You only need it if you want something to be inlined across crates, or if you think the compiler will guess wrong.

(It's harmless so there's no hurry to clean it up, and I did the same thing in `buffer.rs` before I knew this :slightly_smiling_face:)

ducaale: Oh I see. I will remember to remove the inline attribute here before making a release tonight.

PR #192, opened 2021-11-13
Change the unit of elapsed time in download result

sorairolake:
The minimum unit of elapsed time in download result of HTTPie is 10 microseconds.
So, I change this from a second to a nanosecond.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -248,19 +248,16 @@ pub fn download_file(
             copy_largebuf(&mut pb.wrap_read(response), &mut buffer)?;
             let downloaded_length = pb.position() - starting_length;
             pb.finish_and_clear();
-            let time_taken = starting_time.elapsed().as_secs();
-            if let Some(speed) = downloaded_length.checked_div(time_taken) {
+            let time_taken = starting_time.elapsed();
+            if time_taken != Duration::from_nanos(0) {
                 eprintln!(
                     "Done. {} in {} ({}/s)",
                     HumanBytes(downloaded_length),
-                    humantime::format_duration(Duration::from_secs(time_taken)),
-                    HumanBytes(speed)
+                    humantime::format_duration(time_taken),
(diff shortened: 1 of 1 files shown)


src/download.rs
```diff
@@ -248,19 +248,16 @@ pub fn download_file(
...
-            if let Some(speed) = downloaded_length.checked_div(time_taken) {
+            let time_taken = starting_time.elapsed();
+            if time_taken != Duration::from_nanos(0) {
                 eprintln!(
-                    "Done. {} in {} ({}/s)",
+                    "Done. {} in {:?} ({}/s)",
                     HumanBytes(downloaded_length),
-                    humantime::format_duration(Duration::from_secs(time_taken)),
-                    HumanBytes(speed)
+                    Duration::from_micros(time_taken.as_micros() as u64),
```
ducaale, line 256: Can we keep using `humantime`? the maximum unit `Duration` supports is seconds while `humantime` can display time in minutes, hours, days, and so on.

sorairolake: Fixed.

src/download.rs
```diff
@@ -248,19 +248,18 @@ pub fn download_file(
...
+            let time_taken = starting_time.elapsed();
+            if time_taken >= Duration::from_micros(1) {
                 eprintln!(
                     "Done. {} in {} ({}/s)",
                     HumanBytes(downloaded_length),
-                    humantime::format_duration(Duration::from_secs(time_taken)),
-                    HumanBytes(speed)
+                    humantime::format_duration(
+                        Duration::from_micros(time_taken.as_micros() as u64)
+                    ),
```
ducaale, line 258: Can we pass `time_taken` directly to `format_duration()`?

```rust
humantime::format_duration(time_taken)
```

sorairolake: Fixed.

sorairolake: Fixed. Also, change the condition to `Duration` is non-zero.

ducaale (approved):
Approved 🎉

PR #193, opened 2021-11-14
Change to use binary prefix in `--download`

sorairolake:
Use [binary prefix](https://en.wikipedia.org/wiki/Binary_prefix) instead of SI prefix for disambiguation.


ducaale:
As of `indicatif` v0.16, `HumanBytes` and `BinaryBytes` are formatted the same. Do you mind reverting the changes in `download.rs`? thanks.

sorairolake:
@ducaale Reverted `download.rs`.

PR #194, opened 2021-11-16
Do not rebuild when no files changed

blyxxyz:
We were asking cargo to rebuild if `assets/xhs` or `assets/xhs.1.gz` changed. Those are dead symlinks, so cargo always errors and [assumes they changed](https://github.com/rust-lang/cargo/blob/e475fe4a1780dafb771b39cddf8097ec206b0028/src/cargo/core/compiler/fingerprint.rs#L1733).

That caused a lot of unnecessary rebuilds. I found the problem by running with `CARGO_LOG=cargo::core::compiler::fingerprint=info`.

Asking cargo to watch only the files we actually care about fixes it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -35,10 +35,11 @@ fn main() {
     for dir in &["assets", "assets/basic", "assets/large"] {
         println!("cargo:rerun-if-changed={}", dir);
         for entry in read_dir(dir).unwrap() {
-            println!(
-                "cargo:rerun-if-changed={}",
-                entry.unwrap().path().to_str().unwrap()
-            );
+            let path = entry.unwrap().path();
+            let path = path.to_str().unwrap();
+            if path.ends_with(".sublime-syntax") || path.ends_with(".tmTheme") {
+                println!("cargo:rerun-if-changed={}", path);
+            }
         }
     }
 


ducaale:
How often `XH_FEATURES` will be updated with this change?

**PS:** I will not have access to my computer for more than a week. Until then, I won't be able to properly review your PRs 😔

blyxxyz:
> How often `XH_FEATURES` will be updated with this change?

It keeps working, I think cargo keeps track of each set of features separately.

> **PS:** I will not have access to my computer for more than a week. Until then, I won't be able to properly review your PRs :pensive:.

Thank you for the heads-up!

ducaale (approved):
Since `XH_FEATURES` is updated in `build.rs` which should only be rerun whenever a change in syntax files is detected, I was worried about the possibility of `XH_FEATURES` ending up with a stale value. But it turns out cargo will always execute `build.rs` whenever a different set of features is passed to it (verified it by printing inside `build.rs` and using cargo's `-vv` flag).

```
$ cargo run -- --version
   Compiling xh v0.14.0 (/home/ducaale/xh)
    Finished dev [unoptimized + debuginfo] target(s) in 25.03s
     Running `target/debug/xh --version`
xh 0.14.0
-native-tls
$ cargo run --features=native-tls -- --version
   Compiling xh v0.14.0 (/home/ducaale/xh)
    Finished dev [unoptimized + debuginfo] target(s) in 25.39s
     Running `target/debug/xh --version`
xh 0.14.0
+native-tls
```

Approved!

PR #195, opened 2021-11-16
Allow Setting the User Agent For Requests

zicklag:
Hey there! I've been using `xh` for random requests and I had the need to set the user agent for a request I was making. Here's a PR that let's you do that with either the `-u` or `--user-agent` flag. Let me know if there's anything I need to fix!

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -292,6 +292,10 @@ pub struct Cli {
     #[structopt(value_name = "[METHOD] URL")]
     raw_method_or_url: String,
 
+    /// The user agent to make the request with.
+    #[structopt(long, short = "u")]
+    pub user_agent: Option<String>,
+
     /// Optional key-value pairs to be included in the request
     ///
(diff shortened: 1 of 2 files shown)


blyxxyz:
Thank you! But there's already another way to set the user agent, by writing it out as a `header:value` argument:
```console
$ xh -v httpbin.org/get user-agent:foobar
GET /get HTTP/1.1
Accept: */*
Connection: keep-alive
Host: httpbin.org
User-Agent: foobar
```
I think an option for it would be superfluous.

zicklag:
There, just updated this PR with an example for the README. :D

zicklag:
Nevermind, GitHub won't let me re-open this pull request. I'll open a new one.

b1nhack:
Maybe you should reopen this pr, only by adding the --user-agent option, you can change the default `User-Agent: xh/0.22.2` in the config file.

PR #198, opened 2021-11-26
Small fixes

blyxxyz:
Nothing really significant here, it's basically about style.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -56,7 +56,6 @@ impl Buffer {
         matches!(self, Buffer::Redirect(..))
     }
 
-    #[inline]
     pub fn print(&mut self, s: impl AsRef<[u8]>) -> io::Result<()> {
         self.write_all(s.as_ref())
     }
@@ -97,7 +96,6 @@ impl Buffer {
 }
 
 impl Write for Buffer {
-    #[inline]
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         match self {
             Buffer::File(file) => file.write(buf),
@@ -108,15 +106,6 @@ impl Write for Buffer {
 
     fn flush(&mut self) -> io::Result<()> {
         self.inner_mut().flush()
-    }
-
-    #[inline]
-    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
-        match self {
-            Buffer::File(file) => file.write_all(buf),
-            Buffer::Stdout(stream) | Buffer::Stderr(stream) => stream.write_all(buf),
-            Buffer::Redirect(stream) => stream.write_all(buf),
-        }
     }
 }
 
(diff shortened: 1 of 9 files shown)


tests/server.rs
```diff
@@ -71,6 +71,7 @@ where
             .expect("new rt");
         let srv = {
             let hits_counter = hits_counter.clone();
+            #[allow(clippy::async_yields_async)]
```
ducaale, line 74: I haven't seen any Clippy errors from this file. Is this because Clippy ignores test files?

src/buffer.rs
```diff
@@ -109,15 +107,6 @@ impl Write for Buffer {
...
     }
-
-    #[inline]
-    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
-        match self {
-            Buffer::File(file) => file.write_all(buf),
-            Buffer::Stdout(stream) | Buffer::Stderr(stream) => stream.write_all(buf),
-            Buffer::Redirect(stream) => stream.write_all(buf),
-        }
-    }
```
ducaale, line 120: Is this because none of these buffer types have an efficient implementation of `write_all()`?

ducaale (commented):
Looks good. I have 2 questions

tests/server.rs
blyxxyz: The trick is apparently to run `cargo clippy --tests`. We could add that to the CI.

I saw them pop up in VS Code.

ducaale: Ah, I see. Do you mind including that in this PR?

src/buffer.rs
blyxxyz: My thinking when I wrote it was that if `write_all` has to do many writes then it would only have to do the `match` once, instead of doing it for every individual `write`. But writes are expensive, an extra `match` doesn't matter.

I hadn't thought of efficient `write_all()` implementations, but that's a great point, because there could've been something like that. Writes have to acquire a lock and it's nice if that doesn't have to happen for every individual write. But I benchmarked it and it turns out not to make a difference.

Here's the patch I used, for posterity:
```diff
diff --git a/src/buffer.rs b/src/buffer.rs
index 3a5c437..e983c8c 100644
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -114,8 +114,8 @@ impl Write for Buffer {
     fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
         match self {
             Buffer::File(file) => file.write_all(buf),
-            Buffer::Stdout(stream) | Buffer::Stderr(stream) => stream.write_all(buf),
-            Buffer::Redirect(stream) => stream.write_all(buf),
+            Buffer::Stdout(stream) | Buffer::Stderr(stream) => stream.lock().write_all(buf),
+            Buffer::Redirect(stream) => stream.get_mut().lock().write_all(buf),
         }
     }
 }
```

ducaale (approved):
LGTM

PR #199, opened 2021-11-26
Remove http crate from Cargo.toml

ducaale:
This should have been part of https://github.com/ducaale/xh/pull/190 to address https://github.com/ducaale/xh/pull/190#pullrequestreview-807794895 but I forgot to include it.

>I noticed some inconsistency between writing http::Response and hyper::Response, use http::Response; could clean things up. But that doesn't really matter.


blyxxyz (approved):
Looks good.

Welcome back, by the way!

ducaale:
Thanks 🙂

PR #200, opened 2021-11-26
to_curl: Improve quoting, update options

blyxxyz:
The quoting done by shell-escape is very simple and conservative, not great for user-readable output. I recently wrote [`os_display`](https://github.com/blyxxyz/os_display) which I think is more suitable here.

I also went through all the options to bring them up to date.


blyxxyz:
The ARM tests get stuck on `inferred_json_javascript_output`. I can't reproduce that locally, and it seems impossible for it to be caused by these commits, so I don't know what's going on.

ducaale:
All tests have passed now (there is `proxy_all_proxy` that fails silently but it can be adressed in a later PR).

If ARM tests start to fail again, perhaps we could try limiting the number of parallel `qemu-arm` invocations in integration tests?

src/to_curl.rs
```diff
@@ -385,118 +419,91 @@ mod tests {
     #[test]
     fn examples() {
         let expected = vec![
-            ("xh httpbin.org/get", "curl 'http://httpbin.org/get'", "curl http://httpbin.org/get"),
+            ("xh httpbin.org/get", "curl http://httpbin.org/get"),
             (
                 "xh httpbin.org/post x=3",
-                r#"curl 'http://httpbin.org/post' -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d '{"x":"3"}'"#,
-                r#"curl http://httpbin.org/post -H "content-type: application/json" -H "accept: application/json, */*;q=0.5" -d "{\"x\":\"3\"}""#,
+                r#"curl http://httpbin.org/post -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d '{"x":"3"}'"#,
```
ducaale, line 425: Powershell seems to strip double quotes unless they are escaped
```
PS C:\> curl http://httpbin.org/post -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d '{"x":"3"}'
{
  "args": {},
  "data": "{x:3}",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "application/json, */*;q=0.5",
    "Content-Length": "5",
    "Content-Type": "application/json",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.55.1",
    "X-Amzn-Trace-Id": "Root=1-61a25bb0-075508f07946ad341dc7e3f1"
  },
  "json": null,
  "origin": "85.76.18.84",
  "url": "http://httpbin.org/post"
}
PS C:\> curl http://httpbin.org/post -H "content-type: application/json" -H "accept: application/json, */*;q=0.5" -d '{\"x\":\"3\"}'
{
  "args": {},
  "data": "{\"x\":\"3\"}",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "application/json, */*;q=0.5",
    "Content-Length": "9",
    "Content-Type": "application/json",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.55.1",
    "X-Amzn-Trace-Id": "Root=1-61a25bbb-7bdf444d318dfa624eb9bf3a"
  },
  "json": {
    "x": "3"
  },
  "origin": "85.76.18.84",
  "url": "http://httpbin.org/post"
}
```

blyxxyz: Ah, that's true. It has to be fixed in `os_display`. I didn't realize that shell-escape did handle that properly, but that makes sense because it's inherited from cmd.exe.

I'll put this PR on hold for now.

blyxxyz: (The reason this is tricky is that PowerShell removes the quotes when spawning an external command, not when parsing the string. And it doesn't remove them when you pass it to an internal commandlet.)

ducaale: Since the last version of xh has this quoting issue, what if we ignore it in this PR and track it in a GitHub issue instead?

blyxxyz: Does it? Did shell-escape change?

I already wrote a fix FWIW, I just have to get around to deciding whether I like the design and committing/publishing it.

ducaale: It didn't change but we never made use of the cmd.exe-style formatting

```
let escape = if f.alternate() {
    // If formatted with `{:#}`, use cmd.exe-style formatting
    // This is currently not exposed
    shell_escape::windows::escape
}
```

blyxxyz: Ah, I had already forgotten!

I'll just try to get `os_display` updated this weekend, but that would have been reasonable.

blyxxyz:
I haven't done a new `os_display` release yet, I'll do that if this looks ok.

ducaale:
Yup, I can confirm that It is working now.

```
cargo run -- httpbin.org/post x=3 --curl
    Finished dev [unoptimized + debuginfo] target(s) in 0.36s
     Running `target\debug\xh.exe httpbin.org/post x=3 --curl`
curl http://httpbin.org/post -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d '{\"x\":\"3\"}'

PS C:\Users\shara\Desktop\projects\xh> curl http://httpbin.org/post -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d '{\"x\":\"3\"}'
{
  "args": {},
  "data": "{\"x\":\"3\"}",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "application/json, */*;q=0.5",
    "Content-Length": "9",
    "Content-Type": "application/json",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.79.1",
    "X-Amzn-Trace-Id": "Root=1-61ec4860-1bf0ad7325e7ce3746ae47ca"
  },
  "json": {
    "x": "3"
  },
  "origin": "92.40.172.172",
  "url": "http://httpbin.org/post"
}
```

src/to_curl.rs
```diff
@@ -208,21 +217,25 @@ pub fn translate(args: Cli) -> Result<Command> {
             }
         }
     }
+    if let Some(timeout) = args.timeout.and_then(|t| t.as_duration()) {
+        cmd.arg("--connect-timeout");
+        cmd.arg(timeout.as_secs_f64().to_string());
+    }
```
ducaale, line 223: I think `--max-time` aligns more with httpie/xh's timeout behavior:

```
$ xh --timeout=1 httpbin.org/delay/5
xh: error: error sending request for url (http://httpbin.org/delay/5): operation timed out

$ http  --timeout=1 httpbin.org/delay/5
http: error: Request timed out (1.0s).

$ curl --max-time 1 http://httpbin.org/delay/5
curl: (28) Operation timed out after 1011 milliseconds with 0 bytes received

$ curl --connect-timeout 1 http://httpbin.org/delay/5
{
  "args": {},
  "data": "",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "*/*",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.79.1",
    "X-Amzn-Trace-Id": "Root=1-61ec810c-40ce3cb07fd36d5d1ecd85b9"
  },
  "origin": "92.40.168.187",
  "url": "http://httpbin.org/delay/5"
}
```

ducaale (approved):
Approved 🎉

PR #201, opened 2021-11-28
Test server improvements

ducaale:
This PR enhances the test server to automatically check at the end of each test that at least one request has been made and there are no pending or panicked requests.

Before these changes, it was easy to introduce a test that fails silently. For example, the following test wouldn't fail unless `.success()`/`.stdout()` was used or `server.hits` was checked to be greater than 0.

```rust
#[test]
fn basic_get() {
    let server = server::http(|req| async move {
        assert_eq!(req.method(), "INVALID");
        hyper::Response::builder().body("foobar\n".into()).unwrap()
    });
    get_command()
        .args(&["--print=b", "get", &server.base_url()])
        .assert();
}
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,6 +29,7 @@ jobs:
           - target: arm-unknown-linux-gnueabihf
             os: ubuntu-latest
             use-cross: true
+            flags: -- --test-threads=4
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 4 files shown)


blyxxyz (approved):
Looks good! I hope the number of test threads was the culprit.

ducaale:
I hope so too.

Currently, our CI is configured to run tests on the master branch every night so I am thinking of cherry-picking [this commit](https://github.com/ducaale/xh/pull/201/commits/8b73d673cebfba3b4e529a597ad4cb1cade7e840) from this PR to master.

blyxxyz:
Sounds like a good idea.

PR #202, opened 2021-11-29
add support for HTTPie's --raw flag

### LINKED ISSUE #178: Support HTTPie's `--raw` flag.
See https://httpie.io/docs#request-data-via---raw

ducaale:
Resolves #178

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -111,7 +111,7 @@ fn run(args: Cli) -> Result<i32> {
             stdin().read_to_end(&mut buffer)?;
             Body::Raw(buffer)
         }
-        (false, Some(raw), false) => Body::Raw(raw.as_bytes().to_vec()),
+        (false, Some(raw), false) => Body::Raw(raw.into_bytes()),
         (false, None, _) => body_from_request_items,
 
         (true, Some(_), _) => {
--- a/src/main.rs
+++ b/src/main.rs
@@ -121,6 +121,7 @@ fn run(args: Cli) -> Result<i32> {
         }
         (true, _, true) => {
             if args.multipart {
+                // Multipart bodies are never "empty", so we can get here without request items
                 return Err(anyhow!("Cannot build a multipart request body from stdin"));
             } else {
                 return Err(anyhow!(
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -102,10 +102,25 @@ fn run(args: Cli) -> Result<i32> {
...
+
+        (true, Some(_), _) => {
+            return Err(anyhow!(
+                "Request body from stdin and --raw cannot be mixed. \
+                Pass --ignore-stdin to ignore standard input."
+            ))
+        }
+        (true, _, true) => {
+            if args.multipart {
                 return Err(anyhow!("Cannot build a multipart request body from stdin"));
```
blyxxyz, line 124: This code isn't new (apparently I wrote it?) but it did confuse me, I think it's worth a comment:
```suggestion
           if args.multipart {
                // Multipart bodies are never "empty", so we can get here without request items
                return Err(anyhow!("Cannot build a multipart request body from stdin"));
```

tests/cli.rs
```diff
@@ -1350,6 +1350,30 @@ fn mixed_stdin_request_items() {
         ));
 }
 
+#[test]
```
blyxxyz, line 1353: Can you also test successful use of the option?

blyxxyz (changes requested):
Works great!

It's not caused by this change, but I just noticed that HTTPie doesn't use a POST request for a zero-length body, e.g. `printf '' | xh --offline :`. I don't know how straightforward that'll be to fix, so we could make an issue for it instead.

tests/cli.rs
```diff
@@ -502,12 +501,13 @@ fn streaming_binary_detection() {
...
 #[test]
 fn request_binary_detection() {
-    let mut binary_file = tempfile().unwrap();
+    let mut binary_file = NamedTempFile::new().unwrap();
     binary_file.write_all(b"foo\0bar").unwrap();
     binary_file.seek(SeekFrom::Start(0)).unwrap();
     redirecting_command()
         .args(&["--print=B", "--offline", ":"])
-        .stdin(binary_file)
+        .pipe_stdin(binary_file.path())
```
blyxxyz, line 509: Would `.write_stdin(b"foo\0bar")` work?

blyxxyz (approved):
Looks great!

PR #206, opened 2021-12-01
Add Fruity theme

ducaale:
There are several references to this theme in HTTPie docs, so supporting it should lead to less confusion for anyone using HTTPie docs with xh. ~In addition, I have updated all themes to not highlight the escape character with a different color~.

Ticks one item from #4.


blyxxyz:
> I have updated all themes to not highlight the escape character with a different color.

Is that just because HTTPie doesn't do it? I'm not very attached to it but I do like it.

ducaale:
>Is that just because HTTPie doesn't do it? I'm not very attached to it but I do like it.

I didn't have any good reasons beyond making the output similar to HTTPie. If you think the escape character highlighting could be helpful, I could revert the change.

blyxxyz (approved):

PR #207, opened 2021-12-01
Use custom netrc parser

blyxxyz:
This removes our most obscure direct dependency.

The new parser:
- Supports comment syntax (like Python/requests/HTTPie)
- Is more faithful to HTTPie in general
- Is less strict (maybe a good thing because we don't show errors)
- Normalizes hostnames (we could've done this with the old parser)
- Should be more efficient

It's a fair amount of code to add, but I feel better about having it here instead of hiding it in a crate that nobody else uses and we don't control. I hope you agree.


ducaale (approved):
LGTM

PR #208, opened 2021-12-02
Fix Clippy lints for 1.57

blyxxyz:


ducaale (approved):

.github/workflows/ci.yaml
```diff
@@ -67,10 +67,10 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --tests -- -D warnings
+          args: --tests -- -D warnings -A unknown-lints
```
ducaale, line 70: Is ignoring unknown lint rules the best approach? A typo in the lint name would be ignored without any warnings.

blyxxyz: It'd still be visible when running clippy manually, hopefully that'd be enough.

ducaale (approved):

PR #209, opened 2021-12-03
Expand tilde in request items that contain a path

### LINKED ISSUE #189: Expand tilde to home directory
On Linux, the following works in HTPPie:

`http -f POST example.com/files @~/.profile`

But when attempting `xh -f POST example.com/files @~/.profile`, I get "xh: error: No such file or directory (os error 2)".

It turns out the tilde is not being interpreted, and it works if I expand it manually: `xh -f POST example.com/files @/home/user/.profile`. I can also use `$HOME` so that my shell expands it for me, but I much prefer the brevity of `~`.

ducaale:
I have opted not to do the same for paths passed as an argument to flags such as `--session` and `--cert` as Bash and [Powershell 7.1](https://docs.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.2#psnativepspathresolution) both support expanding tilde in paths. Also, see https://stackoverflow.com/a/30027501/5915221.  

Resolves #189

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -59,6 +59,11 @@ pub fn get_home_dir() -> Option<PathBuf> {
     dirs::home_dir()
 }
 
+/// Perform simple tilde expansion if `dirs::home_dir()` is `Some(path)`.
+///
+/// Note that prefixed tilde e.g `~foo` is ignored.
+///
+/// See https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html
 pub fn expand_tilde(path: impl AsRef<Path>) -> PathBuf {
     if let Ok(path) = path.as_ref().strip_prefix("~") {
         let mut expanded_path = PathBuf::new();
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,6 +1,6 @@
 use std::env::var_os;
 use std::io::{self, Write};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -59,6 +59,17 @@ pub fn get_home_dir() -> Option<PathBuf> {
     dirs::home_dir()
 }
 
+pub fn expand_tilde(path: impl AsRef<Path>) -> PathBuf {
```
blyxxyz, line 62: Can you add a note that it doesn't implement `~user/` syntax?

tests/cli.rs
```diff
@@ -2707,3 +2707,40 @@ fn encoding_detection() {
     // (even for non-ASCII-compatible encodings)
     case("text/plain; charset=UTF-16", "\0\0", BINARY_SUPPRESSOR);
 }
+
+#[test]
+fn tilde_expanded_in_request_items() {
+    let homedir = TempDir::new().unwrap();
+
+    std::fs::write(homedir.path().join("secret_key.txt"), "sxemfalm.....").unwrap();
```
blyxxyz, line 2715: Can you look for these contents in the output so we know we grabbed the right file?

blyxxyz (changes requested):
Looks good.

> I have opted not to do the same for paths passed as an argument to flags

I think that's reasonable.

blyxxyz (approved):

PR #211, opened 2021-12-09
Do not make the implementation of server::http() generic

blyxxyz:
This makes rebuilding the tests much quicker. It now takes under 5 seconds, and I saw 20 or sometimes 60 seconds before.


ducaale (approved):
Looks good

Issue #212, opened 2021-12-19
aarch64 / arm64 support


myhro:
Hi,

I was using `xh` on a Raspberry Pi 4 which was recently switched to the [Raspberry Pi OS arm64 release][1]. After that, the `arm-unknown-linux-gnueabihf` release doesn't work anymore:

```
$ uname -a
Linux raspberrypi 5.10.63-v8+ #1488 SMP PREEMPT Thu Nov 18 16:16:16 GMT 2021 aarch64 GNU/Linux
$ cd xh-v0.14.1-arm-unknown-linux-gnueabihf/
$ ls -lh
total 6.1M
drwxr-xr-x 2 pi pi 4.0K Nov 26 20:35 completions
drwxr-xr-x 2 pi pi 4.0K Nov 26 20:35 doc
-rw-r--r-- 1 pi pi 1.1K Nov 26 20:35 LICENSE
-rw-r--r-- 1 pi pi 9.7K Nov 26 20:35 README.md
-rwxr-xr-x 1 pi pi 6.1M Nov 26 20:35 xh
$ ./xh
bash: ./xh: No such file or directory
```

Would it be possible to also release `aarch64`/`arm64` binaries?

Regards,
Tiago.

[1]: http://downloads.raspberrypi.org/raspios_arm64/images/

jihchi:
Hi @myhro , 

It looks like you haven't added "execute" permission to `xh`, for example: `chmod +x ./xh`, is this the case?

myhro:
Hi @jihchi,

Thank you for the suggestion. Unfortunately that's not the case.

As you can see in the output copied above, the `xh` binary has `-rwxr-xr-x` (`755`) permissions.

blyxxyz:
I think the problem is that the executable is dynamically linked:
```console
$ file xh
xh: ELF 32-bit LSB pie executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, stripped
```

I don't have a system to test, but these commands might fix it:
```
sudo dpkg --add-architecture armhf
sudo apt update
sudo apt install libc6:armhf
```

It would be nice to not have to do that. We may be able to manage that by building with musl (`arm-unknown-linux-musleabihf`) so the binary is statically linked.

Starting from 2023 ARM processors apparently won't have 32-bit support at all, so by that point we will need a separate 64-bit build. But for now I think we can still get by with just the one ARM build.

[...]

PR #213, opened 2021-12-20
Add aarch64 support

myhro:
Discussed in #212.

~Still trying to figure why tests are failing with multiple errors. It's not clear which file or directory wasn't found.~

~Was able to reproduce it locally with `cross test --target aarch64-unknown-linux-gnu`. Interestingly enough, switching to `aarch64-unknown-linux-musl` works.~

Edit: fixed switching to the expected `aarch64-unknown-linux-musl`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,6 +26,10 @@ jobs:
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
             use-cross: true
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
+            use-cross: true
+            flags: -- --test-threads=4
           - target: arm-unknown-linux-gnueabihf
             os: ubuntu-latest
             use-cross: true
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,6 +23,10 @@ jobs:
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
             use-cross: true
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
+            use-cross: true
+            flags: -- --test-threads=4
           - target: arm-unknown-linux-gnueabihf
             os: ubuntu-latest
             use-cross: true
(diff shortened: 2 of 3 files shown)


ducaale:
This happens because [we invoke the xh binary with the value of `CARGO_TARGET_*_RUNNER`](https://github.com/ducaale/xh/pull/190#issuecomment-968932013). But for some reason, that value happens to be `/linux-runner aarch64` in `aarch64-unknown-linux-gnu` rather than the usual `qemu-*`.  To make the tests pass, you can update the `find_runner()` function in `tests/cli.rs` to be like this: 

```rust
fn find_runner() -> Option<String> {
    for (key, value) in std::env::vars() {
        if key.starts_with("CARGO_TARGET_") && key.ends_with("_RUNNER") && value.contains("qemu") {
            return Some(value);
        }
    }
    None
}
```

However, I am still in favor of adding `aarch64-unknown-linux-musl` and not `aarch64-unknown-linux-gnu` which doesn't work in termux.

myhro:
> This happens because [we are invoking the xh binary with the value of `CARGO_TARGET_*_RUNNER`](https://github.com/ducaale/xh/pull/190#issuecomment-968932013). But for some reason, this value happens to be `/linux-runner aarch64` in `aarch64-unknown-linux-gnu` rather than the usual `qemu-*`.

Thanks for the explanation. Looking at the `rustembedded/cross` images, I've noticed that their `CARGO_TARGET_*_RUNNER` environment variables are quite different. In [`Dockerfile.aarch64-unknown-linux-gnu`][1] it's `"/linux-runner aarch64"` and on [`Dockerfile.aarch64-unknown-linux-musl`][2] it's just `qemu-aarch64`. That explains why tests ran fine on the latter.

> However, I am still in favor of adding `aarch64-unknown-linux-musl` and not `aarch64-unknown-linux-gnu` which doesn't work in termux.

Sorry, I misread [your suggestion in this comment][3] (and also in the other one before). I've been mentioning the `-gnu` variant for `aarch64` so many times in the discussion that I didn't notice you were suggesting the `-musl` one. I'm all in for the musl statically-linked option, like we are already doing with the `x86_64` release.

I've updated the PR to make use of the `aarch64-unknown-linux-musl` one. Thanks for the clarification.

[1]: https://github.com/rust-embedded/cross/blob/2f1ef07/docker/Dockerfile.aarch64-unknown-linux-gnu#L28
[2]: https://github.com/rust-embedded/cross/blob/2f1ef07/docker/Dockerfile.aarch64-unknown-linux-musl#L27
[3]: https://github.com/ducaale/xh/issues/212#issuecomment-997446288

ducaale (approved):
Looks good, thanks!

nicolasdanelon:
what about apple silicon 😭

ducaale:
@nicolasdanelon can you open an issue for it? Thanks

nicolasdanelon:
@ducaale sure.

nicolasdanelon:
@ducaale it's working now man! 😮

PR #188, opened 2021-11-08
Add feature status to version

sorairolake:
Closes #185

PR #214, opened 2021-12-23
Get version from -V when using help2man

ducaale:
Since PR https://github.com/ducaale/xh/pull/188 was merged, the output from the `-version` flag was unsuitable for use with help2man.

Additionally, I have replaced the usage of `xh --version | head -1` with just `xh -V` in `install.sh`.


blyxxyz (approved):
I don't love the `-V` flag, there's a chance HTTPie starts using it for something.

But it should be fairly easy to stop using it later. `help2man` has a `--version-string` option.

PR #216, opened 2021-12-24
Replace Structopt with Clap v3

ducaale:
* Removed the manual `NO_COLOR` check.
* Removed fish completions workarounds.
* Removed triple `{n}` at the end of each long help.
* Removed case insensitivity for flags that take an enum.
* hide `ssl2.3` option for `--ssl` and `1` for `--http-version`.
* Updated custom clap_error messages to account for `USAGE` string. See https://github.com/clap-rs/clap/issues/3227#issuecomment-1003183572.


src/cli.rs
```diff
@@ -386,39 +380,39 @@ impl Cli {
...
                 // --help is baked into clap. So we intercept its special error that
                 // would print long help and print short help instead. And if we do
-                // want to print long help, then we insert our own error in from_iter_safe
+                // want to print long help, then we insert our own error in try_parse_from
                 // with a special tag.
                 if env::var_os("XH_HELP2MAN").is_some() {
-                    Cli::clap()
-                        .template(
+                    Self::into_app()
+                        .help_template(
```
blyxxyz, line 415: This template needs to be updated:
```console
$ XH_HELP2MAN=1 cargo run help
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/xh help`
thread 'main' panicked at '`{flags}` template variable was removed in clap3, they are now included in `{options}`'
```

blyxxyz (changes requested):
I like how some of our custom logic is handled by clap now.

src/cli.rs
```diff
@@ -34,57 +33,50 @@ use crate::utils::config_dir;
...
     pub multipart: bool,
 
     /// Pass raw request data without extra processing.
-    #[structopt(long, value_name = "RAW")]
+    #[clap(long, value_name = "RAW")]
     pub raw: Option<String>,
 
     /// Controls output processing.
-    #[structopt(long, possible_values = &Pretty::variants(), case_insensitive = true, value_name = "STYLE")]
+    #[clap(long, arg_enum, value_name = "STYLE")]
```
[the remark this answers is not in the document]
ducaale: Yes, I removed case-insensitivity to align with HTTPie's behavior.

>- Removed case insensitivity for flags that take an enum.

ducaale:
Hmm, it seems https://github.com/clap-rs/clap/pull/3232 didn't completely fix https://github.com/clap-rs/clap/issues/3215
```
$ cargo run -- --help
xh 0.14.1
-native-tls
xh is a friendly and fast tool for sending HTTP requests

USAGE:
    xh [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...

ARGS:
    <[METHOD] URL>       The request URL, preceded by an optional HTTP method
    <REQUEST_ITEM>...    Optional key-value pairs to be included in the request
```

blyxxyz:
It's outside the scope of this PR, but [custom help headings](https://docs.rs/clap/latest/clap/struct.Arg.html#method.help_heading) may be interesting.

ducaale:
>It's outside the scope of this PR, but custom help headings may be interesting.

Looks interesting. 
* Columns from different sections are not aligned. But that might change in a future version of clap. See https://github.com/clap-rs/clap/issues/1013 
* Are help headings are needed in short help?

```
xh 0.14.1
-native-tls
xh is a friendly and fast tool for sending HTTP requests

USAGE:
    xh [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...

ARGS:
    <[METHOD] URL>       The request URL, preceded by an optional HTTP method
    <REQUEST_ITEM>...    Optional key-value pairs to be included in the request

OPTIONS:
        --native-tls      Use the system TLS library instead of rustls (if enabled at compile time)
        --https           Make HTTPS requests if not specified in the URL
    -I, --ignore-stdin    Do not attempt to read stdin
        --curl            Print a translation to a `curl` command
        --curl-long       Use the long versions of curl's flags
        --help            Print help information
    -V, --version         Print version information

Predefined Content Types:
    -j, --json         (default) Serialize data items from the command line as a JSON object
    -f, --form         Serialize data items from the command line as form fields
    -m, --multipart    Like --form, but force a multipart/form-data request even without files
        --raw <RAW>    Pass raw request data without extra processing

Output Processing:
        --pretty <STYLE>                 Controls output processing [possible values: all, colors, format, none]
    -s, --style <THEME>                  Output coloring style [possible values: auto, solarized, monokai, fruity]
        --response-charset <ENCODING>    Override the response encoding for terminal display purposes
        --response-mime <MIME_TYPE>      Override the response mime type for coloring and formatting for the terminal

Output Options:
    -p, --print <FORMAT>            String specifying what the output should contain
    -h, --headers                   Print only the response headers. Shortcut for --print=h
    -b, --body                      Print only the response body. Shortcut for --print=b
    -v, --verbose                   Print the whole request as well as the response
        --all                       Show any intermediary requests/responses while following redirects with --follow
    -P, --history-print <FORMAT>    The same as --print but applies only to intermediary requests/responses
    -q, --quiet                     Do not print to stdout or stderr
    -S, --stream                    Always stream the response body
    -o, --output <FILE>             Save output to FILE instead of stdout
    -d, --download                  Download the body to a file instead of printing it
    -c, --continue                  Resume an interrupted download. Requires --download and --output

Sessions:
        --session <FILE>              Create, or reuse and update a session
        --session-read-only <FILE>    Create or read a session without updating it form the request/response exchange

Authentication:
    -A, --auth-type <AUTH_TYPE>         Specify the auth mechanism [possible values: basic, bearer, digest]
    -a, --auth <USER[:PASS] | TOKEN>    Authenticate as USER with PASS or with TOKEN
        --ignore-netrc                  Do not use credentials from .netrc

Network:
        --offline                   Construct HTTP requests without sending them anywhere
        --check-status              (default) Exit with an error status code if the server replies with an error
    -F, --follow                    Do follow redirects
        --max-redirects <NUM>       Number of redirects to follow, only respected if `follow` is set
        --timeout <SEC>             Connection timeout of the request
        --proxy <PROTOCOL:URL>      Use a proxy for a protocol. For example: `--proxy https:http://proxy.host:8080`
        --http-version <VERSION>    HTTP version to use [possible values: 1.0, 1.1, 2]

SSL:
        --verify <VERIFY>    If "no", skip SSL verification. If a file path, use it as a CA bundle
        --cert <FILE>        Use a client side certificate for SSL
        --cert-key <FILE>    A private key file to use with --cert
        --ssl <VERSION>      Force a particular TLS version [possible values: auto, tls1, tls1.1, tls1.2, tls1.3]

Each option can be reset with a --no-OPTION argument.


Run `xh help` for more complete documentation.
```

src/request_items.rs
```diff
@@ -128,9 +130,12 @@ impl FromStr for RequestItem {
             // TODO: We can also end up here if the method couldn't be parsed
             // and was interpreted as a URL, making the actual URL a request
             // item
-            Err(clap::Error::with_description(
-                &format!("{:?} is not a valid request item", request_item),
+            Err(clap::Error::raw(
                 clap::ErrorKind::InvalidValue,
+                format!(
+                    "Invalid value for '[REQUEST_ITEM]...': {:?} is not a valid Request Item",
```
blyxxyz, line 136:
```suggestion
                    "Invalid value for '[REQUEST_ITEM]...': {:?}",
```

blyxxyz (approved):
Looks good!

PR #220, opened 2022-01-12
Add Powershell one-liner for Windows

ChrisK-0:
Added Powershell one-liner, that uses ducaale repository with new install.ps1 file to install xh.

1. Installs the zip from which to extract the latest xh.exe from
2. Checks if xh.exe exists in $destDir folder (Users\"user"\bin). If not, extract it there, if it exists, deletes older one and places the newest xh.exe there.
3. Adds $destDir (Users\"user"\bin) to Environment variables PATH, if ...\bin\ is not already there.
4. Checks if VS C++ Runtime Redistributable 2015 or higher (for vscruntime140.dll file) is installed. If 2015 or higher year version is installed, skip, if not, installs the 2015-2022 with vc_redist.x64.exe ( https://aka.ms/vs/17/release/vc_redist.x64.exe ).


install.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+#extraction
+$entries | foreach {[IO.Compression.ZipFileExtensions]::ExtractToFile( $_, $destDir + $_.Name) }
+
+#free zipfile
+$zip.Dispose()
+
+Remove-Item -Path $zipFile
+
+Write-Host "`n Extracted 'xh.exe' file to $destDir"
```
ducaale, line 38: xh supports [xhs symlink](https://github.com/ducaale/xh#xh-and-xhs), which will default to HTTPS requests. Can you create that here?

install.ps1
```diff
@@ -0,0 +1,107 @@
...
+	$VSRinstallerprocess
+
+	# Delete installer from downloaded path after installer process is done
+	$VSRinstallerprocess.WaitForExit()
+	Remove-Item -Path $outpath -Force -Recurse
+}
+
+Write-Output "`n Done!"
+Write-Host -NoNewLine "`n Press any key to continue...";
+$null = $Host.UI.RawUI.ReadKey('NoEcho,IncludeKeyDown');
```
ducaale, line 107: Is it necessary to ask the user to press a key here?

README.md
```diff
@@ -16,6 +16,13 @@ on improved performance.
 curl -sfL https://raw.githubusercontent.com/ducaale/xh/master/install.sh | sh
 ```
 
+### via Powershell (Windows)
+
+```
+  PowerShell -ExecutionPolicy bypass -Command "iwr -useb https://raw.githubusercontent.com/ducaale/xh/master/install.ps1 | iex"
```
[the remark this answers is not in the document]
ChrisK-0: It is not. I tested this on a virtual machine with clean windows for prompts. 
It can be removed, so: `PowerShell "iwr -useb https://raw.githubusercontent.com/ducaale/xh/master/install.ps1 | iex"`

install.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+	$p += ";$destDir";
+	[System.Environment]::SetEnvironmentVariable('Path',$p,[System.EnvironmentVariableTarget]::User);
+
+	$Env:Path = [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine) + ";" + $p
+
+	Write-Host "`n Modified PATH environment variable. Restart programs which use command line." -Foreground yellow
+}
+
+# visual redistributor version check for vcruntime140.dll for xh
```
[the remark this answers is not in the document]
ChrisK-0: Removed VS C++ Runtime Redistributable checking and installing.

install.ps1
```diff
@@ -0,0 +1,107 @@
...
+Write-Output "Downloading: $($asset.name)"
+Invoke-RestMethod -Method Get -Uri $asset.browser_download_url -OutFile $zipFile
+
+# checks if an older version of xh.exe exists in '$destDir', if yes, then delete it, if not, then download latest zip to extract from
+$xhPath = "${destDir}xh.exe"
+if (Test-Path $xhPath -PathType Leaf) {
+	"`n xh.exe exists in $destDir, deleting..."
+	rm -r -fo $xhPath
+} else {
+	"`n ${destDir}xh.exe does not exist"
```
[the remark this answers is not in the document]
ChrisK-0: Removed it.

install.ps1
```diff
@@ -0,0 +1,107 @@
...
+}
+
+#xh.exe extraction start
+Add-Type -Assembly System.IO.Compression.FileSystem
+
+$zip = [IO.Compression.ZipFile]::OpenRead($zipFile)
+$entries=$zip.Entries | Where {$_.FullName -Like '*.exe'}
+
+#create dir for result of extraction
+New-Item -ItemType Directory -Path $destDir -Force
```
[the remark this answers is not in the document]
ChrisK-0: Yes. Adding "| Out-Null" at the end of 28 (New-Item) gets rid of that output.

install.ps1
ChrisK-0: Sorry about the late reply to all comments. This one is a tricky one. 
1. Powershell needs to be run as an administrator for symbolic link creation. 
2. Not gonna lie, I'm not sure if its working. I added an image to this comment to get confirmation

mstsc_4RvYD2bUhn
.

install.ps1
ChrisK-0: It is not. It was so the user can read what happened and close the window with a key press. This could be run from the Run window (Win+R), but for the symbolic link Powershell must be run as an administrator.

install.ps1
blyxxyz: Maybe it would be better to just copy the executable?

A shortcut [could also work](https://stackoverflow.com/a/30921867), but the user would have to type out `xhs.lnk` (unless we modify `PATHEXT` but that seems too intrusive).

What do the package managers do?

ducaale: Scoop copies the binary (I think). See https://github.com/ScoopInstaller/Main/blob/5180fcd99cc1c1453302da1d0402e1bd7f3362bf/bucket/xh.json#L17.

>Not gonna lie, I'm not sure if its working. I added an image to this comment to get confirmation

Inked149366275-cc883289-6589-4d36-9a68-2037fb0d2125

README.md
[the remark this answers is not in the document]
ducaale: Can we make the command even shorter?

```
# Deno
iwr https://deno.land/install.ps1 -useb | iex

# Scoop
iwr -useb get.scoop.sh | iex

# Chocolatey
iwr https://chocolatey.org/install.ps1 -UseBasicParsing | iex
```

install.ps1
ChrisK-0: I used `cmd /c mklink /d xhs $xhPath` in install.ps1. It launches cmd and uses cmd's mklink command, because powershell's junction did not work for me. Sending a new pull request with the updated powershell file, if I succeed. If not il try to check back tomorrow (16 jan).

install.ps1
```diff
@@ -0,0 +1,56 @@
...
+$entries | foreach {[IO.Compression.ZipFileExtensions]::ExtractToFile( $_, $destdir + $_.Name) }
+#free zipfile
+$zip.Dispose()
+
+Remove-Item -Path $zipfile
+
+Write-Host "`n Exctracted 'xh.exe' file is located in: $destdir."
+
+# Requires powershell to be run as administrator. Creates 'xhs' symbolic link
+cmd /c mklink /d xhs $xhPath
```
ducaale, line 38: I think you meant to use `/H` instead `/d`, which creates a soft link pointing to a directory?
```
cmd /c mklink /H xhs.exe $xhPath
```

However, requiring administrator mode might be inconvenient for users, so how about copying the binary and renaming that to `xhs.exe` instead?

README.md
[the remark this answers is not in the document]
ChrisK-0: So it is still required to have scoop?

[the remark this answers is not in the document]
ducaale: No, I was just listing some existing install instructions that don't start with `PowerShell -ExecutionPolicy bypass -Command`.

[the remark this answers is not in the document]
ChrisK-0: Powershell part is indeed only required when running command from Run and the command to run in Powershell is between quotes. In Powershell it is not needed.

install.ps1
ChrisK-0: If I understood correctly, you meant to copy xh.exe as xhs.exe? Created latest commit with it and it worked on a clean windows virtual machine. Image included. Also removed the press any key at the end.

It is incredibly fast now so all the Write-Host and other comments it sends to Powershell are now pointless. I did not remove them though.

image

README.md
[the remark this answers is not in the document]
ducaale: Let's assume the user will run the command from Powershell.

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+Write-Output "Downloading: $($asset.name)"
+Invoke-RestMethod -Method Get -Uri $asset.browser_download_url -OutFile $zipfile
+
+# checks if an older version of xh.exe (includes xhs.exe) exists in '$destdir', if yes, then delete it, if not, then download latest zip to extract from
+$xhPath = "${destdir}xh.exe"
+$xhsPath = "${destdir}xhs.exe"
+if (Test-Path -Path $xhPath -PathType Leaf) {
+    "`n xh.exe exists in $destdir, deleting xh and xhs"
+	rm -r -fo $xhPath
+	rm -r -fo $xhsPath
```
ducaale, line 18: Can you replace all tabs with four spaces?

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#xh.exe extraction start
```
ducaale, line 21: Can you add space between `#` and comments?

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+Add-Type -Assembly System.IO.Compression.FileSystem
+
+$zip = [IO.Compression.ZipFile]::OpenRead($zipfile)
+$entries=$zip.Entries | where {$_.FullName -like '*.exe'}
```
ducaale, line 25: Can you add space around the assignment operator?

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+$zipfile = "$env:TEMP\$($asset.name)"
+
+Write-Output "Downloading: $($asset.name)"
+Invoke-RestMethod -Method Get -Uri $asset.browser_download_url -OutFile $zipfile
+
+# checks if an older version of xh.exe (includes xhs.exe) exists in '$destdir', if yes, then delete it, if not, then download latest zip to extract from
+$xhPath = "${destdir}xh.exe"
+$xhsPath = "${destdir}xhs.exe"
+if (Test-Path -Path $xhPath -PathType Leaf) {
+    "`n xh.exe exists in $destdir, deleting xh and xhs"
```
ducaale, line 16: Can you remove the leading newline + leading space when outputting a message?

ducaale, line 16: How about something like this?
```
Removing previous installation of xh from $destdir
```

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+$entries | foreach {[IO.Compression.ZipFileExtensions]::ExtractToFile( $_, $destdir + $_.Name) }
+#free zipfile
+$zip.Dispose()
+
+Remove-Item -Path $zipfile
+
+# Copy xh.exe as xhs.exe into bin
+Copy-Item $xhPath $xhsPath
+
+Write-Host "`n Exctracted 'xh.exe' file is located in: $destdir."
```
ducaale, line 40: I think we defer printing the install location to the end of the script. For reference, this is what `install.sh` prints at the end:

```
xh v0.14.1 has been installed to:
 • ~/.local/bin/xh
 • ~/.local/bin/xhs
```

You don't have to print the same thing but it would be good to at least print the version installed + `$destdir`.

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+	Write-Output "`n Adding $destdir to your Path"
+	
+	$p += "$destdir";
+	[System.Environment]::SetEnvironmentVariable('Path',$p,[System.EnvironmentVariableTarget]::User);
+	
+	$Env:Path = [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine) + ";" + $p
+	
+	# Path to xhs.exe
+	
+	Write-Host "`n PATH environment variable changed (restart your applications that use command line)." -foreground yellow
```
ducaale, line 56: Can you double-check if restarting Powershell is required? It wasn't the case for me, at least.

install.ps1
```diff
@@ -0,0 +1,59 @@
...
+
+# Add to environment variables
+$p = [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::User);
+if (!$p.ToLower().Contains($destdir.ToLower()))
+{
+	# path to "user"/bin
+	Write-Output "`n Adding $destdir to your Path"
+	
+	$p += "$destdir";
+	[System.Environment]::SetEnvironmentVariable('Path',$p,[System.EnvironmentVariableTarget]::User);
```
ducaale, line 50: Can you add spaces after each comma?

ducaale (changes requested):
Looks good! Just one more round of nitpicking

install.ps1
```diff
@@ -0,0 +1,59 @@
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
+$ProgressPreference = 'SilentlyContinue'
+$release = Invoke-RestMethod -Method Get -Uri "https://api.github.com/repos/ducaale/xh/releases/latest"
+$asset = $release.assets | Where-Object name -like *.zip
```
blyxxyz, line 5: Looking for a zip file seems fragile, would it be better to look for `x86_64-pc-windows` in the name instead?

install.ps1
ChrisK-0: Was for me. Safer to warn about it, but if you insist we could remove it.

image

install.ps1
ChrisK-0: It very much is. Searching by `x86_64-pc-windows` now.  Made a new commit with updated readme and install files.

install.ps1
ducaale: I see. In that case, let's keep the warning.

ducaale:
There is a [PowerShell extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.PowerShell) for VS Code that can both lint and format PowerShell scripts. @ChrisK-0 Would you like to fix some of the warnings it reports + use the formatter before merging this PR?

image

install.ps1
```diff
@@ -0,0 +1,58 @@
...
+    [System.Environment]::SetEnvironmentVariable('Path', $p, [System.EnvironmentVariableTarget]::User)
+    $Env:Path = [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine) + ";" + $p
+	
+    # Path to xhs.exe
+    Write-Host "PATH environment variable changed (restart your applications that use command line)." -foreground yellow
+}
+
+# Get version from zip file name
+
+Write-Output "xh v$($zipfilename.trim("xh-v -x86_64-pc-windows-msvc.zip")) has been installed to:`n - $xhPath`n - $xhsPath"
```
ducaale, line 58: Can we move `$zipfilename.trim("xh-v -x86_64-pc-windows-msvc.zip")` to its own variable? It would also be good to add `EOL` or you can just let the [PowerShell extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.PowerShell) take care of it.

ChrisK-0: Sure, I'l see to it.

install.ps1
```diff
@@ -0,0 +1,69 @@
...
+	
+    # Path to xhs.exe.
+
+    Write-Host "PATH environment variable changed (restart your applications that use command line)." -foreground yellow
+}
+
+# Get version from zip file name.
+
+$xhVersion = $($zipfilename.trim("xh-v -x86_64-pc-windows-msvc.zip"))
+Write-Output "xh v$($xhVersion) has been installed to:`n - $xhPath`n - $xhsPath"
```
ducaale, line 69: Can you add a new line at the end of this file?

ducaale (approved):
Approved 🎉🎉🎉

Thank you, @ChrisK-0!

PR #221, opened 2022-01-16
Statically link C-runtime for MSVC Windows

### LINKED ISSUE #219: Oneliner to install xh under Windows
Httpie is much more complicated  to install under Windows than under Linux and Mac. Xh is somewhat easier but it still involves many manual steps compared to Linux and Mac (downloading the release, extracting it, putting the executable into PATH, installing the VS C++ Redistributable).

It would be super awesome if there was a oneliner command to install under Windows as well.

ducaale:
On windows, xh requires `VS C++ Runtime Redistributable` to be installed or will fail with `The code execution cannot proceed because VCRUNTIME140.dll was not found`. This PR changes that by statically linking the C-runtime.

Also, see https://github.com/BurntSushi/ripgrep/pull/1613

Partially resolves #219

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -69,16 +69,21 @@ jobs:
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             flags: --features=native-tls
+            rustflags: -C target-feature=+crt-static
     steps:
       - uses: actions/checkout@v2
+
+      - name: Set RUSTFLAGS env variable
+        if: matrix.job.rustflags
+        shell: bash
+        run: echo "RUSTFLAGS=${{ matrix.job.rustflags }}" >> $GITHUB_ENV
+
       - name: Build target
         uses: actions-rs/cargo@v1
         with:
           use-cross: ${{ matrix.job.use-cross }}
           command: build
           args: --release --target ${{ matrix.job.target }} ${{ matrix.job.flags }}
-        env:
-          RUSTFLAGS: ${{ matrix.job.os == 'windows-latest' && '-C target-feature=+crt-static' || '' }}
 
       - name: Strip release binary (linux and macOS)
         if: matrix.job.os != 'windows-latest'
@@ -92,6 +97,7 @@ jobs:
 
       - id: get_version
         uses: battila7/get-version-action@v2
+
       - name: Package
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


blyxxyz:
Hm, is this really something we want to specify for the whole project? Maybe it should be limited to the release action.

ducaale:
The other option would be to use the `RUSTFLAGS` environment variable. I will see how I can do that in `.github/workflows/release.yaml`.

.github/workflows/release.yaml
```diff
@@ -77,6 +77,8 @@ jobs:
           use-cross: ${{ matrix.job.use-cross }}
           command: build
           args: --release --target ${{ matrix.job.target }} ${{ matrix.job.flags }}
+        env:
+          RUSTFLAGS: ${{ matrix.job.os == 'windows-latest' && '-C target-feature=+crt-static' || '' }}
```
ducaale, line 81: Should I set `RUSTFLAGS` in other places as well?

blyxxyz:
Can you run the release action somewhere for a binary we can test with?

ducaale:
Sure, I have created https://github.com/ducaale/xh-temp-111 for testing purposes. I'll grant you push access in a moment.

.github/workflows/release.yaml
```diff
@@ -69,8 +69,15 @@ jobs:
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             flags: --features=native-tls
+            rustflags: -C target-feature=+crt-static
     steps:
       - uses: actions/checkout@v2
+
+      - name: Set RUSTFLAGS env variable
+        if: matrix.job.rustflags
+        shell: bash
```
ducaale, line 78: Unless I set `shell` to `bash`, `RUSTFLAGS` will not get passed to cargo in the next step.

blyxxyz:
Should we perhaps also move LTO from `Cargo.toml` to the release CI? An incremental rebuild takes 10s without it and 1m30s with, that's annoying when benchmarking.

ducaale:
I will need to add `Cross.toml` file so `cross` can be configured to pass `RUSTFLAGS` to the underlying build environment.

blyxxyz (approved):
Looks good! Unlike v0.14.1 it just works on a clean Windows installation.

Issue #224, opened 2022-01-19
UnknownIssuer error with private CA


austinbutler:
With a private CA chain added to `/etc/pki/tls/certs/ca-bundle.crt` I get `error trying to connect: invalid certificate: UnknownIssuer`. The same request works with `curl` or `httpie`. If I use the `--native-tls` flag with `xh` it works.

Without being all that familiar with `rustls`, two potential suggestions:
1. Better documentation around how SSL certs are found/used?
2. Integrate [rustls-native-certs](https://github.com/rustls/rustls-native-certs) so that system certificate store just works?

ducaale:
By default, the reqwest library uses `webpki-roots` but can also integrate with `rustls-native-certs` via the `rustls-native-certs` feature.

@austinbutler can you try compiling `xh` after [adding both](https://github.com/seanmonstar/reqwest/pull/1058#issuecomment-710417994) `rustls-tls-webpki-roots` and `rustls-tls-native-roots` to reqwests's enabled features? thanks

```diff
[dependencies.reqwest]
version = "0.11.6"
default-features = false
- features = ["rustls-tls", "json", "gzip", "brotli", "deflate", "multipart", "blocking", "socks", "cookies"]
+ features = ["rustls-tls", "rustls-tls-webpki-roots", "rustls-tls-native-roots", "json", "gzip", "brotli", "deflate", "multipart", "blocking", "socks", "cookies"]
```

austinbutler:
Yep that seems to do it. Should I open a PR?

ducaale:
Yes, that would be very much appreciated!

ducaale:
This is now fixed in xh v0.15.0

PR #225, opened 2022-01-21
Enable using OS certificate store with rustls

austinbutler:
Closes https://github.com/ducaale/xh/issues/224.

I ran `cargo update` since I didn't know the command to just update the lockfile after changing `Cargo.toml`, but it worked and created no additional test failures so I've left that. If you want the smaller set of `Cargo.lock` changes let me know.

Two tests fail, but they also fail for me on master:
- `basic_auth_from_session_is_used`
- `bearer_auth_from_session_is_used`

Is there any additional setup required for the tests?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
+/.direnv
+/.envrc
 /target
 /.vscode
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1320,6 +1320,7 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "rustls",
+ "rustls-native-certs",
  "serde",
  "serde_json",
  "serde_urlencoded",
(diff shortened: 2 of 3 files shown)


blyxxyz:
Can you target `develop` instead of `master`?

That might get rid of your local test failures too, I think I remember netrc files affecting those tests before #207.

> I ran `cargo update` since I didn't know the command to just update the lockfile after changing `Cargo.toml`, but it worked and created no additional test failures so I've left that.

The lockfile is automatically updated after a rebuild (e.g. `cargo test`, `cargo check`, etc), so you shouldn't have to take any special action.

> If you want the smaller set of `Cargo.lock` changes let me know.

Running `cargo update` every once in a while is good practice, but in this case it seems to have messed with our minimum version requirement, so could you try the smaller set?

austinbutler:
> Can you target develop instead of master?

Done, and that did resolve the test failures. Also removed the `cargo update` changes.

> The lockfile is automatically updated after a rebuild

TIL!

blyxxyz (approved):
Thank you!

PR #226, opened 2022-01-21
Default to a different config_dir during tests

ducaale:
Otherwise, a user's local config can interfere with the integration tests.

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -83,6 +83,7 @@ fn get_base_command() -> Command {
     }
     cmd.env("HOME", "");
     cmd.env("NETRC", "");
+    cmd.env("XH_CONFIG_DIR", "");
     #[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


blyxxyz:
It might be cleaner to set `XH_CONFIG_DIR` in `get_base_command()`.

tests/cli.rs
```diff
@@ -83,6 +83,12 @@ fn get_base_command() -> Command {
     }
     cmd.env("HOME", "");
     cmd.env("NETRC", "");
+    cmd.env(
+        "XH_CONFIG_DIR",
+        dirs::config_dir()
```
blyxxyz, line 88: Would just blanking it like the others work? Or is looking at `~/.config/xh-test` useful for something?

ducaale: Ah, I see what you mean. Let me check if that works...

blyxxyz (approved):
LGTM

PR #232, opened 2022-02-12
Strip leading `://` from URLs

ducaale:
See https://github.com/httpie/httpie/issues/1195


blyxxyz:
I'd do `let url = url.strip_prefix("://”).or(url)`

Maybe with a comment explaining the feature?

PR #233, opened 2022-02-14
Switch to full buffering and manual flushing

blyxxyz:
All output is now fully buffered. That means it might not show up
until `.flush()` is called, so `.flush()` has to be called manually
pretty frequently. The general rule is that it should be called when
the next output might not be ready yet. For example, `.flush()` is
called at the end of `print_request_body()`, so the request is visible
before waiting for the response.

This gets trickiest for streaming responses. They have to flush after
receiving the next piece of response from the server, but not more
often than that (or it'd hurt performance). `BinaryGuard` and
`copy_largebuf` have been changed to make that possible.

But these are also where the greatest gains are made: A (piped)
`--stream --pretty format` JSON response is now ten times as
fast as before. That's admittedly a bit niche, but I think `--stream`
might now always be faster than not streaming. (I haven't benchmarked
it exhaustively yet.)

The `Buffer` type is overhauled, in part for unrelated
optimization. It turns out that it can matter a lot which specific
types from `termcolor` are used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,12 +53,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "assert_matches"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
-
-[[package]]
 name = "async-compression"
 version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2159,7 +2153,6 @@ version = "0.15.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
- "assert_matches",
  "atty",
  "chardetng",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,6 @@ features = ["parsing", "html", "yaml-load", "dump-load", "dump-create", "regex-o
 
 [dev-dependencies]
 assert_cmd = "2.0"
-assert_matches = "1.4.0"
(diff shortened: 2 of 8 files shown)


ducaale (approved):
Looks good. I like how the `BinaryGuard` can detect null bytes earlier compared to the previous implementation.

>But these are also where the greatest gains are made: A (piped)
--stream --pretty format JSON response is now ten times as
fast as before. That's admittedly a bit niche, but I think --stream
might now always be faster than not streaming. (I haven't benchmarked
it exhaustively yet.)

Do you think we're in a position to enable streaming by default?

blyxxyz:
> Do you think we're in a position to enable streaming by default?

The binary detection would have to become worse. (That might be ok.)

A response that's incorrectly marked as JSON currently has its formatting mangled, e.g.
```console
$ xh --print=b :/not.json
The quick brown fox jumps over the lazy dog

$ xh --print=b :/not.json --stream
Thequickbrownfoxjumpsoverthelazydog
```
I've been working on a fork of JSONXF with coloring and basic validation built in that could perhaps solve this by disabling the formatting midway through if something doesn't check out.
HTTPie in streaming mode assumes a single JSON document per line that it can validate before printing. Maybe we could borrow from that approach. (But if we matched it entirely then enabling streaming by default would not be an option because it'd break many normal JSON responses.)

Adding XML/HTML formatting could become more difficult. (This might be the biggest problem.)

If each line has its own JSON document then they'll only be formatted in streaming mode:
```console
$ xh --print=b :/multi.json
{"a": 1}
{"a": 1}

$ xh --print=b :/multi.json --stream
{
    "a": 1
}

{
    "a": 1
}
```
HTTPie does that too.
So streaming by default would change the output for responses like that. But I think that's ok.

PR #234, opened 2022-02-15
Automate the generation of negation flags

blyxxyz:
clap 3 lets us do this.

The only observable difference is that `--no-help` and `--no-version` exist now. Those are a bit useless, but HTTPie also has them, so that's okay.


src/cli.rs
```diff
@@ -553,13 +500,34 @@ impl Cli {
...
+        // because we also need to generate them right here.
+        // We could use Box::leak(), but this OnceCell maneuver keeps valgrind
+        // happy-ish.
+        // We assume that `get_arguments()` has a fixed iteration order.
+        static ARG_STORAGE: OnceCell<Vec<String>> = OnceCell::new();
+        let arg_storage = ARG_STORAGE.get_or_init(|| {
+            opts.iter()
+                .map(|opt| format!("--no-{}", opt.get_long().expect("long option")))
+                .collect()
+        });
```
ducaale, line 525: It would be great to mention this workaround in https://github.com/clap-rs/clap/issues/815

ducaale (approved):
Wow, I've never been more excited for a PR than this one!

PR #236, opened 2022-03-03
Display download's elapsed time as seconds

ducaale:
Since we're planning to support HTTPie's [`--meta` flag](https://httpie.io/docs/cli/response-meta) for displaying the response's elapsed time, I thought we should simplify how we display elapsed time everywhere.

#### Before:
```
Downloading 10.00MiB to "<stdout>"
Done. 10.00MiB in 4s 810ms 951us 800ns (2.08MiB/s)
```

#### After:
```
Downloading 10.00MiB to "<stdout>"
Done. 10.00MiB in 10.892807s (940.07KiB/s)
```


src/download.rs
```diff
@@ -249,11 +249,11 @@ pub fn download_file(
...
             pb.finish_and_clear();
             let time_taken = starting_time.elapsed();
-            if time_taken != Duration::from_nanos(0) {
+            if !time_taken.is_zero() {
                 eprintln!(
-                    "Done. {} in {} ({}/s)",
+                    "Done. {} in {}s ({}/s)",
                     HumanBytes(downloaded_length),
-                    humantime::format_duration(time_taken),
+                    time_taken.as_secs_f32(),
```
ducaale, line 256: I could have used `as_secs_f64()` here, but the resulting output would be too lengthy.

blyxxyz: To give the output the right length you can format it with `{:.5}` instead of `{}`. HTTPie [does the same](https://github.com/httpie/httpie/blob/c901e704633cb79c686b3edaae497ed38b2978ba/httpie/downloads.py#L32).

blyxxyz (approved):

PR #240, opened 2022-03-12
Support outputting the metadata of a response

ducaale:
This PR adds support for outputting response's elapsed time as part of a new metadata section which can be enabled via either `--print=m`, a double invocation of the verbose flag i.e `-vv`, or via the `--meta` flag (similar to `--headers` and `--body`).

The duration of a response's elapsed time begins from when a request is made and ends when the response has been fully printed. 

```
$ xh -vv httpbin.org/status/418

GET /status/418 HTTP/1.1
Accept: */*
Connection: keep-alive
Host: httpbin.org
User-Agent: xh/0.15.0

HTTP/1.1 418 I'm a teapot
Access-Control-Allow-Origin: *
Content-Length: 135
Date: Sat, 12 Mar 2022 21:30:13 GMT
Server: gunicorn/19.9.0
X-More-Info: http://tools.ietf.org/html/rfc2324


    -=[ teapot ]=-

       _...._
     .'  _ _ `.
    | ."` ^ `". _,
    \_;`"---"`|//
      |       ;/
      \_     _/
        `"""`


Elapsed time: 0.37325s

```

Ticks another item from #4


src/cli.rs
```diff
@@ -53,7 +53,7 @@ pub struct Cli {
     pub form: bool,
 
     /// Like --form, but force a multipart/form-data request even without files.
-    #[clap(short = 'm', long, overrides_with_all = &["json", "form"])]
+    #[clap(long, overrides_with_all = &["json", "form"])]
```
ducaale, line 56: This must have been an unintentional mistake since HTTPie never made use of the `-m` until now.

ducaale:
I need to add tests + update the documentation for the `--verbose` flag.

src/printer.rs
```diff
@@ -491,6 +492,14 @@ impl Printer {
         self.buffer.flush()?;
         Ok(())
     }
+
+    pub fn print_response_meta(&mut self, response_meta: ResponseMeta) -> anyhow::Result<()> {
+        let elapsed_time = response_meta.starting_time.elapsed().as_secs_f64();
```
ducaale, line 497: I don't like how `elapsed_time` is affected by when `print_response_meta()` is called.

If only there was a way to read the body without consuming `reqwest::blocking::Response`, We could have something that looked like this:

```rust
dbg!(response.metadata()); // ResponseMeta { request_time: ..., content_download_time: None }

printer.print_response_body(&mut response, response_charset, response_mime)?;

dbg!(response.metadata()); // ResponseMeta { request_time: ..., content_download_time: Some(...) }
```

src/cli.rs
```diff
@@ -80,13 +80,16 @@ pub struct Cli {
...
-    /// String specifying what the output should contain.
+    /// String specifying what the output should contain
     ///
-    /// Use `H` and `B` for request header and body respectively,
-    /// and `h` and `b` for response hader and body.
+    ///     'H' request headers
+    ///     'B' request body
+    ///     'h' response headers
+    ///     'b' response body
+    ///     'm' response metadata
```
ducaale, line 89: This change doesn't seem to play well with manpages :(

ducaale: We can ignore this since manpages will be overhauled in https://github.com/ducaale/xh/pull/250

ducaale:
I think this PR is ready now. I will merge it and we can fix any found issues later

PR #242, opened 2022-03-19
Support sending multiple request headers with the same key

ducaale:
This PR doesn't address how these headers will be saved in a sessions file because we would need to change the structure of the session file and HTTPie hasn't addressed that yet.


blyxxyz:
Works great, but can this be tested?

blyxxyz (approved):

PR #247, opened 2022-04-16
Update Cargo.toml - reqwest dependency from 0.11.6 to 0.11.10

Laiot:
updated reqwest dependency from 0.11.6 to 0.11.10


ducaale:
Can you branch off from develop instead of master for the PR? You might also want to run `cargo build` to update the `Cargo.lock`.

PR #250, opened 2022-04-30
Replace help2man with custom man generation + doc improvements

### LINKED ISSUE #243: Man page improvements
Most of these proposed improvements are inspired by ripgrep:

- Mention that body can be piped to xh in the usage section.
- Briefly mention support for config files in the description section and have a dedicated section for it later.
- Mention the `native-tls` feature flag in the description section.
- Mention Strict compatibility mode in the description section.
- Mention https mode in the description section.
- Move the positional arguments section to outside of the options section.
- Remove any spaces between each flag and its help message.
- Mention `--no-flag` options in the description of the options section.
- Replace `short_help`+`long_help` with `long_help` if needed. (This is done by using `long_help` attribute instead of doc comments)
- Mention flags that affect other flags e.g `--verbose` enables `--all` or `--json` unsets `--form`.
- Document what each possible value does if necessary e.g the values that `--pretty` takes.
- Replace lists with a tabular format and use at least 4 spaces between columns.
- Evaluate alternatives to help2man including [asciidoctor](https://asciidoctor.org/) (used by ripgrep) and [clap_mangen](https://crates.io/crates/clap_mangen).

ducaale:
Doing so lets us have finer control over what the final output looks like. I initially tried to use `clap_mangen` but I wasn't pleased with how the output looked. 

**Edit**: I forgot to mention that the code for the man generation is inspired by `clap_mangen`.

In addition, This PR:
- [x] Documents supported env variables.
- [x] Copies some of the docs in README.md to the manpage.
- [x] Documents any files that `xh` might read i.e sessions, config file, netrc file.
- [x] Documents existing exit codes.
- [x] Documents how xh can be configured.
- [x] Document possible values that a flag might take (if necessary).
- [x] Mention flags that affect other flags e.g `--verbose` enables `--all` or `--json` unsets `--form`.

#### Resources used:
- https://www.youtube.com/watch?v=r2JrbyezhM4
- https://www.systutorials.com/docs/linux/man/7-groff_man/
- https://gnu.org/software/groff/manual/html_node/Man-usage.html
- https://tldp.org/HOWTO/Man-Page/q3.html
- https://www.linuxjournal.com/article/1158
- https://man7.org/linux/man-pages/man7/man-pages.7.html

Resolves #243


src/cli.rs
```diff
@@ -402,6 +441,7 @@ impl Cli {
             }
             "print_completions" => return Err(print_completions(app, cli.raw_rest_args)),
             "generate_completions" => return Err(generate_completions(app, cli.raw_rest_args)),
+            "generate_manpages" => return Err(generate_manpages(app, cli.raw_rest_args)),
```
ducaale, line 444: I am planning to move these subcommands and any dependencies they need (`clap_complete`, `roff`, `chrono`, etc) behind a feature flag.

doc/man-template.roff
```diff
@@ -1,45 +1,143 @@
...
+xh put \fIhttpbin.org/put id:=49 age:=25\fR | less
+Send a PUT request and pipe the result to less.
+.TP
+xh -d \fIhttpbin.org/json\fR -o \fIres.json\fR
+Download and save to res.json.
+.TP
+xh \fIhttpbin.org/get user-agent:foobar\fR
+Make a request with a custom user agent.
+
+.SH REPORTING BUGS
```
ducaale, line 139: I might replace `REPORTING BUG` with `BUGS` as recommended by https://man7.org/linux/man-pages/man7/man-pages.7.html

doc/xh.1
```diff
@@ -1,281 +1,397 @@
...
+        Use a file as the request body
+
+    header:value
+        Add a header
+
+    header:
+        Unset a header
+
+    header;
+        Add a header with an empty value
```
ducaale, line 313: For manpages, manually formatting strings like this instead of using a roff control is not a good idea as it would look bad on smaller screens.

I am yet to figure out how to use roff controls here without having two different help strings and risk them getting out of sync. However, if the two are defined in one place, maybe that would no longer be an issue? 

```rs
const REQUEST_ITEM_HELP: &'static str = "..";
const REQUEST_ITEM_MAN_HELP: &'static str = "..";
```

sorairolake:
I think it would be better to add `curl(1)` and `http(1)` to the SEE ALSO.

1. `xh` has `--curl` flag, so it is useful to be able to see curl's options
2. `http(1)` is already provided by [Debian](https://manpages.debian.org/bullseye/httpie/http.1.en.html) and has been added the upstream as well

ref: [What’s new in HTTPie for Terminal 3.2](https://httpie.io/blog/httpie-3.2.0)

doc/man-template.roff
```diff
@@ -1,45 +1,143 @@
...
+xh's Github issues <https://github.com/ducaale/xh/issues>
+
+.SH SEE ALSO
+curl(1), http(1)
```
sorairolake, line 141: I suggest that the command name in bold. This is inspired by [`curl(1)`](https://curl.se/docs/manpage.html) and [`man-pages(7)`](https://man7.org/linux/man-pages/man7/man-pages.7.html).

ducaale:
One thing that is currently missing is docs for the nested JSON syntax. I wonder if that should have its own section or be part of `REQUEST_ITEM` 🤔.

doc/man-template.roff
```diff
@@ -1,45 +1,145 @@
...
+[\fIREQUEST_ITEM\fR ...]
+
+.SH DESCRIPTION
+
+xh is an HTTP client that aims to reimplement as much as possible of
+HTTPie's excellent design, with a focus on improved performance.
+
+xh uses HTTPie's request-item syntax to set headers, request body,
+query string, etc. For more details, refer to REQUEST_ITEM in the
+OPTIONS section.
```
blyxxyz, line 21: The fact that (like most features) it's copied from HTTPie isn't directly relevant. Maybe this:
```suggestion
Headers, request body, query string, etcetera can be set using
positional arguments. For more details, refer to REQUEST_ITEM in the
OPTIONS section.
```
Or maybe just remove the whole paragraph.

src/cli.rs
```diff
@@ -283,26 +306,68 @@ pub struct Cli {
...
+    ///
+    ///     key:=value
+    ///         Add a field with literal JSON value to the request body e.g. numbers:=[1,2,3]
+    ///         enabled:=true
+    ///
+    ///     key:=@file
+    ///         Add a field with literal JSON value from a file to the request body
+    ///
+    ///     key@file
+    ///         Upload a file from filename (requires either --form  or --multipart).
```
blyxxyz, line 349:
```suggestion
    ///     key==value
    ///         Add a query string to the URL.
    ///
    ///     key=value
    ///         Add a JSON property (--json) or form field (--form) to the request body.
    ///
    ///     key=@file
    ///         Add a JSON property (--json) or form field (--form) from a file to the
    ///         request body.
    ///
    ///     key:=value
    ///         Add a field with literal JSON value to the request body, e.g. numbers:=[1,2,3]
    ///         enabled:=true.
    ///
    ///     key:=@file
    ///         Add a field with literal JSON value from a file to the request body.
    ///
    ///     key@file
    ///         Upload a file from filename (requires either --form  or --multipart).
```

src/cli.rs
```diff
@@ -283,26 +306,68 @@ pub struct Cli {
...
+    ///         Use a file as the request body
+    ///
+    ///     header:value
+    ///         Add a header e.g. user-agent:foobar
+    ///
+    ///     header:
+    ///         Unset a header e.g. connection:
+    ///
+    ///     header;
+    ///         Add a header with an empty value
```
blyxxyz, line 364:
```suggestion
    ///         To set  the filename and mimetype, ";type=" and ";filename=" can be used,
    ///         e.g. pfp@ra.jpg;type=image/jpeg;filename=profile.jpg
    ///
    ///     @filename
    ///         Use a file as the request body.
    ///
    ///     header:value
    ///         Add a header, e.g. user-agent:foobar
    ///
    ///     header:
    ///         Unset a header, e.g. connection:
    ///
    ///     header;
    ///         Add a header with an empty value.
```

src/cli.rs
```diff
@@ -283,26 +306,68 @@ pub struct Cli {
...
+    ///
+    /// A backslash can be used to escape special characters e.g. weird\:key=value.
```
blyxxyz, line 366:
```suggestion
    /// A backslash can be used to escape special characters, e.g. weird\:key=value.
```

src/cli.rs
```diff
@@ -394,14 +459,17 @@ impl Cli {
...
                 app.print_long_help().unwrap();
                 println!();
                 safe_exit();
             }
-            "print_completions" => return Err(print_completions(app, cli.raw_rest_args)),
-            "generate_completions" => return Err(generate_completions(app, cli.raw_rest_args)),
+            #[cfg(feature = "man-completion-gen")]
+            "generate-completions" => return Err(generate_completions(app, cli.raw_rest_args)),
+            #[cfg(feature = "man-completion-gen")]
+            "generate-manpages" => return Err(generate_manpages(app, cli.raw_rest_args)),
```
blyxxyz, line 472: If the feature isn't enabled it would be better to error out instead of treating these as hostnames.

Cargo.toml
```diff
@@ -28,6 +29,7 @@ encoding_rs = "0.8.28"
 encoding_rs_io = "0.1.7"
 flate2 = "1.0.22"
 indicatif = "0.16.2"
+indoc = "1.0"
```
blyxxyz, line 32: It's not a big deal, but do we need this just for that one string?

src/cli.rs
```diff
@@ -619,50 +687,183 @@ fn construct_url(
...
+                header.push(roman("["));
+                header.push(italic(":PASS"));
+                header.push(roman("]|"));
+                header.push(italic("TOKEN"));
+            } else {
+                header.push(italic(&value.join(" ")));
+            }
         }
+        let mut body = vec![];
+        if let Some(help) = opt.get_long_help().or_else(|| opt.get_help()) {
```
blyxxyz, line 758: Are there options without help? If not then we might as well panic here

src/cli.rs
```diff
@@ -619,50 +687,183 @@ fn construct_url(
...
+                );
+                body.push(roman(possible_values_text));
+            }
+        }
+        roff.control("TP", []);
+        roff.text(header);
+        roff.text(body);
     }
+
+    if let Some(opt) = items.iter().find(|opt| opt.get_id() == "raw-method-or-url") {
```
blyxxyz, line 779: I'd go for an `unwrap()` here as well rather than an `if let`. If the argument can't be found then there's a bug somewhere.

src/cli.rs
```diff
@@ -619,50 +687,183 @@ fn construct_url(
...
+        roff.control("TP", []);
+        roff.text(vec![
+            roman("["),
+            italic("METHOD"),
+            roman("]"),
+            italic(" URL"),
+        ]);
```
blyxxyz, line 786: clap considers these to be a single argument but that's not how the user sees it. Can you split them up?

src/cli.rs
```diff
@@ -619,50 +687,183 @@ fn construct_url(
...
+                    }
+                } else {
+                    roff.text(vec![roman(lines[i].trim())]);
+                }
+            }
+            roff.control("RE", []);
+        }
+    }
+
+    let mut manpage = fs::read_to_string(format!("{}/man-template.roff", rest_args[0])).unwrap();
```
blyxxyz, line 857: Maybe we could `include_str!` it instead?

src/cli.rs
```diff
@@ -45,23 +46,38 @@ pub struct Cli {
...
 
     /// Serialize data items from the command line as form fields.
+    ///
+    /// Overrides both --json and --multipart
     #[clap(short = 'f', long, overrides_with_all = &["json", "multipart"])]
     pub form: bool,
 
     /// Like --form, but force a multipart/form-data request even without files.
+    ///
+    /// Overrides both --json and --form
```
blyxxyz, line 62:
```suggestion
    ///
    /// Overrides both --form and --multipart.
    #[clap(short = 'j', long, overrides_with_all = &["form", "multipart"])]
    pub json: bool,

    /// Serialize data items from the command line as form fields.
    ///
    /// Overrides both --json and --multipart.
    #[clap(short = 'f', long, overrides_with_all = &["json", "multipart"])]
    pub form: bool,

    /// Like --form, but force a multipart/form-data request even without files.
    ///
    /// Overrides both --json and --form.
```

src/cli.rs
```diff
@@ -98,6 +114,11 @@ pub struct Cli {
     pub body: bool,
 
     /// Print the whole request as well as the response.
+    ///
+    /// Additionally, this enables --all for printing intermediary
+    /// requests/responses while following redirects
+    ///
+    /// Equivalent to --print=HhBb --all
```
blyxxyz, line 121:
```suggestion
    /// Additionally, this enables --all for printing intermediary
    /// requests/responses while following redirects.
    ///
    /// Equivalent to --print=HhBb --all.
```

blyxxyz (changes requested):
The man page looks slick!

doc/man-template.roff
```diff
@@ -1,45 +1,141 @@
...
+option is "default_options" which is a list of default shell arguments
+that gets passed to xh.
+
+.RS
+.EX
+{
+  "default_options": ["--native-tls", "--style=solarized"]
+}
+.EE
+.RE
```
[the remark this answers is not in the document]
ducaale: Done

doc/man-template.roff
```diff
@@ -1,45 +1,145 @@
...
+HTTP 5xx Server Error.
+.TP
+.B 6
+Too many redirects.
+
+.SH ENVIRONMENT
+.TP
+.B XH_CONFIG_DIR
+Specifies where to look for config.json and named session data.
+The default is $XDG_CONFIG_HOME/xh.
```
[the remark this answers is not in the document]
ducaale: I have replaced `$XDG_CONFIG_HOME` with `~/.config` and mentioned Window's default config directory.

src/cli.rs
```diff
@@ -158,13 +181,13 @@ pub struct Cli {
...
+    /// Authenticate as USER with PASS (-A basic|digest) or with TOKEN (-A bearer).
     ///
-    /// PASS will be prompted if missing. Use a trailing colon (i.e. `USER:`)
+    /// PASS will be prompted if missing. Use a trailing colon (i.e. "USER:")
     /// to authenticate with just a username.
     ///
-    /// TOKEN is expected if `--auth-type=bearer`.
-    #[clap(short = 'a', long, value_name = "USER[:PASS] | TOKEN")]
+    /// TOKEN is expected if --auth-type=bearer.
+    #[clap(short = 'a', long, value_name = "USER[:PASS]|TOKEN")]
```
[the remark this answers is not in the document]
ducaale: I have reverted that change now 👍

src/cli.rs
ducaale: Done

src/cli.rs
ducaale: Done

src/cli.rs
ducaale: I've changed these subcommands to panic when the `man-completion-gen` feature is not enabled

src/cli.rs
ducaale: I think `include_str!` will read the file at compile, but the template path is determined at runtime.

blyxxyz: Do we need to determine it at runtime, though? Is there a case where you'd want to use a different version of the template?

src/cli.rs
blyxxyz: I think an error is more appropriate because it's expected to be possible to trigger it

src/cli.rs
ducaale: Hmm, while there is no reason to determine the path at runtime, I think it would be convenient to read `man-template.roff` at runtime to avoid recompiling xh whenever the template is edited.

src/cli.rs
ducaale: Now, it returns an error

```
$ cargo run -q -- generate-manpages doc
error: generate-manpages requires enabling man-completion-gen feature
```

src/cli.rs
```diff
@@ -619,50 +680,173 @@ fn construct_url(
...
+                );
+                body.push(roman(possible_values_text));
+            }
+        }
+        roff.control("TP", []);
+        roff.text(header);
+        roff.text(body);
     }
+
+    if let Some(opt) = items.iter().find(|opt| opt.get_id() == "raw-method-or-url") {
```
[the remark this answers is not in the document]
ducaale: Done

src/cli.rs
ducaale: This one is a little tricky unless the documentation for Method and URL is duplicated somewhere else 🤔.

doc/man-template.roff
ducaale: I've opted to update it, but I wouldn't mind removing it and the paragraph after it.

Cargo.toml
ducaale: You're right, I've now moved `indoc` back to dev-dependencies.

ducaale:
Similar to ripgrep, should we add a dedicated section for positional arguments?

```
XH(1)                                                           User Commands                                                          XH(1)

NAME
       xh - Friendly and fast tool for sending HTTP requests

SYNOPSIS
       xh [OPTIONS] [METHOD] URL [--] [REQUEST_ITEM ...]

DESCRIPTION
       xh is an HTTP client that aims to reimplement as much as possible of HTTPie's excellent design, with a focus on improved performance.

POSITIONAL ARGUMENTS
       [METHOD] URL
              The request URL, preceded by an optional HTTP method.

       [REQUEST_ITEM ...]
              Optional key-value pairs to be included in the request

OPTIONS
       Each --OPTION can be reset with a --no-OPTION argument.

       -j, --json
              (default) Serialize data items from the command line as a JSON object.

              Overrides both --form and --multipart.
```

blyxxyz:
I think that's a good idea. (Though again with method and URL separate.)

src/cli.rs
ducaale: I don't think the docs for method and URL will frequently be changing, so I duplicated the docs in `man-template.roff`.

src/cli.rs
```diff
@@ -394,14 +470,18 @@ impl Cli {
         let matches = app.try_get_matches_from_mut(iter)?;
         let mut cli = Self::from_arg_matches(&matches)?;
 
+        // opt-out of claps auto-generated possible values for --pretty
+        // as we already list them in the long_help
+        app = app.mut_arg("pretty", |a| a.hide_possible_values(true));
```
ducaale, line 475: **Note to self**: figure out how to restore the order of options  in manpages after mutating one of them them

Cargo.toml
```diff
@@ -18,8 +18,9 @@ anyhow = "1.0.38"
 atty = "0.2"
 brotli = { version = "3.3.0", default-features = false, features = ["std"] }
 chardetng = "0.1.15"
+chrono = { version = "0.4", optional = true }
```
sorairolake, line 21: What about using [`time`](https://crates.io/crates/time) crate instead of this if possible?

src/cli.rs
```diff
@@ -293,24 +325,66 @@ pub struct Cli {
...
+    ///
+    ///     header;
+    ///         Add a header with an empty value.
+    ///
+    /// A backslash can be used to escape special characters, e.g. weird\:key=value.
+    ///
+    /// To construct a complex JSON object, the REQUEST_ITEM's key can be set to a JSON path
+    /// instead of a field name. For more information on the nested json syntax, refer to
+    /// https://httpie.io/docs/cli/nested-json.
     #[clap(value_name = "REQUEST_ITEM", verbatim_doc_comment)]
```
ducaale, line 388: It bothers me a little how the text loses its responsiveness with the `verbatim_doc_comment` attribute. For example, this is how one of the paragraphs looks in 80 column width

```
            To construct a complex JSON object, the REQUEST_ITEM's key can be
            set to a JSON path
            instead of a field name. For more information on the nested json
            syntax, refer to
            https://httpie.io/docs/cli/nested-json.
```

And this is how the help for `key@file` is rendered
```
                    To set  the filename and mimetype, ";type=" and
            ";filename=" can be used,
                    e.g. pfp@ra.jpg;type=image/jpeg;filename=profile.jpg
```

I think it would have been nice if there was a way to force newlines and leading spaces without using `verbatim_doc_comment`. Although this wouldn't address the issue in the 2nd example without having something similar to roff's `TP` keyword.

ducaale:
@blyxxyz, is there anything else we can address in this PR now?

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
+    /// A leading colon works as shorthand for localhost. ":8000" is equivalent
+    /// to "localhost:8000", and ":/path" is equivalent to "localhost/path".
+    ///
+    /// URLs can have a leading "://" which allows quickly converting a URL
+    /// into a valid xh or HTTPie command. For example "http://httpbin.org/json"
+    /// becomes "http ://httpbin.org/json".
     #[clap(value_name = "[METHOD] URL")]
     raw_method_or_url: String,
 
     /// Optional key-value pairs to be included in the request
```
blyxxyz, line 343:
```suggestion
    /// Optional key-value pairs to be included in the request.
```

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
-    ///   - key@filename to upload a file from filename (with --form)
-    ///   - @filename to use a file as the request body
-    ///   - header:value to add a header
-    ///   - header: to unset a header
-    ///   - header; to add a header with an empty value
+    /// The separator is used to determine the type i.e. header, request body, query string, etc.
+    ///
+    /// Possible REQUEST_ITEM types are:
+    ///
+    ///     key==value
```
blyxxyz, line 349:
```suggestion
    /// The separator is used to determine the type:
    ///
    ///     key==value
```

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
+    ///         the request body.
+    ///
+    ///     key=@file
+    ///         Add a JSON property (--json) or form field (--form) from a
+    ///         file to the request body.
     ///
-    /// A backslash can be used to escape special characters (e.g. weird\:key=value).
+    ///     key:=value
+    ///         Add a field with literal JSON value to the request body,
+    ///         e.g. numbers:=[1,2,3] enabled:=true.
```
blyxxyz, line 362: I'm worried these examples don't stand out enough from the surrounding text. Maybe we can put quotes around them? Or use special roff formatting?

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
+    ///
+    ///     key:=@file
+    ///         Add a field with literal JSON value from a file to the
```
blyxxyz, line 365:
```suggestion
    ///         Add a field with a literal JSON value to the request body,
    ///         e.g. numbers:=[1,2,3] enabled:=true.
    ///
    ///     key:=@file
    ///         Add a field with a literal JSON value from a file to the
```

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
+    /// Possible REQUEST_ITEM types are:
+    ///
+    ///     key==value
+    ///         Add a query string to the URL.
+    ///
+    ///     key=value
+    ///         Add a JSON property (--json) or form field (--form) to
+    ///         the request body.
+    ///
+    ///     key=@file
```
blyxxyz, line 356:
```suggestion
    ///     key=@filename
```

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
-    /// A backslash can be used to escape special characters (e.g. weird\:key=value).
+    ///     key:=value
+    ///         Add a field with literal JSON value to the request body,
+    ///         e.g. numbers:=[1,2,3] enabled:=true.
+    ///
+    ///     key:=@file
+    ///         Add a field with literal JSON value from a file to the
+    ///         request body.
+    ///
+    ///     key@file
```
blyxxyz, line 368:
```suggestion
    ///     key:=@filename
    ///         Add a field with literal JSON value from a file to the
    ///         request body.
    ///
    ///     key@filename
```

src/cli.rs
```diff
@@ -293,24 +325,70 @@ pub struct Cli {
...
+    ///         Unset a header, e.g. connection:
+    ///
+    ///     header;
+    ///         Add a header with an empty value.
+    ///
+    /// A backslash can be used to escape special characters, e.g. weird\:key=value.
+    ///
+    /// To construct a complex JSON object, the REQUEST_ITEM's key can be set to a JSON path instead of a field name.
+    ///
+    /// For more information on the nested json syntax, refer to https://httpie.io/docs/cli/nested-json.
```
blyxxyz, line 391: These two paragraphs have much more to do with each other than with the paragraph right before, so I think it's clearest to merge them:
```suggestion
    /// To construct a complex JSON object, the REQUEST_ITEM's key can be set to a JSON path instead of a field name.
    /// For more information on this syntax, refer to https://httpie.io/docs/cli/nested-json.
```

doc/man-template.roff
```diff
@@ -1,45 +1,148 @@
...
+.SH EXAMPLES
+.TP 4
+\fBxh\fR \fIhttpbin.org/json\fR
+Send a GET request.
+.TP
+\fBxh\fR \fIhttpbin.org/post name=ahmed \fIage:=24\fR
+Send a POST request with body {"name": "ahmed", "age": 24}.
+.TP
+\fBxh\fR get \fIhttpbin.org/json id==5 sort==true\fR
+Send a GET request with querystring id=5&sort=true.
```
blyxxyz, line 122: I like these examples!

```suggestion
Send a GET request to http://httpbin.org/json?id=5&sort=true.
```
(Not everyone knows what a querystring is)

doc/man-template.roff
```diff
@@ -1,45 +1,148 @@
...
+.TP
+\fBxh\fR get \fIhttpbin.org/json x-api-key:12345\fR
+Send a GET request and include a header named x-api-key with value 12345.
```
blyxxyz, line 125:
```suggestion
Send a GET request and include a header named X-Api-Key with value 12345.
```

blyxxyz (commented):
Looking good! Only a few more comments about the content.

src/cli.rs
ducaale: I have now added double quotes to the examples inside `[REQUEST_ITEM ...]`.

doc/man-template.roff
```diff
@@ -1,45 +1,148 @@
...
+[\fIOPTIONS\fR]
+[\fIMETHOD\fR]
+\fIURL\fR
+[\-\-\]
+[\fIREQUEST_ITEM\fR ...]
+
+.SH DESCRIPTION
+
+xh is an HTTP client that aims to reimplement as much as possible of
+HTTPie's excellent design, with a focus on improved performance.
```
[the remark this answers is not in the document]
blyxxyz:
```suggestion
.B xh
is an HTTP client with a friendly command line interface. It strives to
have readable output and easy-to-use options.

xh is mostly compatible with HTTPie: see
.BR http (1).
```
I ultimately couldn't think of a lot to write.

It's common to bold the name at the start of the description, Wikipedia-style, so I added that.

A lot of people won't have the HTTPie man page installed, so the reference could be confusing. But it'll get auto-linkified by man page websites ([e.g.](https://man.archlinux.org/man/xzgrep.1.en)) so on balance I think it's for the best.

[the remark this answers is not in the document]
ducaale: Thanks, I think this description fits better on a man page.

Would it be helpful to mention the cURL translation feature since that seems to be the feature that users mostly appreciate? or should we avoid pointing to the same feature multiple times on a man page?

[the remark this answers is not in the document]
blyxxyz: That might be a good one, since it isn't in HTTPie! I've edited my suggestion.

On a related note, the curl2httpie.online link doesn't work right now, so if it isn't fixed we might have to remove or replace it. I've opened dcb9/curl2httpie#15.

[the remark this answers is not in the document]
ducaale: >On a related note, the curl2httpie.online link doesn't work right now, so if it isn't fixed we might have to remove or replace it. I've opened https://github.com/dcb9/curl2httpie/issues/15.

I think the certificate for curl2httpie.online has been invalid for quite a while.

If only there were a Rust version that can be compiled into WebAssembly...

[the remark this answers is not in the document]
blyxxyz: It's currently a Go version that looks like it's compiled to asm.js, and could be compiled to WebAssembly instead. (It wouldn't be as lean as Rust, but it should be leaner than it is now.)

That makes it possible for us to rehost it, only that seems a little impolite and it might technically be illegal since it doesn't have a license.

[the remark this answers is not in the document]
blyxxyz: The website is working again! Maybe we should've opened that issue half a year ago.

src/cli.rs
```diff
@@ -70,22 +89,22 @@ pub struct Cli {
 
     /// Override the response encoding for terminal display purposes.
     ///
-    /// Example: `--response-charset=latin1`
+    /// Example: --response-charset=latin1
     #[clap(long, value_name = "ENCODING", parse(try_from_str = parse_encoding))]
     pub response_charset: Option<&'static Encoding>,
 
     /// Override the response mime type for coloring and formatting for the terminal
```
blyxxyz, line 96:
```suggestion
    /// Override the response mime type for coloring and formatting for the terminal.
```

src/cli.rs
```diff
@@ -631,53 +691,210 @@ fn construct_url(
...
+            } else {
+                header.push(italic(&value.join(" ")));
+            }
+        }
+        let mut body = vec![];
+
+        let help = opt
+            .get_long_help()
+            .or_else(|| opt.get_help())
+            .expect("option is missing help");
```
blyxxyz, line 845:
```suggestion
        let mut help = opt
            .get_long_help()
            .or_else(|| opt.get_help())
            .expect("option is missing help")
            .to_owned();
        if !help.ends_with('.') {
            help.push('.')
        }
```
This gives a more consistent style. I've found manpages that always end an option description with a period, and ones that never end it with a period and always start it with a lowercase letter, but I've found none that mix periods like clap does.

(Maybe we can change this in clap instead? I wouldn't mind seeing more periods in the long help. But that's something to look at once we're on clap 4.)

blyxxyz (approved):
Some concerns about trailing periods but after that I think it's ready to merge.

PR #256, opened 2022-05-14
Add optional install-dir

lispyclouds:
Thanks a lot for this!

This adds a ability to specify an install dir when using the `install.sh` script. I use the script to install via automation on Fedora and it waiting for input is a bit of an issue. This could be a non-intrusive/breaking way for allowing an install dir?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -59,9 +59,25 @@ case $PATH in
         ;;
 esac
 
-printf "Install location [default: %s]: " "$default_bin"
-read -r bindir < /dev/tty
-bindir=${bindir:-$default_bin}
+while [[ $# -gt 0 ]]
+do
+    key="$1"
+    case "$key" in
+        --install-dir)
+            install_dir="$2"
+            shift
+            shift
+            ;;
+    esac
+done
+
+if [[ -z "$install_dir" ]]; then
+    printf "Install location [default: %s]: " "$default_bin"
+    read -r bindir < /dev/tty
+    bindir=${bindir:-$default_bin}
+else
+    bindir=${install_dir}
+fi
 
 while ! test -d "$bindir"; do
     echo "Directory $bindir does not exist"
--- a/install.sh
+++ b/install.sh
@@ -59,31 +59,33 @@ case $PATH in
         ;;
 esac
 
-_read_bindir() {
+_read_installdir() {
     printf "Install location [default: %s]: " "$default_bin"
-    read -r XH_BINDIR < /dev/tty
-    XH_BINDIR=${XH_BINDIR:-$default_bin}
+    read -r xh_installdir < /dev/tty
+    xh_installdir=${xh_installdir:-$default_bin}
 }
 
 if [ -z "$XH_BINDIR" ]; then
-    _read_bindir
+    _read_installdir
 
-    while ! test -d "$XH_BINDIR"; do
-      echo "Directory $XH_BINDIR does not exist"
-      _read_bindir
+    while ! test -d "$xh_installdir"; do
+        echo "Directory $xh_installdir does not exist"
+        _read_installdir
     done
+else
+    xh_installdir=${XH_BINDIR}
 fi
 
 tar xzf xh.tar.gz
 
-if test -w "$XH_BINDIR"; then
-    mv xh-*/xh "$XH_BINDIR/"
-    ln -sf "$XH_BINDIR/xh" "$XH_BINDIR/xhs"
+if test -w "$xh_installdir" || [ -n "$XH_BINDIR" ]; then
+    mv xh-*/xh "$xh_installdir/"
+    ln -sf "$xh_installdir/xh" "$xh_installdir/xhs"
 else
-    sudo mv xh-*/xh "$XH_BINDIR/"
-    sudo ln -sf "$XH_BINDIR/xh" "$XH_BINDIR/xhs"
+    sudo mv xh-*/xh "$xh_installdir/"
+    sudo ln -sf "$xh_installdir/xh" "$xh_installdir/xhs"
 fi
 
-echo "$("$XH_BINDIR"/xh -V) has been installed to:"
-echo " • $XH_BINDIR/xh"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -59,9 +59,25 @@ case $PATH in
         ;;
 esac
 
-printf "Install location [default: %s]: " "$default_bin"
-read -r bindir < /dev/tty
-bindir=${bindir:-$default_bin}
+while [[ $# -gt 0 ]]
```
blyxxyz, line 62: Not all shells have `[[`, but `[` should work:
```suggestion
while [ $# -gt 0 ]
```

install.sh
```diff
@@ -59,9 +59,25 @@ case $PATH in
...
+do
+    key="$1"
+    case "$key" in
```
blyxxyz, line 65: This should have another case to handle an unknown argument, or it'll loop endlessly.

install.sh
```diff
@@ -59,9 +59,25 @@ case $PATH in
...
+        --install-dir)
+            install_dir="$2"
+            shift
+            shift
+            ;;
+    esac
+done
+
+if [[ -z "$install_dir" ]]; then
```
blyxxyz, line 74:
```suggestion
if [ -z "$install_dir" ]; then
```

install.sh
```diff
@@ -59,9 +59,25 @@ case $PATH in
...
+    printf "Install location [default: %s]: " "$default_bin"
+    read -r bindir < /dev/tty
+    bindir=${bindir:-$default_bin}
+else
+    bindir=${install_dir}
+fi
 
 while ! test -d "$bindir"; do
```
blyxxyz, line 82: If the install directory was explicit then we'd better skip this loop and crash instead of prompting.

blyxxyz (commented):
Good idea!

We could instead use an environment variable. This is a minimal implementation:
```diff
diff --git a/install.sh b/install.sh
index 02714d4..5c45567 100644
--- a/install.sh
+++ b/install.sh
@@ -59,12 +59,10 @@ case $PATH in
         ;;
 esac
 
-printf "Install location [default: %s]: " "$default_bin"
-read -r bindir < /dev/tty
-bindir=${bindir:-$default_bin}
-
 while ! test -d "$bindir"; do
-    echo "Directory $bindir does not exist"
+    if test "$bindir"; then
+        echo "Directory $bindir does not exist"
+    fi
     printf "Install location [default: %s]: " "$default_bin"
     read -r bindir < /dev/tty
     bindir=${bindir:-$default_bin}
```
You'd run `curl [...] | bindir=/path/to/dir sh`.

Problems (all fixable):
- `$bindir` is just vague enough that someone might already have it set for another reason, maybe `$XH_BINDIR` would be better
- If the directory doesn't exist it keeps prompting

I'm not sure if I prefer that over your solution with options.

install.sh
ducaale: FYI, you can use https://www.shellcheck.net/ to lint shell scripts

```
[Line 62:](javascript:setPosition(62, 7))
while [[ $# -gt 0 ]]
      ^-- [SC3010](https://www.shellcheck.net/wiki/SC3010) (warning): In POSIX sh, [[ ]] is undefined.
 
[Line 74:](javascript:setPosition(74, 4))
if [[ -z "$install_dir" ]]; then
   ^-- [SC3010](https://www.shellcheck.net/wiki/SC3010) (warning): In POSIX sh, [[ ]] is undefined.
```

lispyclouds:
Thanks for the feedback, I'll push in the changes and as for the options vs env vars, I dont really have a preference; can change it if you recommend it!

ducaale:
Is it possible to set the install directory when piping `install.sh` to `sh`? If not, I would vote for the env var approach

```
curl [...] | XH_BINDIR=/path/to/dir sh
```

lispyclouds:
I think we'd need to use xargs there. If that seems too clunky, I can change this to env vars.

lispyclouds:
Changing to an env var approach, will revert back if needed.

install.sh
```diff
@@ -59,27 +59,31 @@ case $PATH in
...
+    read -r XH_BINDIR < /dev/tty
+    XH_BINDIR=${XH_BINDIR:-$default_bin}
+}
+
+if [ -z "$XH_BINDIR" ]; then
+    _read_bindir
+
+    while ! test -d "$XH_BINDIR"; do
+      echo "Directory $XH_BINDIR does not exist"
+      _read_bindir
```
blyxxyz, line 73:
```suggestion
        echo "Directory $XH_BINDIR does not exist"
        _read_bindir
```

install.sh
```diff
@@ -59,27 +59,31 @@ case $PATH in
...
-    mv xh-*/xh "$bindir/"
-    ln -sf "$bindir/xh" "$bindir/xhs"
+if test -w "$XH_BINDIR"; then
+    mv xh-*/xh "$XH_BINDIR/"
+    ln -sf "$XH_BINDIR/xh" "$XH_BINDIR/xhs"
 else
-    sudo mv xh-*/xh "$bindir/"
-    sudo ln -sf "$bindir/xh" "$bindir/xhs"
+    sudo mv xh-*/xh "$XH_BINDIR/"
+    sudo ln -sf "$XH_BINDIR/xh" "$XH_BINDIR/xhs"
```
blyxxyz, line 84: If `$XH_BINDIR` does not exist we take this branch, which doesn't help.

Maybe we wouldn't want to use sudo at all if the environment variable was used, to guarantee it's non-interactive.

Perhaps:
- Go back to using `$bindir`, with an explicit `bindir=$XH_BINDIR`
- Change the condition to `test -w "$bindir" || [ -n "$XH_BINDIR" ]`, so it doesn't try to be smart

lispyclouds:
@blyxxyz I addressed the feedback, decided to call it `xh_installdir`. Hope that's better?

blyxxyz (approved):
Yes, works great!

PR #257, opened 2022-05-14
Handle empty gzip responses with content-length > 0

### LINKED ISSUE #253: HEAD fails on compressed response
```
$ xh head httpbin.org/gzip
[...]
xh: error: error decoding gzip response body: failed to fill whole buffer
```
This didn't happen before #241, so reqwest deals with it correctly. I don't know how.

ducaale:
Resolves #253

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -51,6 +51,7 @@ pub fn get_compression_type(headers: &HeaderMap) -> Option<CompressionType> {
 
 struct InnerReader<R: Read> {
     reader: R,
+    has_read_data: bool,
     has_errored: bool,
 }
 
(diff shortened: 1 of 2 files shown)


src/decoder.rs
```diff
@@ -86,36 +92,33 @@ impl<R: Read> Read for Decoder<R> {
...
-                    io::Error::new(
-                        e.kind(),
-                        format!("error decoding brotli response body: {}", e),
-                    )
-                }
-            }),
+            Decoder::Gzip(decoder) => match decoder.read(buf) {
+                Ok(n) => Ok(n),
+                Err(e) if decoder.get_ref().has_errored => Err(e),
+                Err(_) if !decoder.get_ref().has_read_data => Ok(0),
```
ducaale, line 98: The check has to be done in this order to avoid ignoring `InnerReader` errors when no data has been read

blyxxyz (approved):
Works great!

PR #258, opened 2022-05-22
Add some exit status

sorairolake:
The exit status will be 2 when the request timed out, and the exit status will be 6 when the redirect exceeds `--max-redirects <NUM>`. This is the same behavior as HTTPie.

ref: <https://httpie.io/docs/cli/scripting>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -84,13 +84,15 @@ fn main() {
                 eprintln!();
                 eprintln!("Try running without the --native-tls flag.");
             }
-            if msg == "operation timed out" {
-                process::exit(2);
-            } else if msg.starts_with("Too many redirects") {
-                process::exit(6);
-            } else {
-                process::exit(1);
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -84,7 +84,13 @@ fn main() {
                 eprintln!();
                 eprintln!("Try running without the --native-tls flag.");
             }
-            process::exit(1);
+            if msg == "operation timed out" {
```
blyxxyz, line 87: There's a more robust way to detect this:
```rust
            if let Some(err) = err.downcast_ref::<reqwest::Error>() {
                if err.is_timeout() {
                    [...]
                }
            }
```

We could also do that for the "Too many redirects" case by turning it into a custom error type, but there we control the error message ourselves so it's not necessary.

blyxxyz (commented):
Good idea!

blyxxyz (approved):

PR #261, opened 2022-07-10
Move cross config from `Cross.toml` to `Cargo.toml`

ducaale:
The recently released [cross-rs v0.2.2](https://github.com/cross-rs/cross/releases/tag/v0.2.2) can read configuration from `Cargo.toml` metadata which allows us to get rid of `Cross.toml` in the top-level directory.

Note that `ci.yaml` is still using v0.2.1 due to cached build artifacts but that shouldn't happen in `release.yaml`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,9 @@ default = ["online-tests"]
 online-tests = []
 native-tls = ["reqwest/native-tls", "reqwest/native-tls-alpn"]
 
+[package.metadata.cross.build.env]
+passthrough = ["CARGO_PROFILE_RELEASE_LTO"]
+
 [package.metadata.deb]
 features = []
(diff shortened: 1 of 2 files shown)


blyxxyz (approved):
Looks good, but can we test this?

ducaale:
* To test if `package.metadata.cross.build.env.passthrough` is working, I tested locally with ` NO_COLOR=1 cross run --target aarch64-unknown-linux-musl -- httpbin.org/json` with and without adding `NO_COLOR` to the list.
* I double-checked that cross v0.2.4 is being used in CI when the caching is disabled. See https://github.com/ducaale/xh/runs/7273340419?check_suite_focus=true

blyxxyz (approved):

PR #264, opened 2022-07-12
Improve PATH handling in Powershell, fixes #263

### LINKED ISSUE #263: Powershell install script messes PATH up if the PATH does not end with a semicolon
1. Install xh as per instructions

Expected: 
xh is runnable from the current terminal window and from the newly opened terminal windows.

Actual:
`xh` is not found when attempting to execute it immediately after installing and even after restarting the window. Inspecting the `PATH` it becomes apparent that the install script expects the `PATH` to end with a semicolon and when it does not, the new `bin` directory gets appended to the PATH without the semicolon, resulting in something like `...;C:\Program Files\nodejsC:\Users\user\bin\`

What needs to be done is not to assume that the semicolon is at the end of the PATH string while also expecting that it might be there.

Also, modifying the `$env:Path` within the script does not modify it outside of the script (when the script has ended). I

henno:


install.ps1
```diff
@@ -3,67 +3,65 @@
...
+{
+    Write-Output "PATH did not contain $destdir"
+    
+    # Update userPath
+    $userPath = $userPath.Trim(";") + ";$destdir"
+    
+    # Modify PATH for new windows
+    Write-Output "Adding $destdir to user PATH."
+	[System.Environment]::SetEnvironmentVariable('Path', $userPath, [System.EnvironmentVariableTarget]::User)
+	Write-Output "Done."
```
ducaale, line 60: I think the first and third lines are redundant and could be removed
```
PATH did not contain $destdir
Adding $destdir to user PATH.
Done.
```

install.ps1
```diff
@@ -3,67 +3,65 @@
...
 	
-    # Path to xhs.exe.
-
-    Write-Host "PATH environment variable changed (restart your applications that use command line)." -foreground yellow
+    # Instruct how to modify PATH for the current window
+    Write-Output "Execute the following command to make xh usable immediately in PowerShell:"
+    Write-Host "`$Env:Path = `$Env:Path.Trim(`";`") + `";$destdir`"" -foreground yellow
+    
+    Write-Host "Or restart any terminal to refresh its PATH and use xh."
```
ducaale, line 66: Can you add a new line before this instruction is outputted?

```
Downloading: xh-v0.16.1-x86_64-pc-windows-msvc.zip
xh v0.16.1 has been installed to:
 - C:\Users\shara\bin\xh.exe
 - C:\Users\shara\bin\xhs.exe

PATH environment variable updated. For changes to take affect, please restart terminal or execute the following command:
$Env:Path = $Env:Path.Trim(";") + ";C:\Users\shara\bin"
```

ducaale (commented):
Thanks, I haven't encountered this edge case before, but I think it is good that it is being handled now.

install.ps1
```diff
@@ -3,67 +3,65 @@
...
+
+# Inform user where the executables have been put
+Write-Output "xh v$( $xhVersion ) has been installed to:`n - $xhPath`n - $xhsPath"
+
+# Make sure destdir is in the path
+$userPath = [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::User)
+$machinePath = [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine)
+
+# If userPath AND machinePath both do not contain bin, then add it to user path
+if (!($userPath -contains $destdir) -and !($machinePath -contains $destdir))
```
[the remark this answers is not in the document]
henno: Done.

install.ps1
henno: Done

install.ps1
henno: The script now automatically updates the current terminal's PATH, no need to run the PS command. To ensure a dependable experience, it now recommends restarting the IDE instead of running that PS command, as new terminal tabs within IDEs would still inherit the outdated environment with the PS command.

ducaale (approved):
Thanks!

Issue #262, opened 2022-07-10
Update Clap to v3.2


ducaale:
~~n anticipation of the next major version, Clap v3.2 deprecated many existing APIs. Therefore, It would be beneficial for us to update early so we can provide early feedback regarding the new APIs.~~

See https://epage.github.io/blog/2022/06/clap-32-last-call-before-40/

**Update**: Clap v4 has been released, see https://epage.github.io/blog/2022/09/clap4/

gautamprikshit1:
Would you like to update to clap v4?

ducaale:
@gautamprikshit1 Unfortunately, clap v4 requires rust 1.64 which is too high for us at the moment. We might reconsider raising the MSRV once reqwest or another essential dependency raises their MSRV.

ducaale:
The MSRV has been raised to 1.64, so we should be able to upgrade clap to v4 as soon as https://github.com/ducaale/xh/pull/269 is merged.

gautamprikshit1:
Please let me know when clap 3 PR is merged

ducaale:
Thanks @gautamprikshit1 - There were a lot of moving parts so I went ahead and updated Clap to v4.0. However, you can still help us with upgrading to v4.3 and that would be greatly appreciated.

gautamprikshit1:
It might take a while as I already have another issue assigned in other project for this weekend, next weekend for sure

PR #269, opened 2022-08-18
Update clap to v3.2

ducaale:
- Fix deprecation warnings
  - ~Replace `args_override_self` with `action` specified on each argument~
  - Replace `ArgEnum` with `ValueEnum`
  - Replace `parse()` with the new `value_parser`
  - implement `ValueParserFactory` for `Verify` See https://github.com/clap-rs/clap/discussions/4007
  - Introduce an enum for TLS version
- ~Bump MSRV to 1.56.1~

Resolves https://github.com/ducaale/xh/issues/262


ducaale:
I am yet to find a replacement for `args_override_self` that works for negation flags
```
$ cargo run -q -- : --no-json --no-json
error: The argument '--no-json' was provided more than once, but cannot be used multiple times

USAGE:
    xh [OPTIONS] <[METHOD] URL> [--] [REQUEST_ITEM]...

For more information try --help
```

ducaale:
Interesting, I wasn't aware that some options were un-deprecated.

I have also reverted to using the un-deprecated [`get_possible_values()`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.get_possible_values) in `generate_manpages()` to ensure `[possible values: ...]` is generated for any `ArgEnum` that doesn't have an explicit name i.e `Theme`, `Pretty` and `AuthType`.

src/cli.rs
```diff
@@ -563,12 +557,18 @@ impl Cli {
...
+        // XH_HTTPIE_COMPAT_MODE
+        if matches!(
+            matches.value_source("check-status"),
+            Some(clap::ValueSource::CommandLine)
+        ) || matches!(
+            matches.value_source("no-check-status"),
+            Some(clap::ValueSource::CommandLine)
+        ) {
+            self.check_status = Some(self.check_status_raw);
+        }
```
blyxxyz, line 571: Maybe it would be better to replace `is_present()` by `get_flag()`

blyxxyz (commented):
Looks good!

- Can you regenerate the man page so we can scrutinize its diff?
- Do you know how much work it'll be to upgrade to v4 after this?

ducaale:
>Do you know how much work it'll be to upgrade to v4 after this?

Apart from a lifetime error which I still need to figure out how to approach, upgrading to v4.0 should require minimal changes.

completions/_xh
```diff
@@ -20,18 +20,19 @@ _xh() {
 colors\:"Apply syntax highlighting to output"
 format\:"Pretty-print json and sort headers"
 none\:"Disable both coloring and formatting"))' \
+'--format-options=[Set output formatting options. The only currently implemented option is json.indent]:FORMAT_OPTIONS: ' \
 '-s+[Output coloring style]:THEME:(auto solarized monokai fruity)' \
 '--style=[Output coloring style]:THEME:(auto solarized monokai fruity)' \
-'--response-charset=[Override the response encoding for terminal display purposes]:ENCODING: ' \
+'*--response-charset=[Override the response encoding for terminal display purposes]:ENCODING: ' \
```
blyxxyz, line 26: It's probably fine but I wonder what's going on here.

blyxxyz (approved):
Looks great!

completions/_xh
blyxxyz: zsh still completes it so ¯\\_ (ツ)_/¯

PR #271, opened 2022-08-28
Remove `#[allow(non_camel_case_types)]`

sorairolake:
This is because variant names are shown up in kebab case by default.


ducaale (approved):
Looks good, thanks!

PR #275, opened 2022-09-05
Support forcing ipv4/ipv6

zuisong:
Resolves #274

PR #276, opened 2022-09-06
Support forcing ipv4/ipv6

### LINKED ISSUE #274: Feature request: support forcing ipv4/ipv6
`cURL` supports `-4`/`--ipv4` and `-6`/`--ipv6` flags. Quoting the `curl(1)` manpage:

``` text
       -4, --ipv4
              This option tells curl to resolve names to IPv4 addresses only, and
              not for example try IPv6.

              Example:
               curl --ipv4 https://example.com

              See also --http1.1 and --http2. This option overrides -6, --ipv6.

       -6, --ipv6
              This option tells curl to resolve names to IPv6 addresses only, and
              not for example try IPv4.

              Example:
               curl --ipv6 https://example.com

              See also --http1.1 and --http2. This option overrides -4, --ipv4.
```

For testing purposes, having this feature would be really convenient.

zuisong:
Resolves #274

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -109,6 +109,14 @@ pub struct Cli {
     #[clap(short = 'P', long, value_name = "FORMAT")]
     pub history_print: Option<Print>,
 
+    /// Resolve hostname to ipv4 addresses only.
+    #[clap(short = '4', long)]
+    pub ipv4: bool,
+
+    /// Resolve hostname to ipv6 addresses only.
+    #[clap(short = '6', long)]
+    pub ipv6: bool,
+
     /// Do not print to stdout or stderr.
     #[clap(short = 'q', long)]
     pub quiet: bool,
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -1199,11 +1199,7 @@ fn cert_without_key() {
 #[test]
 fn use_ipv4() {
     get_command()
-        .args(&[
-            "https://v4v6.ipv6-test.com/api/myip.php",
-            "--body",
-            "--ipv4",
-        ])
+        .args(&["https://api64.ipify.org", "--body", "--ipv4"])
         .assert()
         .stdout(function(|output: &str| {
             IpAddr::from_str(output.trim()).unwrap().is_ipv4()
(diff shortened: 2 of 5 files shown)


zuisong:
Replace #275

src/cli.rs
```diff
@@ -109,6 +109,14 @@ pub struct Cli {
     #[clap(short = 'P', long, value_name = "FORMAT")]
     pub history_print: Option<Print>,
 
+    /// Resolve hostname to ipv4 addresses only.
+    #[clap(short = '4', long)]
+    pub ipv4: bool,
+
+    /// Resolve hostname to ipv6 addresses only.
+    #[clap(short = '6', long)]
+    pub ipv6: bool,
```
ducaale, line 118: Can you include these flags in `to_curl.rs`, which handles translating xh commands to curl?

tests/cli.rs
```diff
@@ -1191,6 +1195,41 @@ fn cert_without_key() {
...
+        .stderr(predicates::str::is_empty());
+}
+
+#[cfg(feature = "online-tests")]
+#[test]
+fn use_ipv6() {
+    get_command()
+        .args(&["https://v4.ipv6-test.com/api/myip.php", "--body", "--ipv4"])
+        .assert()
+        .success();
```
ducaale, line 1225: Can we have an ipv6 test similar to the earlier ipv4 test?

```rust
get_command()
    .args(&[
        "https://v4v6.ipv6-test.com/api/myip.php",
        "--body",
        "--ipv6",
    ])
    .assert()
    .stdout(function(|output: &str| {
        IpAddr::from_str(output.trim()).unwrap().is_ipv6()
    }))
    .stderr(predicates::str::is_empty());
```

tests/cli.rs
```diff
@@ -1191,6 +1195,41 @@ fn cert_without_key() {
...
+        .assert()
+        .stdout(function(|output: &str| {
+            IpAddr::from_str(output.trim()).unwrap().is_ipv4()
+        }))
+        .stderr(predicates::str::is_empty());
+}
+
+#[cfg(feature = "online-tests")]
+#[test]
+fn use_ipv6() {
```
ducaale, line 1221: Since not everyone will be to use ipv6 locally, I am wondering if we should gate this behind a feature flag. Maybe `ipv6-tests`?

ducaale (commented):
Looks good. I had to test this on a DigitalOcean droplet since ipv6 doesn't work for me locally.

tests/cli.rs
zuisong: not everyone will be able to use ipv6 locally, it will be fail if does not have ipv6 address

maybe we can keep those test case, add a test case with a feature

tests/cli.rs
```diff
@@ -1191,6 +1195,58 @@ fn cert_without_key() {
...
+        .args(&[
+            "https://v4v6.ipv6-test.com/api/myip.php",
+            "--body",
+            "--ipv4",
+        ])
+        .assert()
+        .stdout(function(|output: &str| {
+            IpAddr::from_str(output.trim()).unwrap().is_ipv4()
+        }))
+        .stderr(predicates::str::is_empty());
```
ducaale, line 1216: This test seems to fail randomly. Can we have it behind a feature flag as well?

ducaale (changes requested):

tests/cli.rs
zuisong: may be wo can use another api to test ipv4/ipv6, such as https://www.ipify.org/

ducaale (approved):
Thanks!

PR #278, opened 2022-09-11
Enable ipv6 tests in CI

ducaale:


ducaale:
Closing this, since ipv6 tests don't seem to work inside GitHub Actions

https://github.com/ducaale/xh/runs/8295134630?check_suite_focus=true

Issue #277, opened 2022-09-11
Warn users when translating a command that uses `--follow` + a non-GET method


ducaale:
As reported in #270, there is a difference between how curl and xh behave when following redirects and using a non-GET method:

```sh
$ xh post https://httpbin.org/redirect-to url==https://httpbin.org/get --follow --body
{
    "args": {},
    "headers": {
        "Accept": "*/*",
        "Accept-Encoding": "gzip, deflate, br",
        "Host": "httpbin.org",
        "User-Agent": "xh/0.16.1",
        "X-Amzn-Trace-Id": "Root=1-632ee6dd-26529e365e6971a258aa60c5"
    },
    "origin": "92.40.178.172",
    "url": "https://httpbin.org/get"
}

# Use xh's --curl to translate the prev command to curl (ignoring --body)
$ xh post https://httpbin.org/redirect-to url==https://httpbin.org/get --follow --curl
curl -L -X POST 'https://httpbin.org/redirect-to?url=https%3A%2F%2Fhttpbin.org%2Fget'

# Now, let's run the output from the translation
$ curl -L -X POST 'https://httpbin.org/redirect-to?url=https%3A%2F%2Fhttpbin.org%2Fget'
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```

This is due to the `-X` flag, which instructs curl to retain the original HTTP method regardless of the redirect code (see `Why superfluous -X usage is bad` section from https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/).

Given that people relying on xh's curl translation feature might be caught off guard, we should output a warning when translating a combination of `--follow` and non-GET HTTP method:

```
$ xh post https://httpbin.org/redirect-to url==https://httpbin.org/get --follow --curl
Warning: Using a combination of -X/--request and -L/--location which may cause unintended side effects 

curl -L -X POST 'https://httpbin.org/redirect-to?url=https%3A%2F%2Fhttpbin.org%2Fget'
```

jgoday:
Hi, just playing with this awesome project !

created a very simple PR (#279) related to this item.

ducaale:
Thanks, @jgoday.

[...]

PR #280, opened 2022-09-24
Warn users when --curl and --follow + non GET method 

jgoday:
see #277 

It adds following warning when --curl (translating to culr) and --follow is used in combination with a non GET method.
```
Using a combination of -X/--request and -L/--location which may cause unintended side effects.
```

Added basic test (check_non_get_redirect_warning).

### DIFF
--- a/src/to_curl.rs
+++ b/src/to_curl.rs
@@ -118,7 +118,7 @@ pub fn translate(args: Cli) -> Result<Command> {
         }
     }
 
-    if args.follow && !matches!(args.method, Some(Method::GET) | None){
+    if args.follow && !matches!(args.method, Some(Method::GET) | None) {
         cmd.warn("Using a combination of -X/--request and -L/--location which may cause unintended side effects.");
     }
 
--- a/src/to_curl.rs
+++ b/src/to_curl.rs
@@ -118,7 +118,7 @@ pub fn translate(args: Cli) -> Result<Command> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/to_curl.rs
```diff
@@ -118,6 +118,10 @@ pub fn translate(args: Cli) -> Result<Command> {
         }
     }
 
+    if args.curl && args.follow && args.method.as_ref().map(|m| m != &Method::GET).unwrap_or(false) {
```
ducaale, line 121: Can we use the `matches!()` macro to simplify things a little bit?

```rust
if args.follow && !matches!(args.method, Some(Method::GET) | None) { /* ... */ }
```

We can also skip checking `args.curl` since `to_curl::print_curl_translation` is only called when `args.curl`  is true.

tests/cli.rs
```diff
@@ -3056,3 +3056,11 @@ fn empty_response_with_content_encoding_and_content_length() {
 
         "#});
 }
+
+#[test]
+fn check_non_get_redirect_warning() {
```
ducaale, line 3061:
```suggestion
fn non_get_redirect_translation_warning() {
```

src/to_curl.rs
jgoday: Sure, changed in the latest commit. Should I squeash it ?

ducaale (approved):
Thanks!

PR #281, opened 2022-09-25
allow to build with native-tls only

jirutka:
Bundling (static linking) TLS library is a very bad idea from the security perspective and Linux distributions try to avoid it. This will allow to build xh with native-tls (i.e. dynamically linked system-provided TLS library) only, without rustls.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,7 +28,11 @@ fn feature_status(feature: &str) -> String {
 }
 
 fn features() -> String {
-    feature_status("native-tls")
+    format!(
+        "{} {}",
+        &feature_status("native-tls"),
+        &feature_status("rustls")
+    )
 }
 
 fn main() {
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,9 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             flags: --features=native-tls
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            flags: --no-default-features --features=native-tls,online-tests # disables rustls
           - target: x86_64-apple-darwin
             os: macos-latest
             flags: --features=native-tls
@@ -72,7 +75,7 @@ jobs:
           command: fmt
           args: -- --check
 
-      - name: Clippy
+      - name: Clippy (default features)
         uses: actions-rs/cargo@v1
         with:
           command: clippy
@@ -83,3 +86,9 @@ jobs:
         with:
           command: clippy
           args: --all-features --tests -- -D warnings -A unknown-lints
+
+      - name: Clippy (native-tls only)
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --no-default-features --features=native-tls,online-tests --tests -- -D warnings -A unknown-lints
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,6 +14,9 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             flags: --features=native-tls
+          - target: x86_64-unknown-linux-gnu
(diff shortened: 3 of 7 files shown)


ducaale:
Can you target `develop` instead of `master` for your PR? Thanks

build.rs
```diff
@@ -14,24 +14,35 @@ fn build_syntax(dir: &str, out: &str) {
...
+        "{} {}",
+        &feature_status("native-tls"),
+        &feature_status("rustls")
+    )
 }
 
 fn main() {
+    if !feature_enabled("native-tls") && !feature_enabled("rustls") {
+        panic!("Either native-tls or rustls feature must be enabled!");
+    }
```
blyxxyz, line 44: Maybe this should be a compile-time error? Not sure what the best practices for that are.

ducaale: Maybe we could use the [`compile_error!()`](https://doc.rust-lang.org/std/macro.compile_error.html) macro?

#### `panic!()` inside build.rs
```
$ cargo run --no-default-features -- --help
   Compiling xh v0.16.1 (/home/ducaale/xh)
   Compiling reqwest v0.11.10
error: failed to run custom build command for `xh v0.16.1 (/home/ducaale/xh)`

Caused by:
  process didn't exit successfully: `/home/ducaale/xh/target/debug/build/xh-92f620f356085f57/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'Either native-tls or rustls feature must be enabled!', build.rs:43:9
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

#### `compile_error!()` inside build.rs
```
$ cargo run --no-default-features -- --help
   Compiling xh v0.16.1 (/home/ducaale/xh)
error: Either native-tls or rustls feature must be enabled!
  --> build.rs:43:5
   |
43 |     compile_error!("Either native-tls or rustls feature must be enabled!");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `xh` due to previous error
```

jirutka: > Not sure what the best practices for that are.

Me neither.

> Maybe we could use the `compile_error!()` macro?

Yeah, this looks like the right macro for this use case! I’ve updated the PR.

ducaale (commented):
Looks good. I've a couple of comments mostly around tests and CI:

1. Similar to `cert`, can you disable the `--verify` logic if rustls is not enabled? otherwise, it errors at runtime.
2. We should disable `improved_https_ip_error_with_support` unless both `rustls` and `native-tls` are enabled
3. I wonder if we can change one of the targets (e.g `x86_64-unknown-linux-gnu`) in both `ci.yaml` and `release.yaml` to run tests while rustls is disabled? or maybe we could add a new entry?
4. Can we have `Clippy (native-tls only)` in `ci.yaml` to check for warnings when rustls is disabled?

jirutka:
> 1. Similar to cert, can you disable the --verify logic if rustls is not enabled? otherwise, it errors at runtime.

This is already handled, isn’t it? It’s the same situation as when `--native-tls` is used.

jirutka:
@ducaale, all points resolved, just unsure about the first one.

ducaale (approved):
LGTM, thanks! 

>> 1. Similar to cert, can you disable the --verify logic if rustls is not enabled? otherwise, it errors at runtime.
>
>This is already handled, isn’t it? It’s the same situation as when `--native-tls` is used.

Yup, you're right. For some reason, I thought native-tls was falling back to rustls when using `--verify`. But, it turns out that I forgot to pass the `--native-tls` flag when I was comparing this PR against the develop branch.

jirutka:
These `fix up!` were meant to be squashed before merging.

ducaale:
Thanks for pointing that out. `fixup!` commits have now been squashed in the `develop` branch.

PR #282, opened 2022-09-26
Switch to features resolver "2" and disable unnecessary/unused features

jirutka:
See individual commit messages.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ once_cell = "1.8.0"
 os_display = "0.1.3"
 pem = "0.8.2"
 rand = "0.8.3"
-regex = "1"
+regex = { version = "1", default-features = false, features = ["std"] }
 rpassword = "5.0.0"
 serde = "1.0"
 serde_json = { version = "1.0", features = ["preserve_order"] }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -54,6 +55,11 @@ features = ["rustls-tls", "rustls-tls-webpki-roots", "rustls-tls-native-roots",
 [dependencies.syntect]
 version = "4.4"
 default-features = false
+features = ["parsing", "html", "dump-load", "regex-onig"]
```
ducaale, line 58:
```suggestion
features = ["parsing", "dump-load", "regex-onig"]
```

Cargo.toml
```diff
@@ -54,6 +55,11 @@ features = ["rustls-tls", "rustls-tls-webpki-roots", "rustls-tls-native-roots",
...
+
+[build-dependencies.syntect]
+version = "4.4"
+default-features = false
 features = ["parsing", "html", "yaml-load", "dump-load", "dump-create", "regex-onig"]
```
ducaale, line 63:
```suggestion
features = ["yaml-load", "dump-create", "regex-onig"]
```

ducaale (changes requested):
Thanks, I was initially a bit concerned about removing the `perf` feature from the regex crate, but I couldn't find any noticeable difference.

ducaale (approved):
Thanks!

PR #285, opened 2022-09-30
Replace ipify.org with httpbin.org/ip

ducaale:


ducaale:
Superseded by https://github.com/ducaale/xh/pull/286

PR #286, opened 2022-10-01
Switch from ipify.org to icanhazip.com

ducaale:
The latter is currently operated by Cloudflare and should be more [reliable](https://github.com/ducaale/xh/actions/runs/3131272389/jobs/5082439970) than ipify.org.

Also, see https://major.io/icanhazip-com-faq/


ducaale:
Let's revisit this once requests to ipify.org start to fail again

PR #288, opened 2022-10-08
Support reading query param and header values from a file

ducaale:
This PR adds variants of `:` and `==` with `@` postfix for reading values from a file.

```sh
# send a query param with a value read from temp.txt
$ xh httpbin.org/get test==@~/temp.txt

# send a header with a value read from temp.txt
$ xh httpbin.org/get x-api-key:@~/temp.txt
```

 Also, see https://github.com/httpie/httpie/pull/1225.

Ticks another item from https://github.com/ducaale/xh/issues/4.


blyxxyz (approved):
Works great!

PR #294, opened 2022-11-29
Add request and keep encoding for download body flag

trueb2:
## Changes

Add a `--download-encoding` flag to choose an Accept-Encoding header other than `"identity"` during a download request. During the download, the "decompress" function that reads the response and writes it to the output will passthrough data in whatever Content-Encoding was received if `--download-encoding` was specified.

## Motivation
When downloading a large file, I want to be able to transfer it compressed and leave it compressed. For example, stream download a 500MiB CSV file in the gzip encoding, saving it to a data.csv.gz file.

```
jwtrueb@jmbp xh % cargo run --release -- get http://localhost:3000/download/csv/asdf -d --download-encoding gzip
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Type: application/csv
Date: Tue, 29 Nov 2022 23:17:12 GMT
Transfer-Encoding: chunked
Vary: accept-encoding

Downloading to "asdf"
Done. 147B in 0.00177s (80.92KiB/s)
jwtrueb@jmbp xh % gzcat asdf
ts,user_id,device_id,source,pressure,temperature,humidity
2022-11-28T20:13:12.704030+00:00,2,5,1,99999.0,25.3,16.4
2022-11-28T20:14:02.446438+00:00,2,5,1,99699.0,25.1,16.2
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -159,6 +159,10 @@ Defaults to \"format\" if the NO_COLOR env is set and to \"none\" if stdout is n
     #[clap(short = 'd', long)]
     pub download: bool,
 
+    /// During download, request a content encoding and skip decompression.
+    #[clap(long)]
+    pub download_encoding: Option<String>,
+
     /// Resume an interrupted download. Requires --download and --output.
     #[clap(
         short = 'c',
--- a/src/download.rs
+++ b/src/download.rs
@@ -160,6 +160,7 @@ const UNCOLORED_SPINNER_TEMPLATE: &str = "{spinner} {bytes} {bytes_per_sec} {wid
 
 pub fn download_file(
     mut response: Response,
+    do_decompress: bool,
     file_name: Option<PathBuf>,
     // If we fall back on taking the filename from the URL it has to be the
     // original URL, before redirects. That's less surprising and matches
@@ -244,9 +245,13 @@ pub fn download_file(
         pb.reset_eta();
     }
 
+    let compression_type = if do_decompress {
+        get_compression_type(response.headers())
+    } else {
+        None
+    };
     match pb {
         Some(ref pb) => {
-            let compression_type = get_compression_type(response.headers());
             copy_largebuf(
                 &mut decompress(&mut pb.wrap_read(response), compression_type),
                 &mut buffer,
@@ -267,7 +272,6 @@ pub fn download_file(
             }
         }
         None => {
-            let compression_type = get_compression_type(response.headers());
             copy_largebuf(
                 &mut decompress(&mut response, compression_type),
                 &mut buffer,
--- a/src/main.rs
+++ b/src/main.rs
@@ -465,9 +465,14 @@ fn run(args: Cli) -> Result<i32> {
     };
 
     if args.download {
-        request
-            .headers_mut()
-            .insert(ACCEPT_ENCODING, HeaderValue::from_static("identity"));
+        let encoding = args
+            .download_encoding
+            .clone()
+            .unwrap_or_else(|| "identity".to_string());
+        request.headers_mut().insert(
+            ACCEPT_ENCODING,
+            HeaderValue::from_str(encoding.as_str()).unwrap(),
+        );
     }
 
     let buffer = Buffer::new(
@@ -555,6 +560,7 @@ fn run(args: Cli) -> Result<i32> {
             if exit_code == 0 {
                 download_file(
                     response,
+                    args.download_encoding.is_none(),
                     args.output,
                     &args.url,
                     resume,


blyxxyz:
One reason to use `Accept-Encoding: identity` is to be able to resume downloads (with `--continue`). We'd have to figure out how these features interact. Maybe that should be an error message.

This option seems too specific. Maybe we should allow overriding the encoding via a normal header argument and add a `--preserve-encoding`/`--no-decode`/`--verbatim-body`/whatever option? Then you'd be able to run `xh -d --preserve-encoding :3000/download/csv/asdf accept-encoding:gzip`.

We have to coordinate this with HTTPie, that's where we got the current behavior. Are you interested in making an issue?

trueb2:
> One reason to use `Accept-Encoding: identity` is to be able to resume downloads (with `--continue`). We'd have to figure out how these features interact. Maybe that should be an error message.
> 
> This option seems too specific. Maybe we should allow overriding the encoding via a normal header argument and add a `--preserve-encoding`/`--no-decode`/`--verbatim-body`/whatever option? Then you'd be able to run `xh -d --preserve-encoding :3000/download/csv/asdf accept-encoding:gzip`.
> 
> We have to coordinate this with HTTPie, that's where we got the current behavior. Are you interested in making an issue?

Sure whatever works. I couldn't figure out how to get HTTPie to do what I wanted as easily as xh. Looks like it wouldn't be too bad to switch this to a `--preseve-encoding` and a manual header. That still seems to be in the spirit of the CLI.

I don't really care to use `--continue` for downloads, so I did not consider how that interacts with anything other than identity. I think an error message makes sense if both `--continue` and `--preserve-encoding` are used.

I could just implement that instead of making an issue. Is it a problem to have features in xh that aren't in HTTPie?

ducaale:
>Is it a problem to have features in xh that aren't in HTTPie?

We are open to adding features not supported in HTTPie. But we don't want to end up in a situation where the same feature is implemented differently in xh and HTTPie. Many xh users are coming from HTTPie, and some of them might even rely on [HTTPie docs](https://httpie.io/docs/cli) while using xh.

Let's either open a new issue or maybe leave a comment on https://github.com/httpie/httpie/issues/178, and then we can continue with `--preserve-encoding`/`--no-decode`/`--verbatim-body`/something else.

trueb2:
@ducaale That is quite the old issue you dug up! I see it is related to https://github.com/ducaale/xh/pull/130

I think we could comment on the xh approach in the httpie issue, then implement an approach. It seems like this isn't a very hot feature request, and I don't want it to end up as another issue that goes 9 years with an implementation as easy as this. What is your preferred implementation/flag?

ducaale:
We can go with the proposed implementation in https://github.com/ducaale/xh/pull/294#issuecomment-1331882126 and use `--preserve-encoding` (cURL uses `--raw` for a similar thing but that option already used in HTTPie/xh).

Note that we currently don't allow the user to set `accept-encoding` in download mode (see https://github.com/ducaale/xh/pull/141). We can roll back that behavior and error when `--continue` is used with custom a `accept-encoding`.

src/cli.rs
```diff
@@ -159,6 +159,12 @@ Defaults to \"format\" if the NO_COLOR env is set and to \"none\" if stdout is n
     #[clap(short = 'd', long)]
     pub download: bool,
 
+    /// During download, keep the raw encoding of the body. Requires --download.
```
ducaale, line 162: It would be nice to use this option outside of download mode e.g to debug a server that returns an incorrect `content-encoding` header.

Can you disable the decompression in `Printer::print_response_body` when this flag is used?

ducaale (changes requested):

src/cli.rs
trueb2: Good idea and thanks for the feedback. I believe I updated this as you wanted.

src/cli.rs
```diff
@@ -159,6 +159,13 @@ Defaults to \"format\" if the NO_COLOR env is set and to \"none\" if stdout is n
     #[clap(short = 'd', long)]
     pub download: bool,
 
+    /// During download, keep the raw encoding of the body. Intended for use with download or
+    /// to debug an encoded response body.
+    ///
+    /// For example, set Accept-Encoding: gzip and use --preserve-encoding to skip decompression.
+    #[clap(long)]
+    pub preserve_encoding: bool,
+
```
ducaale, line 168: Let's shorten the first line to fit within ~80 chars. Any extended help can go in subsequent lines.

(to see the difference, try `cargo run -- --help` and `cargo run -- help`)

```rs
/// Skip decompressing the response body
///
/// longer help
```

Also, should we move this option to after `response_mime` to improve discoverability?

src/printer.rs
```diff
@@ -428,15 +428,33 @@ impl Printer {
...
         let content_type =
             mime.map_or_else(|| get_content_type(response.headers()), ContentType::from);
         let encoding = encoding.or_else(|| get_charset(response));
-        let compression_type = get_compression_type(response.headers());
+        let (may_decode, compression_type) = if !preserve_encoding {
+            let compression_type = get_compression_type(response.headers());
+            (true, compression_type)
+        } else {
+            (false, None)
+        };
```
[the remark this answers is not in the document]
trueb2: I don't think that it is possible since the logic below assumes that the content has been decompressed before colorizing, formatting, or decoding. We wouldn't want that logic to run on potentially encoded data. The other branches call code like decode_stream to interpret UTF8 or other format types.

PR #295, opened 2022-12-12
Fix panic when parsing connection timeout

sorairolake:
Panic if seconds of connection timeout is greater than `Duration::MAX` or not finite (see [`Duration::from_secs_f64`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs_f64)), so fix this.

Note this can be simplified when [`Duration::try_from_secs_f64`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64) is stabilized.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1070,13 +1070,20 @@ impl FromStr for Timeout {
     type Err = anyhow::Error;
 
     fn from_str(sec: &str) -> anyhow::Result<Timeout> {
-        let pos_sec: f64 = match sec.parse::<f64>() {
-            Ok(sec) if sec.is_sign_positive() => sec,
-            _ => return Err(anyhow!("Invalid seconds as connection timeout")),
-        };
-
-        let dur = Duration::from_secs_f64(pos_sec);
-        Ok(Timeout(dur))
+        match f64::from_str(sec) {
+            Ok(s) => {
+                if !s.is_finite() {
+                    Err(anyhow!("Connection timeout is not finite"))
+                } else if s.is_sign_negative() {
+                    Err(anyhow!("Connection timeout is negative"))
+                } else if s >= Duration::MAX.as_secs_f64() {
+                    Err(anyhow!("Connection timeout is too big"))
+                } else {
+                    Ok(Timeout(Duration::from_secs_f64(s)))
+                }
+            }
+            _ => Err(anyhow!("Connection timeout is not float value")),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -1070,13 +1070,20 @@ impl FromStr for Timeout {
...
+                } else if s.is_sign_negative() {
+                    Err(anyhow!("Connection timeout is negative"))
+                } else if s >= Duration::MAX.as_secs_f64() {
+                    Err(anyhow!("Connection timeout is too big"))
+                } else {
+                    Ok(Timeout(Duration::from_secs_f64(s)))
+                }
+            }
+            _ => Err(anyhow!("Connection timeout is not float value")),
+        }
```
blyxxyz, line 1086:
```suggestion
        match f64::from_str(sec) {
            Ok(s) if !s.is_nan() => {
                if s.is_sign_negative() {
                    Err(anyhow!("Connection timeout is negative"))
                } else if s >= Duration::MAX.as_secs_f64() || !s.is_finite() {
                    Err(anyhow!("Connection timeout is too big"))
                } else {
                    Ok(Timeout(Duration::from_secs_f64(s)))
                }
            }
            _ => Err(anyhow!("Connection timeout is not a valid number")),
        }
```

- NaN is caught by the `is_finite` check but that's not really the right error message.

- `s` also becomes infinite if you enter a really large number, so I think it's best to merge those cases.

- I changed the fallback message to one that's a little less technical.

blyxxyz (changes requested):
Nice catch!

Two concerns:

- This solution rejects `-0`, while HTTPie accepts it. Do we want that?

- If the timeout is too large then maybe we should set it to `Duration::MAX` instead? That's still a couple billion years so the user won't be able to tell :slightly_smiling_face:

src/cli.rs
sorairolake: I change `!s.is_finite()` to `s.is_infinite()`. Because it will never be NaN in this case.

sorairolake:
If the concerns need to be solved, I will do it.

blyxxyz (approved):
Looks good. Maybe @ducaale has an opinion about the points from earlier but I don't think they're very important.

ducaale:
>Two concerns:
>
>- This solution rejects -0, while HTTPie accepts it. Do we want that?
>- If the timeout is too large then maybe we should set it to Duration::MAX instead? That's still a couple billion years so the user won't be able to tell 🙂

It would've been if the first point could be addressed but I don't see it as a blocker 👍

Issue #296, opened 2022-12-13
Feature request: Ignore PROXY ENVIRONMENT on Windows


oryon-dominik:
xh behaves differently from curl considering PROXY environment variables.
On Windows xh reads proxy settings form the registry (that's unexpected from curl).

Need an option to force --NO-PROXY.

ducaale:
It is most likely that curl will support [system proxy on windows](https://github.com/curl/curl/issues/3965) in the future. 

We are open to adding something similar to cURL's `--noproxy` option but In the meantime, you can use `NO_PROXY` env variable to prevent xh from using the system proxy

```pwsh
PS> $env:NO_PROXY = '*'
PS> xh example.com
```

PR #297, opened 2022-12-13
Document `NO_PROXY` env variable

ducaale:
This is something I noticed while looking into https://github.com/ducaale/xh/issues/296.

I also wanted to document `ALL_PROXY` but that [isn't supported in reqwest](https://github.com/seanmonstar/reqwest/issues/1414) yet. Perhaps we could send a PR to reqwest or add to it xh manually?

### DIFF
--- a/doc/man-template.roff
+++ b/doc/man-template.roff
@@ -88,8 +88,8 @@ Sets the proxy server to use for HTTP.
 Sets the proxy server to use for HTTPS.
 .TP
 .B NO_PROXY
-List of comma seperated host names where proxy server is ignored. Use "*"
-to match all host names.
+List of comma-separated hosts for which to ignore the other proxy environment
+variables. "*" matches all host names.
 .TP
 .B NETRC
 Location of the .netrc file.
--- a/doc/man-template.roff
+++ b/doc/man-template.roff
@@ -81,11 +81,15 @@ environment variable is to rename the binary to either http or https.
 .BR REQUESTS_CA_BUNDLE ", " CURL_CA_BUNDLE
 Sets a custom CA bundle path.
 .TP
+.B http_proxy
+Sets the proxy server to use for HTTP.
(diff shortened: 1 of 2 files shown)


doc/man-template.roff
```diff
@@ -81,11 +81,15 @@ environment variable is to rename the binary to either http or https.
 .BR REQUESTS_CA_BUNDLE ", " CURL_CA_BUNDLE
 Sets a custom CA bundle path.
 .TP
+.B http_proxy
+Sets the proxy server to use for HTTP.
+.TP
 .B HTTPS_PROXY
 Sets the proxy server to use for HTTPS.
 .TP
```
blyxxyz, line 89: While we're at it we should describe the format of the values:
```suggestion
.BR http_proxy "=[protocol://]<host>[:port]"
Sets the proxy server to use for HTTP.
.TP
.BR HTTPS_PROXY "=[protocol://]<host>[:port]"
Sets the proxy server to use for HTTPS.
.TP
```

doc/man-template.roff
```diff
@@ -81,11 +81,15 @@ environment variable is to rename the binary to either http or https.
...
-.B http_proxy
-Sets the proxy server to use for HTTP.
+.B NO_PROXY
+List of comma seperated host names where proxy server is ignored. Use "*"
+to match all host names.
```
blyxxyz, line 92:
```suggestion
List of comma-separated hosts for which to ignore the other proxy environment
variables. "*" matches all host names.
```
It doesn't seem to affect the `--proxy` option.

The full rules are [annoyingly complicated](https://github.com/seanmonstar/reqwest/blob/5656fb861635a05fcd0ea32ab5251d954414a39f/src/proxy.rs#L384) and probably not worth documenting in our man page.

blyxxyz (approved):

PR #308, opened 2023-03-20
to_curl: Support --raw flag

### LINKED ISSUE #252: `--raw` is ignored when translating xh commands using `--curl`
#### Expected
```sh
$ xh httpbin.org/post --raw=hello --curl
# Not sure if content-type should be set if body is invalid JSON
curl http://httpbin.org/post -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d 'hello'
```

#### Actual
```sh
$ xh httpbin.org/post --raw=hello --curl
curl http://httpbin.org/post
```

ducaale:
Resolves #252

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -59,7 +59,7 @@ pub struct Cli {
     /// Like --form, but force a multipart/form-data request even without files.
     ///
     /// Overrides both --json and --form.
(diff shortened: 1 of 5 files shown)


blyxxyz (approved):
Looks good!

PR #313, opened 2023-05-07
Support duplicate header keys in session files

### LINKED ISSUE #245: Handle saving multiple headers with same key in sessions
See https://github.com/httpie/httpie/pull/1335

ducaale:
Headers will be saved as key-value pairs, and any session using the old format will be migrated automatically

#### Before:
```json
{
  "__meta__": { "about": "xh session file", "xh": "0.0.0" },
  "auth": { "type": null, "raw_auth": null },
  "cookies": {},
  "headers": {
    "hello": "world"
  }
}
```

#### After:
```json
{
  "__meta__": { "about": "xh session file", "xh": "0.0.0" },
  "auth": { "type": null, "raw_auth": null },
  "cookies": {},
  "headers": [
    { "key": "hello", "value": "world" }
  ]
}
```

Resolves #245

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -321,8 +321,13 @@ fn run(args: Cli) -> Result<i32> {
 
     if let Some(ref mut s) = session {
         auth = s.auth()?;
-        for (key, value) in s.headers()?.iter() {
-            headers.entry(key).or_insert_with(|| value.clone());
+
+        let mut session_headers = s.headers()?;
+        for key in headers.keys() {
+            session_headers.remove(key);
+        }
+        for (key, value) in session_headers.iter() {
+            headers.append(key, value.clone());
         }
         s.save_headers(&headers)?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -322,13 +322,11 @@ fn run(args: Cli) -> Result<i32> {
     if let Some(ref mut s) = session {
         auth = s.auth()?;
 
-        let mut session_headers = s.headers()?;
-        for key in headers.keys() {
-            session_headers.remove(key);
-        }
-        for (key, value) in session_headers.iter() {
-            headers.append(key, value.clone());
-        }
+        headers = {
+            let mut session_headers = s.headers()?;
+            session_headers.extend(headers);
+            session_headers
+        };
         s.save_headers(&headers)?;
 
         let mut cookie_jar = cookie_jar.lock().unwrap();
--- a/src/session.rs
+++ b/src/session.rs
@@ -295,124 +295,69 @@ fn path_from_url(url: &Url) -> Result<String> {
 mod tests {
     use super::*;
 
+    use anyhow::Result;
     use reqwest::header::HeaderValue;
 
-    use crate::utils::random_string;
-    use anyhow::Result;
+    fn load_session_from_str(s: &str) -> Result<Session> {
+        Ok(Session {
+            content: serde_json::from_str::<Content>(s)?.migrate(),
+            path: PathBuf::new(),
+            read_only: false,
+        })
+    }
 
     #[test]
-    fn can_read_httpie_session_file() -> Result<()> {
-        let mut path_to_session = std::env::temp_dir();
-        let file_name = random_string();
-        path_to_session.push(file_name);
-        fs::write(
-            &path_to_session,
-            indoc::indoc! {r#"
-                {
-                    "__meta__": {
-                        "about": "HTTPie session file",
-                        "help": "https://httpie.org/doc#sessions",
-                        "httpie": "2.3.0"
-                    },
-                    "auth": {
-                        "password": null,
-                        "type": null,
-                        "username": null
-                    },
-                    "cookies": {
-                        "__cfduid": {
-                            "expires": 1620239688,
-                            "path": "/",
-                            "secure": false,
-                            "value": "d090ada9c629fc7b8bbc6dba3dde1149d1617647688"
-                        }
-                    },
-                    "headers": {
-                        "hello": "world"
-                    }
-                }
-            "#},
-        )?;
-
-        let session = Session::load_session(
-            &Url::parse("http://localhost")?,
-            path_to_session.into(),
-            false,
-        )?;
+    fn can_parse_old_httpie_session() -> Result<()> {
+        let session = load_session_from_str(indoc::indoc! {r#"
+            {
+                "__meta__": {
+                    "about": "HTTPie session file",
+                    "help": "https://httpie.org/doc#sessions",
+                    "httpie": "2.3.0"
+                },
+                "auth": { "password": null, "type": null, "username": null },
+                "cookies": {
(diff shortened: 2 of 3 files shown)


src/session.rs
```diff
@@ -108,9 +142,17 @@ impl Session {
...
             if key != "cookie" && !key.starts_with("content-") && !key.starts_with("if-") {
-                self.content
-                    .headers
-                    .insert(key.into(), value.to_str()?.into());
+                match self.content.headers {
+                    Headers::Map(ref mut headers) => {
+                        headers.insert(key.into(), value.to_str()?.into());
+                    }
+                    Headers::List(ref mut headers) => {
+                        headers.push(Header {
```
ducaale, line 150: **Note to self**: overwrite old entries to avoid accumulation of headers

src/session.rs
```diff
@@ -49,18 +49,46 @@ pub struct Cookie {
...
+struct Header {
+    key: String,
+    value: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+enum Headers {
+    Map(HashMap<String, String>),
+    List(Vec<Header>),
```
blyxxyz, line 62: Can you add a note explaining that this is to support duplicate keys, and that we keep `Map` around for backward compatibility?

src/session.rs
blyxxyz: Do you mean clearing the container before filling it or something else?

src/session.rs
```diff
@@ -49,18 +49,46 @@ pub struct Cookie {
     secure: Option<bool>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+struct Header {
+    key: String,
```
blyxxyz, line 54: HTTPie calls these `name`, not `key`, so they fail to load each other's sessions. Maybe we can copy a file from HTTPie's test suite?

blyxxyz (commented):
When the session file has multiple headers only the first one is sent. Replacing `headers.entry(key).or_insert_with(|| value.clone());` by `headers.append(key, value.clone());` in `main.rs` fixes it. (Could probably use a test!)

src/session.rs
ducaale: Yeah, I meant clearing the container. So right now, If I make multiple requests with different `Authorization` headers, old values are not being removed

```json
{
    "__meta__": {
        "about": "xh session file",
        "xh": "0.18.0"
    },
    "auth": {
        "type": null,
        "raw_auth": null
    },
    "cookies": {},
    "headers": [
        {
            "key": "authorization",
            "value": "bearer secret-1"
        },
        {
            "key": "authorization",
            "value": "bearer secret-2"
        },
        {
            "key": "hello",
            "value": "world"
        }
    ]
}
```

src/session.rs
ducaale: Yup, that is a good idea 👍

ducaale: **Update:** I haven't used a verbatim test file from HTTPie since we are not yet 100% compatible with the current session format. I am planning to address this in a follow-up PR that adds support for the new style cookies introduced in https://github.com/httpie/httpie/pull/1312

src/main.rs
```diff
@@ -343,13 +343,11 @@ fn run(args: Cli) -> Result<i32> {
...
-            session_headers.remove(key);
-        }
-        for (key, value) in session_headers.iter() {
-            headers.append(key, value.clone());
-        }
+        headers = {
+            let mut session_headers = s.headers()?;
+            session_headers.extend(headers);
+            session_headers
+        };
```
blyxxyz, line 350: Don't we still need to remove those keys?

src/session.rs
```diff
@@ -98,19 +128,37 @@ impl Session {
...
             // see https://github.com/httpie/httpie/commit/e09b74021c9c955fd7c3bab11f22801aaf9dc1b8
             // we will also ignore cookies as they are taken care of by save_cookies()
             if key != "cookie" && !key.starts_with("content-") && !key.starts_with("if-") {
-                self.content
-                    .headers
-                    .insert(key.into(), value.to_str()?.into());
+                match self.content.headers {
+                    Headers::Map(ref mut headers) => {
+                        headers.insert(key.into(), value.to_str()?.into());
+                    }
```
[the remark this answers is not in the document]
blyxxyz: `unreachable!()` sounds fine, as long as we have a test that loads the old format to make sure. This code is a bit misleading.

It would be nice to just not have this case at all but none of the solutions I can think of are a net improvement.

src/main.rs
ducaale: I think `.extend()` should take of it. See the example in https://docs.rs/http/latest/http/header/struct.HeaderMap.html#impl-Extend%3C(Option%3CHeaderName%3E,+T)%3E-for-HeaderMap%3CT%3E

```rust
let mut map = HeaderMap::new();

map.insert(ACCEPT, "text/plain".parse().unwrap());
map.insert(HOST, "hello.world".parse().unwrap());

let mut extra = HeaderMap::new();

extra.insert(HOST, "foo.bar".parse().unwrap());
extra.insert(COOKIE, "hello".parse().unwrap());
extra.append(COOKIE, "world".parse().unwrap());

map.extend(extra);

assert_eq!(map["host"], "foo.bar");
assert_eq!(map["accept"], "text/plain");
assert_eq!(map["cookie"], "hello");

let v = map.get_all("host");
assert_eq!(1, v.iter().count());

let v = map.get_all("cookie");
assert_eq!(2, v.iter().count());
```

src/session.rs
blyxxyz: > not yet 100% compatible

Do you mean that we crash or that we interpret the session files incorrectly?

ducaale: Actually, there are test files that mix the new header format with the old cookie format. See https://github.com/httpie/httpie/blob/master/tests/fixtures/session_data/old/cookies_dict_with_extras.json

src/session.rs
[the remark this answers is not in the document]
ducaale: Done

src/main.rs
blyxxyz: Ahh, got it. I was looking at the wrong Extend impl.

src/session.rs
```diff
@@ -383,4 +430,48 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    fn can_read_session_with_duplicate_keys() -> Result<()> {
+        let mut path_to_session = std::env::temp_dir();
+        let file_name = random_string();
+        path_to_session.push(file_name);
+        fs::write(
```
ducaale, line 439: Should we remove file operations and focus on just the permutations of session content that xh can load? Our integration tests should cover the rest (I think).

blyxxyz: Yeah, I don't think this is the right place to test the I/O.

Maybe the `load_session` constructor should be split into two layers so you can easily feed it a string.

blyxxyz (approved):
Looks good! I didn't find any differences compared to HTTPie's behavior.

Issue #244, opened 2022-04-10
Revisit how cookies are handled when using anonymous sessions


ducaale:
See https://github.com/httpie/httpie/issues/1004 and https://github.com/httpie/httpie/pull/1312

PR #314, opened 2023-05-29
Support persisting cookies from multiple domains

ducaale:
This can be useful when reusing anonymous sessions for multiple domains or when cookies are set as users are redirected from one domain to another.

#### Before:
```json
{
  "__meta__": { "about": "xh session file", "xh": "0.0.0" },
  "auth": { "type": null, "raw_auth": null },
  "cookies": {
    "lang": { "value": "en" }
  },
  "headers": []
}
```

#### After:
```json
{
  "__meta__": { "about": "xh session file", "xh": "0.0.0" },
  "auth": { "type": null, "raw_auth": null },
  "cookies": [
    { "name": "lang", "value": "en", "domain": "example.com" },
    { "name": "lang", "value": "ar", "domain": "example.org" }
  ],
  "headers": []
}
```

Resolves https://github.com/ducaale/xh/issues/244


src/session.rs
```diff
@@ -218,39 +264,55 @@ impl Session {
...
+                expires: cookie
+                    .expires()
+                    .and_then(|v| v.datetime())
+                    .map(|v| v.unix_timestamp()),
+                path: cookie.path().map(Into::into),
+                secure: cookie.secure(),
+                domain: cookie
+                    .domain()
+                    .map(Into::into)
+                    .or_else(|| self.host.clone()),
```
ducaale, line 314: 1. What if the cookie was set by a domain different from `self.host`?
2. The Cookie domain cannot be an IP address e.g 127.0.0.1. See https://stackoverflow.com/a/30676300/5915221

src/main.rs
```diff
@@ -578,9 +592,16 @@ fn run(args: Cli) -> Result<i32> {
         let cookie_jar = cookie_jar.lock().unwrap();
         s.save_cookies(
             cookie_jar
-                .matches(&url)
-                .into_iter()
-                .map(|c| cookie_crate::Cookie::from(c.clone()))
+                .iter_unexpired()
+                .map(|c| {
```
ducaale, line 596: **Note to self:** change `session::save_cookies()` arg type to `cookie_store::Cookie` and move this logic to `session.rs`

src/main.rs
```diff
@@ -330,15 +332,27 @@ fn run(args: Cli) -> Result<i32> {
         s.save_headers(&headers)?;
 
         let mut cookie_jar = cookie_jar.lock().unwrap();
-        for cookie in s.cookies() {
-            match cookie_jar.insert_raw(&cookie, &url) {
-                Ok(..) | Err(cookie_store::CookieError::Expired) => {}
+        for mut cookie in s.cookies()? {
```
ducaale, line 335: Similar to above, return `cookie_store::Cookie` instead of `cookie:Cookie`.

src/session.rs
```diff
@@ -97,19 +126,35 @@ impl Content {
...
+            )
+        }
 
         self
     }
 }
 
-pub struct Session {
+pub struct Session<'a> {
+    url: &'a Url,
```
blyxxyz, line 150: I think it'd be easier to just clone this one, we're not creating lots of `Session`s.

src/session.rs
```diff
@@ -217,40 +263,70 @@ impl Session {
...
+                if let Some(path) = &cookie.path {
+                    cookie_builder = cookie_builder.path(path.clone());
+                }
+                if let Some(secure) = cookie.secure {
+                    cookie_builder = cookie_builder.secure(secure);
+                }
+
+                let mut cookie_url = self.url.clone();
+                if let Some(domain) = &cookie.domain {
+                    cookie_url = format!("http://{domain}").parse()?;
```
blyxxyz, line 286: [`url::Host::parse`](https://docs.rs/url/latest/url/enum.Host.html#method.parse) looks a little easier

src/main.rs
```diff
@@ -330,15 +332,11 @@ fn run(args: Cli) -> Result<i32> {
         s.save_headers(&headers)?;
 
         let mut cookie_jar = cookie_jar.lock().unwrap();
-        for cookie in s.cookies() {
-            match cookie_jar.insert_raw(&cookie, &url) {
-                Ok(..) | Err(cookie_store::CookieError::Expired) => {}
-                Err(err) => return Err(err.into()),
-            }
-        }
+        *cookie_jar = CookieStore::from_cookies(s.cookies(), false)?;
```
blyxxyz, line 335: I'm unsure if we should abort on invalid cookies. It's probably fine though. I want to compare this to HTTPie's behavior when I do a functional test.

It's weird that `CookieStore::from_cookies()` takes an iterator of `Result`s.

blyxxyz (commented):
Haven't done a functional test yet but the code looks good!

src/session.rs
ducaale: That would return a `Host` but I need to construct a `cookie_url` that matches the cookie's domain. This is to avoid any `DomainMismatch` errors later when calling `try_from_raw_cookie()`.

src/main.rs
ducaale: I generally find the `cookie_store` crate hard to work with. I think it would benefit from a breaking change release that revisits its current API.

> I want to compare this to HTTPie's behavior when I do a functional test.

Sounds good. I have compared the unhappy path between HTTPie and xh yet.

blyxxyz (commented):
For `https://www.youtube.com`, HTTPie stores the cookies under the domain `.youtube.com` rather than `youtube.com`. (Cookies for `https://www.wikipedia.org` do end up under `wikipedia.org`.)

I haven't read the spec but I assume these have different meanings that we need to keep?

(Reusing an xh session file in HTTPie causes one of the three YouTube cookies to be stored twice, once with the dot and once without. Then HTTPie will also send it twice, but xh only once.)

ducaale:
The cookie crate is where the leading dot is removed from the domain. See https://github.com/SergioBenitez/cookie-rs/issues/73

I was initially tempted to declare it as an HTTPie bug, but the behaviour is consistent with Requests' cookie jar

```
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests
>>> import http.cookies
>>>
>>> cookie_jar = requests.cookies.RequestsCookieJar()
>>> cookie_jar.update(http.cookies.SimpleCookie('hello=world; Path=/; HttpOnly; Secure; Domain=example.com;'))
>>> cookie_jar.update(http.cookies.SimpleCookie('hello=world; Path=/; HttpOnly; Secure; Domain=.example.com;'))
>>> cookie_jar
<RequestsCookieJar[Cookie(version=0, name='hello', value='world', port=None, port_specified=False, domain='.example.com', domain_specified=True, domain_initial_dot=True, path='/', path_specified=True, secure=True, expires=None, discard=False, comment='', comment_url=False, rest={'HttpOnly': True}, rfc2109=False), Cookie(version=0, name='hello', value='world', port=None, port_specified=False, domain='example.com', domain_specified=True, domain_initial_dot=False, path='/', path_specified=True, secure=True, expires=None, discard=False, comment='', comment_url=False, rest={'HttpOnly': True}, rfc2109=False)]>
```

blyxxyz:
Hmmm. Both Firefox and Chromium seem to remember (in the dev tools and in the cookies sqlite db) whether there was a leading dot, even if they presumably ignore it in the end.

Is there an easy way for us to remember it too?

ducaale:
I can't think of a way that doesn't involve forking cookie-rs and cookie-store.

A somewhat random question; do you go with rebase or merge when bringing changes back from master to a PR?

blyxxyz:
Yeah, I don't see a way either. Let's leave this as it is.

I think it'd be good to document this divergence from HTTPie somewhere, though it's too obscure to go in the README. ("Formatted output is always UTF-8" also doesn't really belong there I think, it needs a longer explanation to clarify what it even means.) Maybe a separate markdown file?

> do you go with git-rebase or git-merge when bringing changes back from master to a feature branch?

I usually don't rebase if I already made a PR or if I touched code/logic that changed in master. Otherwise things can get messy.

src/main.rs
blyxxyz: An unparsable `"domain"` crashes xh but not HTTPie. That seems acceptable, though we could maybe turn it into a warning and a skip.

This is what the crash looks like though:
```
xh: error: empty host
```
We probably need a `.context()` call somewhere.

ducaale:
>I think it'd be good to document this divergence from HTTPie somewhere, though it's too obscure to go in the README. ("Formatted output is always UTF-8" also doesn't really belong there I think, it needs a longer explanation to clarify what it even means.) Maybe a separate markdown file?

Shall we track it as an issue for now? I have now documented couple of existing differences in https://github.com/ducaale/xh/issues/322

src/main.rs
ducaale: I somehow missed your last comment about unparsable "domain" errors 😬. 

>...though we could maybe turn it into a warning and a skip.

My initial approach was to filter out any `err` values from `s.cookies()` but that caused the cookie to be discarded from the session file. In the end, I went with just adding context to cookie errors.

```
xh: error: Failed to load the cookies from the session file

Caused by:
    empty host
```

blyxxyz: Looks good! I'll try to give this PR a final look tomorrow.

blyxxyz (approved):

PR #318, opened 2023-07-17
add format options: json indentation

### LINKED ISSUE #316: Make JSON indentation level configurable (e.g. 2 or 4 spaces)
Currently 4 spaces are hardcoded. But I prefer 2 spaces, as it's better for deeply nested JSON and allows it to better fit into terminal width.

Bnyro:
closes #316

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -9,7 +9,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::time::Duration;
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Context};
 use clap::{self, AppSettings, ArgEnum, Error, ErrorKind, FromArgMatches, Result};
 use encoding_rs::Encoding;
 use once_cell::sync::OnceCell;
@@ -82,6 +82,15 @@ Controls output processing. Possible values are:
 Defaults to \"format\" if the NO_COLOR env is set and to \"none\" if stdout is not tty."
     )]
     pub pretty: Option<Pretty>,
+
+    /// Set output formatting options. The only currently implemented option
+    /// is json.indent.
+    ///
+    /// Different options need to be separated by a comma.
+    ///
+    /// Example: --format-options=json.indent:2
+    #[clap(long, value_name = "FORMAT_OPTIONS")]
+    pub format_options: Option<FormatOptions>,
 
     /// Output coloring style.
     #[clap(short = 's', long, arg_enum, value_name = "THEME")]
@@ -946,6 +955,40 @@ impl Pretty {
 
     pub fn format(self) -> bool {
         matches!(self, Pretty::Format | Pretty::All)
+    }
+}
+
+#[derive(Debug)]
+pub struct FormatOptions {
+    pub json_indent: usize,
+}
+
+impl Default for FormatOptions {
+    fn default() -> Self {
+        Self { json_indent: 4 }
+    }
+}
+
+impl FromStr for FormatOptions {
+    type Err = anyhow::Error;
+    fn from_str(options: &str) -> anyhow::Result<FormatOptions> {
+        let mut format_options = FormatOptions::default();
+
+        // generate a map of the specified options
+        for argument in options.split(',') {
+            let (key, value) = argument
+                .split_once(':')
+                .context("Format options consist of a key and a value, separated by a \":\".")?;
+            match key {
+                "json.indent" => {
+                    format_options.json_indent = value.parse().context("Invalid indent value")?;
+                }
+                _ => {
+                    return Err(anyhow!("Unknown option '{key}'"));
+                }
+            }
+        }
+        Ok(format_options)
     }
 }
 
--- a/src/formatting.rs
+++ b/src/formatting.rs
@@ -7,11 +7,14 @@ use syntect::parsing::SyntaxSet;
 use syntect::util::LinesWithEndings;
 use termcolor::WriteColor;
 
-use crate::{buffer::Buffer, cli::Theme};
+use crate::{
+    buffer::Buffer,
+    cli::{FormatOptions, Theme},
+};
 
-pub fn get_json_formatter() -> jsonxf::Formatter {
+pub fn get_json_formatter(format_options: &FormatOptions) -> jsonxf::Formatter {
     let mut fmt = jsonxf::Formatter::pretty_printer();
-    fmt.indent = String::from("    ");
+    fmt.indent = " ".repeat(format_options.json_indent);
     fmt.record_separator = String::from("\n\n");
     fmt.eager_record_separators = true;
     fmt
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -13,6 +13,7 @@ use reqwest::header::{
 use reqwest::Version;
 use url::Url;
 
+use crate::cli::FormatOptions;
 use crate::decoder::{decompress, get_compression_type};
 use crate::{
     buffer::Buffer,
@@ -114,10 +115,17 @@ pub struct Printer {
     sort_headers: bool,
     stream: bool,
     buffer: Buffer,
+    format_options: FormatOptions,
 }
 
 impl Printer {
-    pub fn new(pretty: Pretty, theme: Option<Theme>, stream: bool, buffer: Buffer) -> Self {
+    pub fn new(
+        pretty: Pretty,
+        theme: Option<Theme>,
+        stream: bool,
+        buffer: Buffer,
+        format_options: FormatOptions,
+    ) -> Self {
         let theme = theme.unwrap_or(Theme::Auto);
 
         Printer {
@@ -127,6 +135,7 @@ impl Printer {
             stream,
             theme,
             buffer,
+            format_options,
         }
     }
 
@@ -159,15 +168,16 @@ impl Printer {
             return self.print_syntax_text(text, "json");
         }
 
+        let mut formatter = get_json_formatter(&self.format_options);
         if self.color {
             let mut buf = Vec::new();
-            get_json_formatter().format_buf(text.as_bytes(), &mut buf)?;
+            formatter.format_buf(text.as_bytes(), &mut buf)?;
             // in principle, buf should already be valid UTF-8,
             // because JSONXF doesn't mangle it
             let text = String::from_utf8_lossy(&buf);
             self.print_colorized_text(&text, "json")
         } else {
-            get_json_formatter().format_buf(text.as_bytes(), &mut self.buffer)
+            formatter.format_buf(text.as_bytes(), &mut self.buffer)
         }
     }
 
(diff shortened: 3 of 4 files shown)


src/printer.rs
```diff
@@ -107,17 +110,57 @@ impl<'a, T: Read> BinaryGuard<'a, T> {
     }
 }
 
+pub struct FormatOptions {
```
ducaale, line 113: All of our CLI-related data types live in `src/cli.rs`. Can you move this struct to that file as well?

src/printer.rs
```diff
@@ -107,17 +110,57 @@ impl<'a, T: Read> BinaryGuard<'a, T> {
...
+    pub json_indent: usize,
+    pub json_record_separator: String
+}
+
+impl FormatOptions {
+    pub fn from_string(options: Option<String>) -> FormatOptions {
```
ducaale, line 119: We usually implement `FromStr` trait for data types so the error can be reported by clap. See `Print` in `src/cli.rs` as an example.

src/cli.rs
```diff
@@ -83,6 +83,13 @@ Defaults to \"format\" if the NO_COLOR env is set and to \"none\" if stdout is n
     )]
     pub pretty: Option<Pretty>,
 
+    // Set formatting options, such as the amount of spaces per JSON indentation
+    // 
+    // This only works when pretty is set to all or format
+    // Example: --format-options=json.indent:2,json.record.separator:' EOF'
+    #[clap(long, value_name = "FORMAT_OPTIONS")]
+    pub format_options: Option<String>,
```
ducaale, line 91: Once you implement `FromStr` trait for `FormatOptions`, you can pass it to clap

```suggestion
    pub format_options: Option<FormatOptions>,
```

src/formatting.rs
```diff
@@ -7,12 +7,12 @@ use syntect::parsing::SyntaxSet;
...
-use crate::{buffer::Buffer, cli::Theme};
+use crate::{buffer::Buffer, cli::Theme, printer::FormatOptions};
 
-pub fn get_json_formatter() -> jsonxf::Formatter {
+pub fn get_json_formatter(format_options: &FormatOptions) -> jsonxf::Formatter {
     let mut fmt = jsonxf::Formatter::pretty_printer();
-    fmt.indent = String::from("    ");
-    fmt.record_separator = String::from("\n\n");
+    fmt.indent = String::from(" ".repeat(format_options.json_indent));
+    fmt.record_separator = format_options.json_record_separator.clone();
```
ducaale, line 15: I don't think HTTPie allows customising the record separator. Can you open an issue before implementing this change?

src/printer.rs
```diff
@@ -107,17 +110,57 @@ impl<'a, T: Read> BinaryGuard<'a, T> {
...
+                format_options.json_record_separator = record_separator.to_string();
+            }
+        }
+        return format_options
+    }
+}
+
+impl Default for FormatOptions {
+    fn default() -> Self {
+        Self { json_indent: DEFAULT_JSON_INDENT, json_record_separator: String::from(DEFAULT_JSON_RECORD_SEPARATOR) }
```
ducaale, line 148: Since these values are not used anywhere, I would prefer to inline them within this function.

src/printer.rs
```diff
@@ -107,17 +110,57 @@ impl<'a, T: Read> BinaryGuard<'a, T> {
...
+    pub json_record_separator: String
+}
+
+impl FormatOptions {
+    pub fn from_string(options: Option<String>) -> FormatOptions {
+        let mut format_options = FormatOptions::default();
+
+        // generate a map of the specified options
+        if let Some(options) = options {
+            let mut arguments = HashMap::new();
```
ducaale, line 124: Can we avoid allocating this hashmap?

src/formatting.rs
Bnyro: Ah, I see.
Since the goal is to implement all features httpie does, wouldn't it make sense to also add new features now that xh is almost fully compatible with httpie?
I only added the record separator option to be able to test whether the parser works properly though, so I don't mind dropping it.

Bnyro:
Thanks for the review though, I'm way more happy with the solution now than I was before!

src/formatting.rs
ducaale: I feel like this option might complicate defaulting to [streamed responses](https://github.com/ducaale/xh/pull/233#issuecomment-1052445527) in the future.

Let's drop it for now and we can think about adding it in the feature.

src/cli.rs
```diff
@@ -949,6 +957,51 @@ impl Pretty {
...
+
+            // save the specified options to the `FormatOptions` struct
+            if key == "json.indent" {
+                if let Ok(json_indent) = value.parse::<usize>() {
+                    format_options.json_indent = json_indent;
+                }
+            } else {
+                return Err(anyhow!("Format option \"{}\" not supported", key));
+            }
+        }
```
ducaale, line 999: We can make this more strict by disallowing options with multiple colons e.g `json.indent:8:8` or invalid indent number e.g `json.indent:-2`
 
```suggestion
        for argument in options.split(',') {
            let argument_parts: Vec<&str> = argument.splitn(2, ':').collect();
            match argument_parts[..] {
                ["json.indent", value] => {
                    format_options.json_indent =
                        value.parse().map_err(|_| anyhow!("Invalid indent value"))?;
                }
                _ => {
                    return Err(anyhow!(
                        "Format options consist of a key and a value, separated by a \":\"."
                    ));
                }
            }
        }
```

ducaale (commented):
Only one more thing left to fix. Other than that, this looks great!

ducaale:
Can you also run `cargo fmt` to fix the CI error? Thanks

src/cli.rs
Bnyro: Really neat, I wasn't aware of that handy match expression using `[..]`

src/cli.rs
```diff
@@ -949,6 +957,41 @@ impl Pretty {
...
+        for argument in options.split(',') {
+            let argument_parts: Vec<&str> = argument.splitn(2, ':').collect();
+            match argument_parts[..] {
+                ["json.indent", value] => {
+                    format_options.json_indent =
+                        value.parse().map_err(|_| anyhow!("Invalid indent value"))?;
+                }
+                _ => {
+                    return Err(anyhow!(
+                        "Format options consist of a key and a value, separated by a \":\"."
```
blyxxyz, line 986: I think we should have different error messages for a missing `:` and an unknown key.

And since `json.sort_keys` exists in HTTPie maybe it can get a special message explaining that it's unimplemented?

src/cli.rs
```diff
@@ -83,6 +83,14 @@ Defaults to \"format\" if the NO_COLOR env is set and to \"none\" if stdout is n
     )]
     pub pretty: Option<Pretty>,
 
+    // Set formatting options, such as the amount of spaces per JSON indentation
+    //
+    // Different options need to be separated by a comma
+    // This only works when pretty is set to all or format
+    // Example: --format-options=json.indent:2
```
blyxxyz, line 90:
```suggestion
    /// Set output formatting options. The only currently implemented option
    /// is json.indent.
    ///
    /// Different options need to be separated by a comma.
    ///
    /// Example: --format-options=json.indent:2
```

This has to be `///`, not `//`, or it won't show up in the help text. And it looks like HTTPie does have an option that applies regardless of `--pretty` (`headers.sort`), so maybe we shouldn't mention that.

src/cli.rs
blyxxyz: Ah, it looks like HTTPie has more options besides those:
- `headers.sort`
- `json.format`
- `json.sort_keys`
- `xml.format`
- `xml.indent`

It would be nice to hardcode that list for a different error message but if you want to skip that then that's fine by me.

blyxxyz (commented):
I have a few comments but it works great, thank you!

src/cli.rs
```diff
@@ -949,6 +957,41 @@ impl Pretty {
...
+    fn from_str(options: &str) -> anyhow::Result<FormatOptions> {
+        let mut format_options = FormatOptions::default();
+
+        // generate a map of the specified options
+        for argument in options.split(',') {
+            let argument_parts: Vec<&str> = argument.splitn(2, ':').collect();
+            match argument_parts[..] {
+                ["json.indent", value] => {
+                    format_options.json_indent =
+                        value.parse().map_err(|_| anyhow!("Invalid indent value"))?;
```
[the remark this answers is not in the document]
ducaale: Since the clap will report this error, it wouldn't make a difference (I think):

```
error: Invalid value "json.indent:-5" for '--format-options <FORMAT_OPTIONS>': Invalid indent value

For more information try --help
``` 

But from perspective of code aesthetic, `.context()` looks much better than `.map_error()` 👍

src/cli.rs
ducaale: I plan to look at the remaining options in a follow-up PR so we can leave the error as it is now.

src/cli.rs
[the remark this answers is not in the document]
blyxxyz: Oh yeah, looks like it'll print it the same!

The native error message is a little too technical anyway, if you pass it a negative number it says "invalid digit found in string". So maybe this is for the best.

src/cli.rs
```diff
@@ -949,6 +957,41 @@ impl Pretty {
...
+}
+
+impl FromStr for FormatOptions {
+    type Err = anyhow::Error;
+    fn from_str(options: &str) -> anyhow::Result<FormatOptions> {
+        let mut format_options = FormatOptions::default();
+
+        // generate a map of the specified options
+        for argument in options.split(',') {
+            let argument_parts: Vec<&str> = argument.splitn(2, ':').collect();
```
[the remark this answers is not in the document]
Bnyro: `split_once` returns an `Option<(&str, &str)>`, thus we'd need to unwrap it first and do some extra error handling. Since we already handle errors in the match statement, I think that'd just be redundant code here.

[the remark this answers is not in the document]
blyxxyz: I think with anyhow you can write:
```rust
let (key, value) = argument.split_once(':')
    .context("Format options consist of a key and a value, separated by a \":\".")?;
```
(And then it's easier to have a different message for an unknown key.)

But either is fine as long as it works.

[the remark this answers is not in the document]
ducaale: @Bnyro Something like this should work:

```rs
match argument.split_once(':') {
    Some(("json.indent", value)) => {
        format_options.json_indent = value.parse().context("Invalid indent value")?;
    }
    _ => {
        return Err(anyhow!(
            "Format options consist of a key and a value, separated by a \":\"."
        ));
    }
}
```

src/cli.rs
```diff
@@ -10,6 +10,7 @@ use std::str::FromStr;
 use std::time::Duration;
 
 use anyhow::anyhow;
+use anyhow::Context;
```
ducaale, line 13:
```suggestion
use anyhow::{anyhow, Context};
```

src/cli.rs
```diff
@@ -949,6 +958,42 @@ impl Pretty {
...
+                .split_once(':')
+                .context("Format options consist of a key and a value, separated by a \":\".")?;
+            match key {
+                "json.indent" => {
+                    format_options.json_indent = value.parse().context("Invalid indent value")?;
+                }
+                _ => {
+                    return Err(anyhow!(
+                        "Format options consist of a key and a value, separated by a \":\"."
+                    ));
```
blyxxyz, line 989:
```suggestion
                    return Err(anyhow!("Unknown option '{key}'"));
```

blyxxyz (approved):

ducaale (approved):
Thanks!

PR #319, opened 2023-07-30
Add format options: json.format and headers.sort

ducaale:
This is a follow-up PR to https://github.com/ducaale/xh/pull/318. 

Also, see https://httpie.io/docs/cli/format-options for all the format options supported by HTTPie.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -923,18 +923,6 @@ pub enum AuthType {
     Digest,
 }
 
-#[derive(ArgEnum, Debug, PartialEq, Eq, Clone, Copy)]
-pub enum Pretty {
-    /// (default) Enable both coloring and formatting
-    All,
-    /// Apply syntax highlighting to output
-    Colors,
-    /// Pretty-print json and sort headers
-    Format,
-    /// Disable both coloring and formatting
-    None,
-}
-
 /// The caller must check in advance if the string is valid. (clap does this.)
 fn parse_tls_version(text: &str) -> Option<tls::Version> {
     match text {
@@ -946,6 +934,18 @@ fn parse_tls_version(text: &str) -> Option<tls::Version> {
         "tls1.3" => Some(tls::Version::TLS_1_3),
         _ => unreachable!(),
     }
+}
+
+#[derive(ArgEnum, Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Pretty {
+    /// (default) Enable both coloring and formatting
+    All,
+    /// Apply syntax highlighting to output
+    Colors,
+    /// Pretty-print json and sort headers
+    Format,
+    /// Disable both coloring and formatting
+    None,
 }
 
 impl Pretty {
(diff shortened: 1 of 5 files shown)


ducaale:
I need to update the docs + add support for passing `--format-options` multiple times. From HTTPie help:

```
You may use this option multiple times, as well as specify multiple comma-separated options at the same time
```

src/cli.rs
```diff
@@ -924,14 +931,25 @@ impl Pretty {
...
+    pub fn merge(mut self, other: &Self) -> Self {
+        if let Some(json_indent) = other.json_indent {
+            self.json_indent = Some(json_indent);
+        }
+        if let Some(json_format) = other.json_format {
+            self.json_format = Some(json_format);
+        }
+        if let Some(headers_sort) = other.headers_sort {
+            self.headers_sort = Some(headers_sort);
+        }
```
blyxxyz, line 951:
```suggestion
        self.json_indent = other.json_indent.or(self.json_indent);
        self.json_format = other.json_format.or(self.json_format);
        self.headers_sort = other.headers_sort.or(self.headers_sort);
```

src/cli.rs
```diff
@@ -1645,4 +1675,48 @@ mod tests {
...
+        let invalid_format_options = vec![
+            // malformed strings
+            ":8",
+            "json.indent:",
+            ":",
+            "",
+            "json.format:true, json.indent:4",
+            // invalid values
+            "json.indent:-8",
+            "json.format:False",
```
blyxxyz, line 1690: HTTPie actually supports this case. It seems to be case-insensitive in general, even for the keys:

```py
            path, value = option.lower().split(':')
```
https://github.com/httpie/cli/blob/e52a60e67ca15291c9ab18f395d4d7af3a6204ff/httpie/cli/argtypes.py#L220

(so `JSON.format:TRUE` is OK.)

tests/cli.rs
blyxxyz, line 1: Can you add a test for merged format options? That way we'll notice if something goes wrong at the clap layer.

src/printer.rs
```diff
@@ -124,18 +124,21 @@ impl Printer {
         theme: Option<Theme>,
         stream: bool,
         buffer: Buffer,
-        format_options: FormatOptions,
+        format_options: Vec<FormatOptions>,
     ) -> Self {
         let theme = theme.unwrap_or(Theme::Auto);
+        let format_options = format_options
+            .iter()
+            .fold(FormatOptions::default(), FormatOptions::merge);
```
blyxxyz, line 132: I think I'd rather do this in `main.rs`, as close to `cli.rs` as we can easily manage. It's not really the printer's responsibility.

blyxxyz (commented):
Works great! Just a few code nitpicks and one compatibility problem.

blyxxyz (approved):
Nice cleanup for `Theme`!

PR #321, opened 2023-08-13
 Update clap to v4.0

ducaale:


src/cli.rs
```diff
@@ -609,7 +588,7 @@ impl Cli {
         static ARG_STORAGE: OnceCell<Vec<String>> = OnceCell::new();
```
blyxxyz, line 588: I think we can get rid of this whole kludge now and just pass `String`s into `Arg::new()`

src/cli.rs
```diff
@@ -631,7 +593,7 @@ impl Cli {
             .collect();
 
         app.args(negations)
-            .after_help("Each option can be reset with a --no-OPTION argument.")
+            .after_help(format!("Each option can be reset with a --no-OPTION argument.\n\nRun \"{} help\" for more complete documentation.", env!("CARGO_PKG_NAME")))
```
blyxxyz, line 596: We used to show the `xh help` suggestion only if `--help` were used. Is that possible?

blyxxyz (approved):
Works great! I love the cleanup.

blyxxyz:
Hm, now `xh help` ends with a blank line and `xh --help` doesn't. I can live with that though.

PR #325, opened 2023-09-16
Disable cURL's URL globbing

### LINKED ISSUE #324: xh doesn't translate `[` and `]` to curl correctly.
`xh --curl https://example.com/[something]`

gives

`curl 'https://example.com/[something]'`

which doesn't work.

`[` and `]` will need to be percent encoded to `%5B` and `%5D` respectively or they need to be escaped.

This works - `curl 'https://example.com/%5Bsomething%5D'`

As does this - `curl 'https://example.com/\[something\]'`

Could you fix this ?

Thanks!

ducaale:
Resolves #324

### DIFF
--- a/src/to_curl.rs
+++ b/src/to_curl.rs
@@ -270,6 +270,11 @@ pub fn translate(args: Cli) -> Result<Command> {
     }
 
     let url = url_with_query(args.url, &args.request_items.query()?);
+
+    if url.as_str().contains(['[', ']', '{', '}']) {
+        cmd.opt("-g", "--globoff")
+    }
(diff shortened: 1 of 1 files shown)


blyxxyz (approved):
Looks good!

I notice we already URL-encode `{}` if it's part of the path, e.g. in `https://example.com/{80,90}`. But that's good. `curl -g 'localhost:8080/{80,90}'` sends `/{80,90}` whereas `xh ':8080/{80,90}'` sends `/%7B80,90%7D`, so we still need it to URL-encode them.

PR #327, opened 2023-10-01
Add `--resolve` flag for overriding DNS resolution

### LINKED ISSUE #317: Alternative for cURL --resolve flag
Is there an alternative for the `--resolve` flag of cURL? Setting the `Host` header alone does not work when the target server is using SNI. In `httpie` there is a feature request for this, but is pending on an upstream PR.

https://github.com/httpie/httpie/issues/99

ducaale:
Resolves #317


src/cli.rs
```diff
@@ -1175,6 +1184,31 @@ impl FromStr for Proxy {
...
+    pub addr: IpAddr,
+}
+
+impl FromStr for Resolve {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        let (domain, addr) = s
+            .split_once(':')
+            .context("The value passed to --resolve should be formatted as <HOST>:<ADDRESS>")?;
```
blyxxyz, line 1199:
```suggestion
            .context("Value should be formatted as <HOST>:<ADDRESS>")?;
```
From
`error: invalid value 'example.com1.1.1.1' for '--resolve <HOST:ADDRESS>': The value passed to --resolve should be formatted as <HOST>:<ADDRESS>`
To
`error: invalid value 'example.com1.1.1.1' for '--resolve <HOST:ADDRESS>': Value should be formatted as <HOST>:<ADDRESS>`

src/cli.rs
```diff
@@ -1175,6 +1184,31 @@ impl FromStr for Proxy {
...
+pub struct Resolve {
+    pub domain: String,
+    pub addr: IpAddr,
+}
+
+impl FromStr for Resolve {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        let (domain, addr) = s
```
blyxxyz, line 1197:
```suggestion
    fn from_str(s: &str) -> anyhow::Result<Self> {
        if s.chars().filter(|&c| c == ':').count() == 2 {
            // More than two colons could mean an IPv6 address.
            // Exactly two colons probably means the user added a port, curl-style.
            return Err(anyhow!(
                "Value should be formatted as <HOST>:<ADDRESS> (not <HOST>:<PORT>:<ADDRESS>)"
            ));
        }

        let (domain, addr) = s
```

blyxxyz:
curl additionally supports enclosing the IP address in brackets, e.g. `--resolve example.org:80:[::1]`. This is by analogy with IPv6 addresses in URLs (`http://[::1]`). It happens to work with IPv4 addresses as well.

`--resolve example.org:::1` does look confusing so maybe we should support it too. I'd prefer to limit it to IPv6 addresses though.

src/to_curl.rs
```diff
@@ -289,6 +289,21 @@ pub fn translate(args: Cli) -> Result<Command> {
         cmd.arg(interface);
     };
 
+    if !args.resolve.is_empty() {
+        let port = match url.port() {
+            Some(port) => port,
+            None if url.scheme() == "http" => 80,
+            None if url.scheme() == "https" => 443,
+            None => return Err(anyhow!("Unsupported URL scheme: '{}'", url.scheme())),
+        };
```
blyxxyz, line 298:
```suggestion
        let port = url
            .port_or_known_default()
            .with_context(|| format!("Unsupported URL scheme: '{}'", url.scheme()))?;
```
(plus an `anyhow::Context` import)

blyxxyz (approved):
Works great!

Issue #330, opened 2023-10-21
Latest release fails to build on NetBSD


0323pin:
Hi,

Trying to update our package from 0.18.0 to 0.19.2 fails with:

```
error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> /usr/pkgsrc/wip/xh/work/vendor/network-interface-1.0.0/src/target/getifaddrs.rs:25:18
   |
25 |         unsafe { libc::freeifaddrs(self.base) }
   |                  ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> /usr/pkgsrc/wip/xh/work/vendor/network-interface-1.0.0/src/target/getifaddrs.rs:31:20
   |
31 |     match unsafe { libc::getifaddrs(addr.as_mut_ptr()) } {
   |                    ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
 --> /usr/pkgsrc/wip/xh/work/vendor/network-interface-1.0.0/src/target/getifaddrs.rs:5:16
  |
5 |     base: *mut libc::ifaddrs,
  |                ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
 --> /usr/pkgsrc/wip/xh/work/vendor/network-interface-1.0.0/src/target/getifaddrs.rs:6:16
  |
6 |     next: *mut libc::ifaddrs,
  |                ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> /usr/pkgsrc/wip/xh/work/vendor/network-interface-1.0.0/src/target/getifaddrs.rs:10:17
   |
10 |     type Item = libc::ifaddrs;
   |                 ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> /usr/pkgsrc/wip/xh/work/vendor/network-interface-1.0.0/src/target/getifaddrs.rs:30:44
   |
30 |     let mut addr = mem::MaybeUninit::<*mut libc::ifaddrs>::uninit();
   |                                            ^^^^ use of undeclared crate or module `libc`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `network-interface` (lib) due to 6 previous errors
warning: build failed, waiting for other jobs to finish...
*** Error code 101

[...]

PR #332, opened 2023-10-21
Make network-interface an optional dependency

blyxxyz:
See https://github.com/ducaale/xh/issues/330

There are packages that build with `--no-default-features` but should probably enable this feature, like [Alpine's](https://git.alpinelinux.org/aports/plain/community/xh/APKBUILD?h=3.17-stable). Not sure what the best way is to communicate these things. Maybe a note in the changelog is enough.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1029,9 +1029,9 @@ dependencies = [
 
 [[package]]
 name = "network-interface"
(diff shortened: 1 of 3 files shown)


ducaale (approved):
LGTM 💯

Issue #298, opened 2022-12-15
Update dependencies


sorairolake:
The dependencies should be as the latest as possible (considering the MSRV).

ducaale:
Note that we can raise the MSRV to 1.56 once the next reqwest version with HTTPS IP address support is released (see recent activity in webpki fork by rustls https://github.com/rustls/webpki).

ducaale:
**Update**: MSRV has been raised to 1.64 as a result of upgrading reqwest to v0.11.18

PR #335, opened 2023-10-22
Bump dependencies

jayvdb:
Related to https://github.com/ducaale/xh/issues/298


assets/xhs
jayvdb, line 1: these broken symlinks break the build.rs in `syntect::highlighting::ThemeSet::load_from_folder`

blyxxyz: I think it would be best to move the themes into their own subdirectory.

And the symlinks too, we only use them for `cargo-deb` but this is not obvious right now.

src/formatting.rs
```diff
@@ -56,7 +56,12 @@ impl<'a> Highlighter<'a> {
     /// Write a single piece of highlighted text.
     pub fn highlight(&mut self, text: &str) -> io::Result<()> {
         for line in LinesWithEndings::from(text) {
-            for (style, component) in self.highlighter.highlight(line, self.syntax_set) {
+            // The following unwrap can only be removed with a breaking change to this functions return type.
+            for (style, component) in self
+                .highlighter
+                .highlight_line(line, self.syntax_set)
+                .unwrap()
+            {
```
blyxxyz, line 64:
```suggestion
            for (style, component) in self
                .highlighter
                .highlight_line(line, self.syntax_set)
                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?
            {
```
In the long term we should probably switch to anyhow for these but that has to be propagated in a lot of places.

blyxxyz (commented):
Thank you!

src/formatting.rs
jayvdb: Umm, it isnt an `io::Error`.

Anyway, feel free to add your suggestion if you feel it is appropriate.

assets/xhs
jayvdb: IMO the symlinks should be added as part of that process, when the symlinks are valid, not store them in the repo.

src/formatting.rs
blyxxyz: Questionable things get folded into `io::Error` even in the stdlib. Usually as [`ErrorKind::InvalidData`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.InvalidData).

It's not ideal but I'd rather have a controlled crash than a panic.

assets/xhs
blyxxyz: That would be better, yeah. It's annoying to have broken symlinks in the repo.

But if we remove the symlinks now with no replacement then that breaks the build.

ducaale:
@jayvdb do you mind moving syntax files to their own folder? e.g

```
assets/
├── syntax
│   ├── basic
│   └── large
└── themes
    ├── ansi.tmTheme
    ├── fruity.tmTheme
    ├── monokai.tmTheme
    └── solarized.tmTheme
```

ducaale (approved):

PR #339, opened 2023-11-19
Update Cargo.lock

ducaale:
This PR also raises MSRV to 1.70 and removes ~once_cell~ `atty` from dependencies.


blyxxyz:
Perhaps we can also move from `regex` to [`regex-lite`](https://crates.io/crates/regex-lite)? Saves 250 KB, all the tests still pass. (It has a MSRV of 1.65 though.)

All of our regexes seem to run at most once per process on a small haystack, so it shouldn't have any performance impact. (We probably don't even get any performance benefit from that `regex!()` macro outside of tests.)

blyxxyz:
Can't we remove `once_cell` as well?

ducaale:
Yeah, I wanted to do that, but then I noticed that `once_cell` is a transitive dependency of both `reqwest` and `syntext`.

blyxxyz:
Ah, I see. Probably a matter of time until they get rid of it but there's no hurry then.

blyxxyz (approved):

PR #344, opened 2024-01-06
Fix broken HTTPie request item link in README.md

jihchi:
## [Rendered markdown](https://github.com/jihchi/ht/blob/patch-1/README.md)


ducaale (approved):
Thanks!

Issue #352, opened 2024-01-26
macos: `xh` config directory not in xdg dirs


skshetry:
The manpage documents `~/.config/xh` as the default config directory for macOS and Linux. 

https://github.com/ducaale/xh/blob/7c7b1e0f0d3b07978c6769d7930a0d36bb96637b/doc/xh.1#L347-L349

https://github.com/ducaale/xh/blob/7c7b1e0f0d3b07978c6769d7930a0d36bb96637b/doc/xh.1#L377-L379

https://github.com/ducaale/xh/blob/7c7b1e0f0d3b07978c6769d7930a0d36bb96637b/doc/xh.1#L389-L390 

But `xh` uses the `/Users/Alice/Library/Application Support/xh` location on macOS, which comes from  [`dirs::config_dir()`](https://docs.rs/dirs/latest/dirs/fn.config_dir.html), which uses the above-mentioned location.

This behavior differs from `httpie`, which uses xdg dirs, eg: `~/.config/httpie` on non-Windows platforms. 

See https://httpie.io/docs/cli/config.

PR #353, opened 2024-01-28
Default `XH_CONFIG_DIR` to `~/.config/xh` in macOS

ducaale:
This is a potentially disruptive change but aligns us better with our docs and HTTPie.

Addresses #352

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -73,7 +73,7 @@ pub fn random_string() -> String {
 pub fn config_dir() -> Option<PathBuf> {
     if let Some(dir) = std::env::var_os("XH_CONFIG_DIR") {
         Some(dir.into())
-    } else if cfg!(macos) {
+    } else if cfg!(target_os = "macos") {
         dirs::home_dir().map(|dir| dir.join(".config").join("xh"))
     } else {
         dirs::config_dir().map(|dir| dir.join("xh"))
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -77,7 +77,11 @@ pub fn config_dir() -> Option<PathBuf> {
 
     if cfg!(target_os = "macos") {
         let legacy_config_dir = dirs::config_dir()?.join("xh");
-        let new_config_dir = dirs::home_dir()?.join(".config").join("xh");
+        let config_home = match var_os("XDG_CONFIG_HOME") {
+            Some(dir) => dir.into(),
+            None => dirs::home_dir()?.join(".config"),
+        };
+        let new_config_dir = config_home.join("xh");
         if legacy_config_dir.exists() && !new_config_dir.exists() {
             Some(legacy_config_dir)
         } else {
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -72,11 +72,19 @@ pub fn random_string() -> String {
 
 pub fn config_dir() -> Option<PathBuf> {
     if let Some(dir) = std::env::var_os("XH_CONFIG_DIR") {
-        Some(dir.into())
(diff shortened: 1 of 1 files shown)


blyxxyz:
Maybe we can use the old location if it exists and the new one doesn't?

I'm also curious what other tools use in this case, I'll try to do a survey later.

blyxxyz:
I checked a few programs in my (Linux) `~/.config`:

- Fish [uses `~/.config` and listens to XDG variables](https://github.com/fish-shell/fish-shell/blob/a03162bd5b454cfbe7e4606f467845ffcae485a2/src/path.rs#L712). And IIRC the maintainer works at Apple.
- htop [follows XDG](https://github.com/htop-dev/htop/blob/32cb30273b9785ba2dd2cada5d0a677f97db7ec0/Settings.c#L802) on all platforms.
- psysh [follows XDG](https://github.com/bobthecow/psysh/blob/7d7f26d3dbba0a28a59974782d49b98f3280d441/src/ConfigPaths.php#L91) everywhere besides Windows.
- mpv [follows XDG](https://github.com/mpv-player/mpv/blob/c6ccabe5854254320b175cde4106ec40d7a30946/osdep/path-darwin.c#L33) on macOS, with a similar fallback where it uses a legacy dir only if the legacy dir exists and the new one doesn't. It only uses fancy macOS dirs for its cache.
- Syncthing [uses `Application Support/`](https://github.com/syncthing/syncthing/blob/45beb28fa5b051d14a10f102a05ec46d36ecaa75/lib/locations/locations.go#L209). But this is not a CLI tool.

I'm sold, `~/.config` is the way to go for CLI tools.

HTTPie prioritizes [its own legacy `~/.httpie` fallback](https://github.com/httpie/cli/blob/9e8e3691c8d40ca8dcdc90fbd9be12016eacdd56/httpie/config.py#L45-L49) if that exists, without checking the XDG dir, so a bit simpler than my suggestion. But I think the more complicated logic is worth it.

src/utils.rs
```diff
@@ -72,9 +72,19 @@ pub fn random_string() -> String {
 
 pub fn config_dir() -> Option<PathBuf> {
     if let Some(dir) = std::env::var_os("XH_CONFIG_DIR") {
-        Some(dir.into())
+        return Some(dir.into());
+    }
+
+    if cfg!(target_os = "macos") {
+        let legacy_config_dir = dirs::config_dir()?.join("xh");
+        let new_config_dir = dirs::home_dir()?.join(".config").join("xh");
```
blyxxyz, line 80:
```suggestion
        let config_home = match var_os("XDG_CONFIG_HOME") {
            Some(dir) => dir.into(),
            None => dirs::home_dir()?.join(".config"),
        };
        let new_config_dir = config_home.join("xh");
```

src/utils.rs
```diff
@@ -72,9 +72,19 @@ pub fn random_string() -> String {
 
 pub fn config_dir() -> Option<PathBuf> {
     if let Some(dir) = std::env::var_os("XH_CONFIG_DIR") {
-        Some(dir.into())
+        return Some(dir.into());
+    }
+
+    if cfg!(target_os = "macos") {
+        let legacy_config_dir = dirs::config_dir()?.join("xh");
```
blyxxyz, line 79:
```suggestion
        // On macOS dirs returns `~/Library/Application Support`.
        // ~/.config is more usual so we switched to that. But first we check for
        // the legacy location.
        let legacy_config_dir = dirs::config_dir()?.join("xh");
```

blyxxyz (approved):

PR #357, opened 2024-03-22
hyper v1 upgrade

zuisong:


zuisong:
two test cases that cannot pass, I have already submitted an issue to reqwest.
https://github.com/seanmonstar/reqwest/issues/2201

Cargo.toml
```diff
@@ -49,9 +49,11 @@ unicode-width = "0.1.9"
 url = "2.2.2"
 
 [dependencies.reqwest]
-version = "0.11.18"
+git = "https://github.com/seanmonstar/reqwest"
```
zuisong, line 52: Wait for the new version of reqwest to be released, the current version 0.12.1 has 2 failing tests

tests/cli.rs
```diff
@@ -2846,7 +2841,7 @@ fn method_is_changed_when_following_302_redirect() {
             }
             "/second_page" => {
                 assert_eq!(req.method(), "GET");
-                assert!(req.headers().get("Content-Length").is_none());
+                assert_eq!(req.headers().get("Content-Length").unwrap(), "0");
```
zuisong, line 2844: Currently Hyper for empty bodies with Content-Length header unset automatically sets the value 0

maybe we should change this test case

ducaale: Sounds good to me 👍

blyxxyz: Is this server-side or client-side? If it's just some weirdness in the test server I'm not worried but if xh is sending this header then I'd like to look into it

zuisong: it is server-side action, hyper server auto set conent-length to 0.
I started a test server to test this feature It will automatically return conent-length: 0


source code:
https://github.com/hyperium/hyper/blob/0013bdda5cd34ed6fca089eceb0133395b7be041/src/proto/h1/role.rs#L881


demo server,you can verify it by run `curl -v http://localhost:3000`

```rust
// tokio = { version = "1.36.0", features = ["full"] }
// hyper = { version = "1.2.0", features = ["server", "http1"] }
// hyper-util = { version = "0.1.3", features = ["full"] }
// http-body-util = { version = "0.1.1",features = [] }

use std::net::SocketAddr;
use http_body_util::Full;
use hyper::body::Bytes;
use hyper::service::service_fn;
use hyper_util::rt::TokioIo;
use tokio::net::TcpListener;

type Builder = hyper_util::server::conn::auto::Builder<hyper_util::rt::TokioExecutor>;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    let listener = TcpListener::bind(addr).await?;
    loop {
        let (stream, _) = listener.accept().await?;
        let io = TokioIo::new(stream);
        let builder = Builder::new(hyper_util::rt::TokioExecutor::new());
        tokio::task::spawn(async move {
            if let Err(err) = builder
                .clone()
                .serve_connection(
                    io,
                    service_fn( |req| async {
                        hyper::Response::builder()
                            .header("date", "N/A")
                            .header("content-encoding", "gzip")
                            .body::<Full<Bytes>>("".into())
                    }),
                )
                .await
            {
                println!("Error serving connection: {:?}", err);
            }
        });
    }
}

```

tests/server/mod.rs
```diff
@@ -51,95 +54,127 @@ impl Drop for Server {
...
+    http_with_config(func, |_builder| {})
 }
 
-type Serv = dyn Fn(Request<Body>) -> Box<ServFut> + Send + Sync;
-type ServFut = dyn Future<Output = Response<Body>> + Send + Unpin;
+type Builder = hyper_util::server::conn::auto::Builder<hyper_util::rt::TokioExecutor>;
 
-fn http_inner(func: Arc<Serv>) -> Server {
-    //Spawn new runtime in thread to prevent reactor execution context conflict
+pub fn http_with_config<F1, Fut, F2, Bu>(func: F1, apply_config: F2) -> Server
```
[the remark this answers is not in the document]
blyxxyz: Awesome, thank you!

Cargo.toml
zuisong: Reqwest just released a new version 0.12.2, so this PR can be ready to go

blyxxyz:
xh does send `Content-Length` now even for GET requests. That seems weird since RFC 9110 says
> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain content and the method semantics do not anticipate such data.

It seems to be a bug? See https://github.com/seanmonstar/reqwest/issues/2202#issuecomment-2016577201. Notably this is causing real-world issues for someone, so we shouldn't send this header for GET requests if we can avoid it.

Otherwise we'd have to add it to the request headers that we print.
<details>
<summary>Test case</summary>

Run this in one terminal:
```sh
while nc -l -p 8080; do echo; done
```
Then run this in another:
```sh
cargo run -- -v :8080
```
This is what xh claims to send:
```
GET / HTTP/1.1
Accept: */*
Connection: keep-alive
Host: localhost:8080
User-Agent: xh/0.21.0
```
This is what's actually received:
```
GET / HTTP/1.1
User-Agent: xh/0.21.0
Connection: keep-alive
Accept: */*
Host: localhost:8080
Content-Length: 0
```
</details>

zuisong:
Thanks for pointing that out，I've submitted a PR to fix it.
https://github.com/seanmonstar/reqwest/pull/2207

blyxxyz:
This is very odd, but with your patched reqwest xh seems to send a body of `0` by default, i.e. the text `0`. This is what I receive through netcat from `xh post :8080`:
```
POST / HTTP/1.1
User-Agent: xh/0.21.0
Connection: keep-alive
Accept: */*
Host: localhost:8080
Transfer-Encoding: chunked

0

```
(or escaped: `POST / HTTP/1.1\r\nAccept-Encoding: gzip, deflate, br\r\nUser-Agent: xh/0.21.0\r\nConnection: keep-alive\r\nAccept: */*\r\nHost: localhost:8080\r\nTransfer-Encoding: chunked\r\n\r\n0\r\n\r\n`)

It doesn't happen if I comment out the reqwest patch in `Cargo.toml`.

-----

I've also noticed that with hyper v1 xh now errors if you override the Content-Length to be too large, e.g. `xh post :8080 Content-Length:10`. I don't think that's a big deal though, definitely not blocking for this PR.

zuisong:
It's really weird. I'll keep looking for a solution.

zuisong:
The issue is fixed now, could you check if there are other issues?

blyxxyz (approved):
Looks good!

Cargo.toml
```diff
@@ -48,10 +48,14 @@ time = "0.3.16"
...
+[patch.crates-io.reqwest]
+git = "https://github.com/seanmonstar/reqwest"
+branch = "master"
+
 [dependencies.reqwest]
-version = "0.11.18"
+version = "0.12.2"
 default-features = false
-features = ["json", "multipart", "blocking", "socks", "cookies"]
+features = ["json", "multipart", "blocking", "socks", "cookies", "http2"]
```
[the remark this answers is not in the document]
ducaale: Yup, I can confirm xh will not pick up the configured `Web proxy (HTTP)` without enabling this feature.

Cargo.toml
```diff
@@ -48,10 +48,14 @@ time = "0.3.16"
 unicode-width = "0.1.9"
 url = "2.2.2"
 
+[patch.crates-io.reqwest]
+git = "https://github.com/seanmonstar/reqwest"
+branch = "master"
```
sorairolake, line 53: Perhaps, I think this should be removed before publishing on [crates.io](https://crates.io/).

> Note: [crates.io](https://crates.io/) does not allow packages to be published with dependencies on code published outside of [crates.io](https://crates.io/) itself ([dev-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies) are ignored). See the [Multiple locations](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) section for a fallback alternative for git and path dependencies.[^1]

[^1]: <https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies>

zuisong: Once the new version of reqwest is released, it will be removed.

This PR will not merge into master before the new version of reqwest is released.

zuisong:
Reqwest just released a new version 0.12.3, so this PR can be ready to go

Cargo.toml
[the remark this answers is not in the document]
ducaale: @zuisong can you add `macos-system-configuration` to the list of enabled features?

ducaale (approved):

PR #358, opened 2024-03-25
update http2-prior-knowledge test case

zuisong:
modify test cases so that it can be run offline


ducaale:
@zuisong did you run to some issues before closing this PR? It would be nice to remove the dependency on x.com if possible, thanks!

ducaale (approved):
Approved 🎉

PR #359, opened 2024-04-01
Directly bind to interface name on supported platforms

ducaale:
The latest release of reqwest (v0.12) has added `ClientBuilder::interface()` which removes the need for translating interface names to IP addresses before binding to them.

However, this is only supported on limited platforms i.e linux, android and fuchsia so we still can't remove the `network-interface` crate/feature.

Depends on #357

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,7 +81,7 @@ hyper-util = { version = "0.1.3", features = ["server"] }
 http-body-util = "0.1.1"
 
 [features]
-default = ["online-tests", "rustls"]
+default = ["online-tests", "rustls", "network-interface"]
 native-tls = ["reqwest/native-tls", "reqwest/native-tls-alpn"]
 rustls = ["reqwest/rustls-tls", "reqwest/rustls-tls-webpki-roots", "reqwest/rustls-tls-native-roots"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,10 +48,14 @@ time = "0.3.16"
 unicode-width = "0.1.9"
 url = "2.2.2"
 
+[patch.crates-io.reqwest]
+git = "https://github.com/seanmonstar/reqwest"
+branch = "master"
+
 [dependencies.reqwest]
-version = "0.11.18"
+version = "0.12.2"
 default-features = false
-features = ["json", "multipart", "blocking", "socks", "cookies"]
+features = ["json", "multipart", "blocking", "socks", "cookies", "http2"]
 
 [dependencies.syntect]
 version = "5.1"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -68,16 +72,18 @@ assert_cmd = "2.0.8"
...
 [features]
-# Ideally network-interface would be disabled by default on certain platforms
-# However: https://github.com/rust-lang/cargo/issues/1197
-default = ["online-tests", "rustls", "network-interface"]
+default = ["online-tests", "rustls"]
 native-tls = ["reqwest/native-tls", "reqwest/native-tls-alpn"]
 rustls = ["reqwest/rustls-tls", "reqwest/rustls-tls-webpki-roots", "reqwest/rustls-tls-native-roots"]
+
+# To be used by platforms that don't support binding to interface via SO_BINDTODEVICE
 network-interface = ["dep:network-interface"]
```
blyxxyz, line 87: Can we limit this so the dependency isn't pulled in on Linux (etc) even with the feature enabled?

blyxxyz: `network-interface` now shows up in `cargo tree --target x86_64-pc-windows-gnu` but not `cargo tree` so I think this works.

Cargo.toml
```diff
@@ -75,11 +77,11 @@ hyper-util = { version = "0.1.3", features = ["server"] }
 http-body-util = "0.1.1"
 
 [features]
-# Ideally network-interface would be disabled by default on certain platforms
-# However: https://github.com/rust-lang/cargo/issues/1197
```
blyxxyz, line 79: I think this comment still applies. Maybe we can also include a link to the xh issue with the NetBSD problems?

blyxxyz (commented):
`xh --interface=lo example.org` now hangs forever. (It used to give a connect error.) Any idea?

ducaale:
>xh --interface=lo example.org now hangs forever. (It used to give a connect error.) Any idea?

I'm not really sure. But if cURL is behaving the same, then it should be okay?

```sh
$ curl http://example.org/ --interface lo
curl: (28) Failed to connect to example.org port 80 after 131767 ms: Connection timed out
```

blyxxyz:
Another difference is that my (disconnected) ethernet interfaces now also hang forever while the old implementation gave a "Couldn't bind" error (because `NetworkInterface::show()` returned it without an IP address).

The crux seems to be that reqwest uses `SO_BINDTODEVICE` to basically hand the interface name directly to the kernel instead of the IP address thing we were doing.

But yeah, it matches curl, good point. So it's probably for the best.

blyxxyz (approved):
Looks good! We'll have to watch out for other platforms getting supported.

PR #360, opened 2024-04-07
enable stream when content-type is text/event-stream

### LINKED ISSUE #204: Support auto-enable --stream  when meet server sent events
#https://github.com/httpie/httpie/issues/376

zuisong:
closes #204


zuisong:
I have no idea how to write a test case for this feature.

src/main.rs
```diff
@@ -523,6 +529,11 @@ fn run(args: Cli) -> Result<i32> {
                     if history_print.response_headers {
                         printer.print_response_headers(prev_response)?;
                     }
+                    if args.stream.is_none()
+                        && is_text_event_stream_body(prev_response) == Some(true)
```
ducaale, line 533: Can you do this check inside `Printer::print_response_body()`? We are already extracting the response's `content_type` on L448 and use it to determine how to format the response body.

src/printer.rs
```diff
@@ -119,6 +119,10 @@ pub struct Printer {
 }
 
 impl Printer {
+    pub fn set_stream(&mut self, stream: bool) {
+        self.stream = stream
+    }
+
```
ducaale, line 125: This will not be necessary if we check the content type inside `Printer::print_response_body`.

src/main.rs
zuisong: Sure

src/printer.rs
```diff
@@ -538,15 +541,36 @@ enum ContentType {
...
     Unknown,
 }
 
 impl ContentType {
     fn is_text(&self) -> bool {
-        !matches!(
-            self,
-            ContentType::Unknown | ContentType::UrlencodedForm | ContentType::Multipart
-        )
+        match self {
```
zuisong, line 550: ensure the author updates this method when adding a content-type in the future.

ducaale (commented):
Looks good! just one final comment:

src/cli.rs
```diff
@@ -168,8 +168,8 @@ Example: --print=Hb"
     pub quiet: bool,
 
     /// Always stream the response body.
-    #[clap(short = 'S', long)]
-    pub stream: bool,
+    #[clap(short = 'S', long, default_missing_value = "true", num_args= 0..=1, require_equals = true)]
```
ducaale, line 171: Is there a reason to prefer `Option<bool>` over `bool`? It affects the generated help output, but in terms of functionality, the result is the same: `false`/`true` versus `None`/`Some(true)`

#### Before
```
-S, --stream      Always stream the response body
```
#### After
```
-S, --stream[=<STREAM>]   Always stream the response body [possible values: true, false]
```

zuisong: I think we need to provide users with the ability to choose to turn off automatic print body as stream.

|cli args|stream value|description|
|--|--|--|
xh -S| Some(true) | alway enable stream  |
xh -S=true| Some(true) |  alway enable stream   |
xh -S=false|  Some(false) |  alway disable stream 
xh|None| auto enable stream when meet stream body |

src/printer.rs
```diff
@@ -447,6 +447,9 @@ impl Printer {
         let compression_type = get_compression_type(response.headers());
         let mut body = decompress(response, compression_type);
 
+        // Automatically activate stream mode when it hasn't been set by the user and the content type is stream
+        let stream = self.stream.unwrap_or(content_type.is_stream());
```
zuisong, line 451: if user set `-S=false`, it will't print body as stream

ducaale (approved):

PR #361, opened 2024-04-09
use serde_json format full json

zuisong:
use serde_json format full json 

format json stream still uses jsonxf

closes #309

PR #362, opened 2024-04-12
Use serde-transcode to optimize JSON formatting

blyxxyz:
Followup to #361 (which I meant to look at but forgot).

Instead of parsing the JSON input into a `serde::Value` we can use [`serde-transcode`](https://lib.rs/crates/serde-transcode) to serialize straight from the input to the output. On an articial very large JSON file I use for testing this ends up five times faster. (Maybe slightly faster than `jsonxf` was?) I got this trick from jsonxf's [serde benchmark](https://github.com/gamache/jsonxf/blob/ab914dc7ed9dfaa642a37b316211328626dbef0c/benchmark/serdexf/src/main.rs).

(The rest of the PR looked great. Thanks, @zuisong!)

I also included a tiny theoretical bugfix for decompression that I wrote a while ago but didn't think was worth a PR, see the explanation in 398d5673800d06c06b39b91f24a097864b83679a

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1649,6 +1649,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde-transcode"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ducaale (approved):
Approved 🎉

PR #365, opened 2024-04-23
Remove `battila7/get-version-action`

sorairolake:
This uses deprecated commands (e.g., `set-output`) and actions (e.g., Node.js 16 actions). Also, the version number can be obtained from the [`GITHUB_REF_NAME`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) environment variable. So, I remove this action.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -107,9 +107,6 @@ jobs:
             strip "target/${{ matrix.job.target }}/release/xh"
           fi
 
-      - id: get_version
-        uses: battila7/get-version-action@v2
-
       - name: Package
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -118,7 +115,7 @@ jobs:
           else
             bin="target/${{ matrix.job.target }}/release/xh"
           fi
-          staging="xh-${{ steps.get_version.outputs.version }}-${{ matrix.job.target }}"
+          staging="xh-${GITHUB_REF_NAME}-${{ matrix.job.target }}"
```
ducaale, line 118: Can we use `github.ref_name` here?
```suggestion
          staging="xh-${{ github.ref_name }}-${{ matrix.job.target }}"
```

ducaale (approved):
Thanks!

PR #372, opened 2024-06-07
Apply TLS options even for non-HTTPS URLs

### LINKED ISSUE #368: `--verify=no` doesn't apply to redirects
Installed via homebrew.

```
❯ xh --version
xh 0.22.0
-native-tls +rustls

❯ xh -F --verify=no  cli.gorilla-llm.com
xh: error: error sending request for url (https://cli.gorilla-llm.com/)

Caused by:
    0: client error (Connect)
    1: invalid peer certificate: Expired

❯ xh -F --verify no  cli.gorilla-llm.com
xh: error: error sending request for url (https://cli.gorilla-llm.com/)

Caused by:
    0: client error (Connect)
    1: invalid peer certificate: Expired
```

No matter what I do I can't get `xh` to ignore the tls expiry date. This might be by design, I don't know. 

Thanks.

blyxxyz:
TLS-related options were only being applied for URLs with a `https://` scheme. But that's too cautious, because we could run into a HTTPS URL after a redirect.

I've fixed this by just removing the check. Another option is to also check `args.follow`, but I think it's nicer to do this stuff unconditionally. That way we can consistently give errors for bad options (e.g. `--verify=path/wiht/typo`).

Test case:
```shell
xh --verify=no --follow http://httpbin.org/redirect-to url==https://expired.badssl.com
```

Fixes #368.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -178,85 +178,83 @@ fn run(args: Cli) -> Result<i32> {
     let mut auth = None;
     let mut save_auth_in_session = true;
 
-    if url.scheme() == "https" {
-        let verify = args.verify.unwrap_or_else(|| {
-            // requests library which is used by HTTPie checks for both
-            // REQUESTS_CA_BUNDLE and CURL_CA_BUNDLE environment variables.
-            // See https://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification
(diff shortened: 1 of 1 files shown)


ducaale (approved):
LGTM

Issue #323, opened 2023-08-20
IRI output


ghost:
If I run this command, I get encoded output:

~~~
> xh -h https://wiktionary.org/wiki/Ῥόδος
HTTP/2.0 301 Moved Permanently
location: "https://www.wiktionary.org/wiki/\xe1\xbf\xac\xcf\x8c\xce\xb4\xce\xbf\xcf\x82"
~~~

but with this Go code:

~~~go
package main

import (
   "net/http"
   "os"
)

func main() {
   req, _ := http.NewRequest("HEAD", os.Args[1], nil)
   res, _ := new(http.Transport).RoundTrip(req)
   println(res.Header.Get("location"))
}
~~~

I get expected output:

~~~
> go run http.go https://wiktionary.org/wiki/Ῥόδος
https://www.wiktionary.org/wiki/Ῥόδος
~~~

<https://wikipedia.org/wiki/Internationalized_Resource_Identifier>

blyxxyz:
This one's tricky. Historically header values were defined to be ISO-8859-1. But the [latest recommendation](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.4) is this:
> A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

And that's how the library we use handles it. [It only allows string conversion if it's pure-ASCII.](https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.to_str)
And otherwise we print it in the form you see there.

HTTPie decodes it as ISO-8859-1, which is wrong in this case:
```console
$ http -h https://wiktionary.org/wiki/Ῥόδος
HTTP/1.1 301 Moved Permanently
[...]
location: https://www.wiktionary.org/wiki/á¿¬ÏÎ´Î¿Ï
```
And that's also how the dev tools of Firefox and Chromium handle it:
![image](https://github.com/ducaale/xh/assets/15526524/07cd4323-7225-47f7-bf84-4f35d6a986cb)

So I see a few options:
- Change nothing, pedantically follow the RFC.
- Try to decode as UTF-8 (with `std::str::from_utf8(value.as_bytes())`). Other bytes are somewhat unlikely to be valid UTF-8, so this won't have that many false negatives, hopefully.
- Try to decode as ISO-8859-1, to match the browser tools and HTTPie.

I'm torn between all three, I don't know what's best.

(The current behavior deserves a mention on https://github.com/ducaale/xh/issues/322.)

1268:
[...]

PR #371, opened 2024-06-05
Add `--debug` option for logging and backtraces

blyxxyz:
Add `env_logger` to be able to print the logs that our libraries already generate.

Add some logging to the application code. We'll probably want more log messages, and we might want to upgrade some of them from trace to debug.

Add a `--debug` flag that automatically enables `env_logger` as well as backtraces for `anyhow` and panics.

```console
$ xh --debug :
[2024-06-05T13:54:40Z DEBUG xh] xh 0.22.0 -native-tls +rustls
[2024-06-05T13:54:40Z DEBUG xh] Cli {
        httpie_compat_mode: false,
        [...]
    }
[2024-06-05T13:54:40Z DEBUG xh] Complete URL: http://localhost/
[2024-06-05T13:54:40Z DEBUG xh] HTTP method: GET
[2024-06-05T13:54:40Z DEBUG reqwest::connect] starting new connection: http://localhost/
[2024-06-05T13:54:40Z DEBUG hyper_util::client::legacy::connect::dns] resolving host="localhost"
[...]
xh: error: error sending request for url (http://localhost/)

Caused by:
    0: client error (Connect)
    1: tcp connect error: Connection refused (os error 111)
    2: Connection refused (os error 111)

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.82/src/error.rs:565:25
[...]
```

We could ask users to post `--debug` output when reporting bugs.

PR #375, opened 2024-06-21
decode header value with utf-8

zuisong:
- [x] add test case 

---
For reference, this behavior is the same as the redirection behavior of reqwest.

see https://github.com/seanmonstar/reqwest/blob/29d4cff234b37065632512f002b9785700c51aa8/src/async_impl/client.rs#L2624

---

related issue: https://github.com/ducaale/xh/issues/323

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -13,7 +13,6 @@ use reqwest::header::{
 use reqwest::Version;
 use url::Url;
 
-use crate::utils::HeaderValueExt;
 use crate::{
     buffer::Buffer,
     cli::FormatOptions,
@@ -337,7 +336,7 @@ impl Printer {
                 header_string.push_str(key.as_str());
             }
             header_string.push_str(": ");
-            match value.to_utf8_str() {
+            match value.to_str() {
                 Ok(value) => header_string.push_str(value),
                 #[allow(clippy::format_push_string)]
                 Err(_) => header_string.push_str(&format!("{:?}", value)),
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -336,8 +336,8 @@ impl Printer {
                 header_string.push_str(key.as_str());
             }
             header_string.push_str(": ");
-            match value.to_str() {
-                Ok(value) => header_string.push_str(value),
+            match String::from_utf8(value.as_bytes().to_vec()) {
+                Ok(value) => header_string.push_str(&value),
                 #[allow(clippy::format_push_string)]
                 Err(_) => header_string.push_str(&format!("{:?}", value)),
             }
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -13,6 +13,7 @@ use reqwest::header::{
 use reqwest::Version;
 use url::Url;
 
+use crate::utils::HeaderValueExt;
 use crate::{
     buffer::Buffer,
     cli::FormatOptions,
@@ -336,8 +337,8 @@ impl Printer {
                 header_string.push_str(key.as_str());
             }
             header_string.push_str(": ");
-            match String::from_utf8(value.as_bytes().to_vec()) {
-                Ok(value) => header_string.push_str(&value),
+            match value.to_utf8_str() {
+                Ok(value) => header_string.push_str(value),
                 #[allow(clippy::format_push_string)]
                 Err(_) => header_string.push_str(&format!("{:?}", value)),
             }
--- a/src/redirect.rs
+++ b/src/redirect.rs
@@ -51,8 +51,8 @@ fn get_next_request(mut request: Request, response: &Response) -> Option<Request
         response
             .headers()
             .get(LOCATION)
-            .and_then(|location| location.to_str().ok())
-            .and_then(|location| request.url().join(location).ok())
+            .and_then(|location| String::from_utf8(location.as_bytes().to_vec()).ok())
+            .and_then(|location| request.url().join(&location).ok())
     };
 
     match response.status() {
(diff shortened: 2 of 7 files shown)


src/redirect.rs
```diff
@@ -51,8 +51,8 @@ fn get_next_request(mut request: Request, response: &Response) -> Option<Request
         response
             .headers()
             .get(LOCATION)
-            .and_then(|location| location.to_str().ok())
-            .and_then(|location| request.url().join(location).ok())
+            .and_then(|location| String::from_utf8(location.as_bytes().to_vec()).ok())
+            .and_then(|location| request.url().join(&location).ok())
```
blyxxyz, line 55: I'm realizing now that if decoding or joining fails we should `log::warn!()` (after https://github.com/ducaale/xh/pull/371). We can pick that up in another PR though.

src/printer.rs
```diff
@@ -336,8 +336,8 @@ impl Printer {
                 header_string.push_str(key.as_str());
             }
             header_string.push_str(": ");
-            match value.to_str() {
-                Ok(value) => header_string.push_str(value),
+            match String::from_utf8(value.as_bytes().to_vec()) {
```
blyxxyz, line 339: You can use https://doc.rust-lang.org/std/str/fn.from_utf8.html to go directly from `&[u8]` to `&str`

blyxxyz (commented):
Nice catch!

zuisong:
I'm wondering if we should change all calls to HeaderValue.to_str() to HeaderValue.to_utf8_str(). 

I think it's a good idea.

For example, this will allow us to download filenames that support UTF-8 encoding. 
https://github.com/ducaale/xh/blob/ca629c803f78834a32994f6ca6754cd6fc896e63/src/download.rs#L34

blyxxyz:
Officially you're not supposed to do that so we should decide on a case-by-case basis.

For filenames we can actually pass the bytes through directly depending on the platform. But we should check what different servers and clients do.

zuisong:
Yes, you're right. 

I decided to use UTF-8 encoding to decode the CONTENT_DISPOSITION response header, which is very useful for downloading files with Unicode names.
And Chrome implements it the same way.



before
```shell
❯ xh  httpbingo.org/response-headers 'content-disposition==attachment; filename="🥰.json"' -d
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Connection: close
Content-Disposition: "attachment; filename=\"\xf0\x9f\xa5\xb0.json\""
Content-Length: 76
Content-Type: application/json; charset=utf-8
Date: Fri, 21 Jun 2024 16:07:28 GMT
Fly-Request-Id: 01J0XRV0MHF1VBYPJY576KHWSM-sjc
Server: Fly/ebd3372a (2024-06-19)
Via: 1.1 fly.io

Downloading 76 B to "response-headers.json-3"
Done. 76 B in 0.00108s (68.80 KiB/s)
```

after
```shell
❯ /Users/chen/.rust-target/debug/xh httpbingo.org/response-headers 'content-disposition==attachment; filename="🥰.json"' -d
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Connection: close
Content-Disposition: attachment; filename="🥰.json"
Content-Length: 76
Content-Type: application/json; charset=utf-8
Date: Fri, 21 Jun 2024 16:06:44 GMT
Fly-Request-Id: 01J0XRSP2WEE21V3XAHNGBFCBY-sjc
Server: Fly/ebd3372a (2024-06-19)
Via: 1.1 fly.io

Downloading 76 B to "🥰.json"
Done. 76 B in 0.00144s (51.63 KiB/s)
```

tests/cli.rs
```diff
@@ -1657,6 +1678,90 @@ fn body_from_raw() {
...
+        .stdout(contains("curl https://exmaple.com/ -H 'hello: 你好'"))
+        .success();
+
+    get_command()
+        .args(["https://exmaple.com/", "hello:你好", "--curl-long"])
+        .assert()
+        .stdout(contains("curl https://exmaple.com/ --header 'hello: 你好'"))
+        .success();
+}
+
```
blyxxyz, line 1764: Can you add these to the test cases at the bottom of `to_curl.rs` instead?

blyxxyz (approved):
Thank you!

src/printer.rs
```diff
@@ -336,7 +337,7 @@ impl Printer {
                 header_string.push_str(key.as_str());
             }
             header_string.push_str(": ");
-            match value.to_str() {
+            match value.to_utf8_str() {
```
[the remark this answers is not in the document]
zuisong: done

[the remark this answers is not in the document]
zuisong: I still feel that outputting in UTF-8 is better.

As a reference, the output of curl looks like this:
```text
❯ curl  http://httpbingo.org/response-headers --url-query  hello=😀 -v
* Host httpbingo.org:80 was resolved.
* IPv6: (none)
* IPv4: 66.241.125.232
*   Trying 66.241.125.232:80...
* Connected to httpbingo.org (66.241.125.232) port 80
> GET /response-headers?hello=%f0%9f%98%80 HTTP/1.1
> Host: httpbingo.org
> User-Agent: curl/8.6.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< access-control-allow-credentials: true
< access-control-allow-origin: *
< content-type: application/json; charset=utf-8
< hello: 😀
< date: Sat, 22 Jun 2024 01:32:27 GMT
< content-length: 32
< server: Fly/ebd3372a (2024-06-19)
< via: 1.1 fly.io
< fly-request-id: 01J0YS5HC16CHAQGZWANZ140DP-nrt
< 
{
  "hello": [
    "😀"
  ]
}
* Connection #0 to host httpbingo.org left intact
```

PR #377, opened 2024-07-04
Decode headers as latin1/UTF-8, show real reason phrase

### LINKED ISSUE #323: IRI output

blyxxyz:
image

External changes:

- We now print the actual reason phrase sent by the server instead of guessing it from the status code. That is, if servers reply with "200 Wonderful" instead of "200 OK" then we show that. This is especially useful for status codes that xh doesn't recognize.

- Header values are now decoded as latin1, with the UTF-8 decoding also shown if applicable. (Resolves #323.)

- A new FAQ file with an entry that explains header value encoding. Header output now hyperlinks to this entry when relevant and if supported by the terminal.

Under the hood we now color headers manually. It's still hooked up to the `.tmTheme` files but not to the `.sublime-syntax` file. This lets us highlight the latin1 header values differently. In the future we could use the same approach to optimize JSON highlighting.

I'm unsure about the position of the hyperlink. Currently it's the text "UTF-8" in `<latin1 value> (UTF-8: <utf-8 value>)`. But that means it's only shown if the value can be decoded as UTF-8. An alternative is to turn the latin1 value itself into a hyperlink, but that's confusing if the value itself is already a URL (which is a common case for the `Location` header).

I also don't feel that our text is quite distinct enough from the header value in the default `ansi` theme. Though the hyperlink does help to set it apart.

We should use a stable URL that keeps working in the future. I linked to `FAQ.md` on this repo's master branch (https://github.com/ducaale/xh/blob/master/FAQ.md#header-value-encoding, won't work until it's merged) but maybe there's a better option?


blyxxyz:
(Sorry for the big diff for such a small feature! I'm hoping it'll be useful for other things in the future.)

ducaale:
Apologies for taking so long to review this 😬. I’ll make sure to leave a review by this weekend.

assets/themes/solarized.tmTheme
```diff
@@ -174,6 +174,18 @@
...
+            <!-- FIXME: does this color fit the theme? -->
+            <dict>
+                <key>name</key>
+                <string>Error</string>
+                <key>scope</key>
+                <string>error</string>
+                <key>settings</key>
+                <dict>
+                    <key>foreground</key>
+                    <string>#01000000</string>
```
ducaale, line 186: I'm not sure what color would work best. But since this is rarely shown, defaulting to Ansi-red sound good 👍

src/formatting.rs
ducaale, line 1: Could we rename this file to `formatting/mod.rs`?

ducaale (commented):
Looks good! Loving the new FAQ.md file

>I'm unsure about the position of the hyperlink. Currently it's the text "UTF-8" in <latin1 value> (UTF-8: <utf-8 value>). But that means it's only shown if the value can be decoded as UTF-8. 

I think linking the FAQ file is only necessary when xh prints extra information that wasn't part of the request/response, e.g `(UTF-8: <utf-8 value>)`

src/formatting.rs
blyxxyz: Done!

ducaale (approved):
Approved 🎉

PR #387, opened 2024-10-30
support path-as-is

zuisong:
A clever method to avoid URL normalization. see https://github.com/epi052/feroxbuster/issues/878#issuecomment-1518485667

Here's a cheeky way to implement `path-as-is`.
      
image


zuisong:
It does not support the "./" path. This implementation differs from httpie and curl.

image

src/cli.rs
```diff
@@ -724,23 +730,47 @@ fn parse_method(method: &str) -> Option<Method> {
...
+}
+
+fn build_raw_url(url_string: &str) -> std::result::Result<Url, url::ParseError> {
+    let url_parsed = Url::parse(url_string)?;
+    let after_scheme = url_string
+        .split_once("://")
+        .map(|it| it.1)
+        .unwrap_or(url_string);
+    let path = after_scheme.split_once('/').unwrap().1;
+    let mut url = Url::from_file_path(format!("/{}", path)).unwrap();
```
zuisong, line 765: Special handling is required for the Windows system.

zuisong:
The effect cannot be achieved like curl or xh, so it has been temporarily disabled.

blyxxyz:
Interesting (and too bad you can't make it work).

Hyper uses a [`Uri`](https://docs.rs/hyper/latest/hyper/struct.Request.html#method.uri_mut) type that doesn't have this limitation. But I don't think we can access it directly.

reqwest issue: https://github.com/seanmonstar/reqwest/issues/2396

zuisong:
Indeed, this is quite sad.

This implementation currently does not work on the Windows platform and does not support "./" paths. 

This feature can only be implemented after changes are made upstream.

PR #396, opened 2025-01-02
Security update: upgrade to ruzstd v0.7.3

zuisong:
Security update
RUSTSEC-2024-0400: ruzstd uninit and out-of-bounds memory reads 
KillingSpark/zstd-rs#75

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1668,9 +1668,9 @@ dependencies = [
 
 [[package]]
 name = "ruzstd"
-version = "0.7.2"
+version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "99c3938e133aac070997ddc684d4b393777d293ba170f2988c8fd5ea2ad4ce21"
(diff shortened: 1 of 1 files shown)


zuisong:
`#[expect]` attribute stabilized in Rust 1.81
exclusive range patterns stabilized in Rust 1.80

Should we increase the MSRV to 1.81? 
Alternatively, we could use ruzstd v0.7.3.

blyxxyz (approved):

ducaale:
Thanks @zuisong, this has now been released as xh v0.23.1

>Should we increase the MSRV to 1.81?

I don't have strong opinions but I was thinking of holding off until one of our dependencies requires us to bump up our MSRV. What do you think @blyxxyz?

blyxxyz:
For security fixes it's best not to bump the MSRV unless we have to, to make it as easy as possible for people to update.

I'm still not really sure what to do with MSRVs beyond that. If a dependency forces it then that's a good enough reason. It's nice that our current MSRV is [supported by old Ubuntu releases](https://launchpad.net/ubuntu/+source/rustc) but I don't know how many users that actually helps. I don't know any compelling reasons to bump it right now but maybe we don't have to be so strict about it, IDK.

PR #397, opened 2025-01-05
upgrade cookie_store to 0.21.1

kranurag7:
upgrading cookie_store to 0.21.1 via this commit because it was pulling idna which contained a CVE.


src/session.rs
```diff
@@ -281,7 +281,7 @@ impl Session {
             Cookies::Map(_) => unreachable!(),
             Cookies::List(cookies) => cookies.iter().map(|cookie| {
                 let mut cookie_builder =
-                    cookie_store::RawCookie::build(cookie.name.clone(), cookie.value.clone());
+                    cookie_store::RawCookie::build(cookie.name.clone());
```
kranurag7, line 284: this was changed in the 0.21.1 release of cookie_store so required this modification, not entirely sure here if I'm doing it correctly. 😕

blyxxyz: The tests fail so looks like this is wrong yeah. According to the [docs](https://docs.rs/cookie_store/0.21.1/cookie_store/struct.RawCookie.html#method.build) we only need to turn the argument into a tuple:
```suggestion
                    cookie_store::RawCookie::build((cookie.name.clone(), cookie.value.clone()));
```

blyxxyz (commented):
This gets rid of 2 dependencies and 5 duplicated dependencies so definitely a nice change, thank you!

I don't think that [the CVE](https://rustsec.org/advisories/RUSTSEC-2024-0421.html) is a security concern for us, since we run the hostname through a URL parse before we set it on the cookie and we don't have any crazy authentication system where you can benefit from *not* matching a domain. So we don't need to rush out another release.

src/session.rs
kranurag7: @blyxxyz thank you for looking into this. I think while compiling I'm getting this deprecation warning via clippy which will fail the CI. 

```bash
warning: use of deprecated method `cookie::builder::CookieBuilder::<'c>::finish`: `CookieBuilder` can be passed in to methods expecting a `Cookie`; for other cases, use `CookieBuilder::build()`
   --> src/session.rs:308:37
    |
308 |                     &cookie_builder.finish(),
    |                                     ^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

blyxxyz: `&cookie_builder.into()` works.

ducaale:
@kranurag7 do you mind resolving current merge conflicts? We can fix any clippy warnings after merging this PR

blyxxyz (approved):
Looks great!

Issue #399, opened 2025-01-09
Reason for vendoring reqwest_cookie_store is now obsolete


nc7s:
https://github.com/ducaale/xh/blob/196ca1f162a4f1c27ecaf4e6ca79631ffa0a1ee6/src/vendored/reqwest_cookie_store.rs#L1-L6

Mentioned PR was long merged.

Issue #400, opened 2025-01-18
Cookies without explicit path are stored in session without _any_ path


otaconix:
As the title explains, xh doesn't store _any_ path for session cookies when `Set-Cookie` headers don't explicitly set a `path` attribute. This means that cookies stored by xh may erroneously be sent to wrong endpoints.

---

_Original issue below, which was bogus_

I was playing around with an endpoint that:
1. sets a cookie, then
2. redirects to someplace else

## Example

```
λ xh \
          --session ./xh-session \
          -F \
          --all \
          --print hbHB \
          POST http://localhost:8080/jwt/with-redirect \
          key=value
POST /jwt/with-redirect HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Content-Length: 164
Content-Type: application/json
Host: localhost:8080
User-Agent: xh/0.23.1

{
    "key": "value"
}



HTTP/1.1 303 See Other
Content-Length: 0
Location: http://localhost:9080/other-url
Set-Cookie: Bearer=JWT-here;Version=1



GET /other-url HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Host: localhost:9080
User-Agent: xh/0.23.1

HTTP/1.1 200 OK
Content-Language: en-NL
Content-Length: 3
Content-Type: application/json
Date: Sat, 18 Jan 2025 17:45:14 GMT

{}
```

NB: actual contents redacted, so `Content-Length` isn't correct for the redacted values.

## Expected outcome

I would've expected the cookie I got from the initial request to be sent to the second one. Rerunning the command above does in fact send the cookie in the request (since it's been saved from the first request).

## Actual outcome

Cookie from earlier request isn't sent to the second one.

## Additional info

This may not be a bug, it all depends on what one expects from `xh`. But I was a little surprised, so here's the issue anyway 🙂 

```
λ xh --version                                                                                                                                  at 18:50:22
xh 0.23.1
-native-tls +rustls
```

ducaale:
I think this happens because the `set-cookie` header doesn't specify a path, so it **correctly** defaults to `/jwt/with-redirect`. As a result, any subsequent requests that do not fall under `/jwt/with-redirect` directory will not include the cookie. See https://stackoverflow.com/a/43336097/5915221

[...]

PR #401, opened 2025-01-19
Remove vendored reqwest_cookie_store & store default paths for cookies without an explicit path attribute

otaconix:
Copy-paste of note in commit:

> Note that this commit does not alter the storage format of sessions.
> Perhaps requiring a cookie path at all times would be a good idea, but
> I haven't done that here, because I don't want to break existing
> sessions for users.

Fixes: #399 #400


otaconix:
This PR fixes two separate issues. Let me know if you'd prefer separate PR's!

ducaale (approved):
Thanks!

PR #402, opened 2025-01-20
Generate man page with reproducible timestamp

nc7s:
As [reported in Debian](https://bugs.debian.org/1092917), current generation code uses the time point it executed at as the date of the man page, so a build on e.g. 2025-01-20 would differ from a build on 2025-01-19, even if the content hasn't actually changed. This changes it to use the `SOURCE_DATE_EPOCH` environment variable if present, as explained in the [Reproducible Builds docs](https://reproducible-builds.org/docs/source-date-epoch/).


blyxxyz (approved):
Thanks!

PR #403, opened 2025-01-22
support compress request body 

zuisong:
implements https://httpie.io/docs/cli/compressed-request-body


src/main.rs
```diff
@@ -508,6 +510,27 @@ fn run(args: Cli) -> Result<i32> {
 
         let mut request = request_builder.headers(headers).build()?;
 
+        if args.compress >= 1 && request.headers().get(CONTENT_ENCODING).is_none() {
+            let mut compressed = false;
+            if let Some(body) = request.body_mut() {
+                if let Some(body_bytes) = body.as_bytes() {
```
blyxxyz, line 516: Can we make this work for streaming requests as well (i.e. `xh -x : @file.txt`)? In that case it might be best to just compress unconditionally even with a single `-x` so that we don't have to buffer it.

(HTTPie does buffer the whole file in this case.)

zuisong: I think it's possible; the latest commit supports this. 😄

blyxxyz: Awesome, that gets us the same behavior as HTTPie. But we might want to diverge from HTTPie by streaming the body in this case instead of buffering it, so that it's possible to upload files that don't fit in RAM. That should be possible with [`ReqwestBody::new`](https://docs.rs/reqwest/0.12.12/reqwest/blocking/struct.Body.html#method.new).

(I also suggested this in https://github.com/httpie/cli/pull/1613 so we could wait to see what they think.)

src/main.rs
```diff
@@ -508,6 +510,27 @@ fn run(args: Cli) -> Result<i32> {
 
         let mut request = request_builder.headers(headers).build()?;
 
+        if args.compress >= 1 && request.headers().get(CONTENT_ENCODING).is_none() {
+            let mut compressed = false;
+            if let Some(body) = request.body_mut() {
+                if let Ok(body_bytes) = body.buffer() {
```
blyxxyz, line 516: Looks like `.buffer()` only returns `Err` for IO errors, so `let body_bytes = body.buffer()?;` is probably better.

src/main.rs
zuisong: We can directly compress the stream without buffering it, which would make it different from httpie's implementation. 
Should we do it this way in this PR?

src/main.rs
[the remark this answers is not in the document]
zuisong: The into_reader() method of reqwest#Body is pub(crate). We cannot access it. Currently, it is not possible to achieve our goal.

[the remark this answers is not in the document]
blyxxyz: We'd have to wrap the file after opening it and before turning it into a `Body`. Currently we open the file here: https://github.com/ducaale/xh/blob/f75d09216ff19f89fea9c606dc05c54a5aa9f53b/src/main.rs#L444

But it's also OK to leave this out for now.

tests/cli.rs
```diff
@@ -779,6 +780,43 @@ fn successful_digest_auth() {
         .stdout(contains("HTTP/1.1 200 OK"));
 }
 
+#[cfg(feature = "online-tests")]
```
ducaale, line 783: Can you move the tests into a separate file under `tests/cases`?

src/cli.rs
```diff
@@ -182,6 +182,14 @@ Example: --print=Hb"
     #[clap(short = 'S', long = "stream", name = "stream")]
     pub stream_raw: bool,
 
+    ///  Content compressed (encoded) with Deflate algorithm.
+    ///  The Content-Encoding header is set to deflate.
+    ///
+    ///  Compression is skipped if it appears that compression ratio is negative.
+    ///  Compression can be forced by repeating this option.
```
ducaale, line 189: Should we also mention that compression cannot be forced if the `Content-Encoding` request header is present?

tests/cli.rs
```diff
@@ -3483,6 +3521,198 @@ fn zstd() {
...
+        ])
+        .assert()
+        .stdout(indoc::formatdoc! {r#"
+            HTTP/1.1 200 OK
+            Date: N/A
+            Content-Type: text/plain
+            Content-Length: 1004
+
+            key={c}
+        "#, c = "1".repeat(1000),});
```
ducaale, line 3667: Is the duplicate command invocation necessary here?

tests/cli.rs
```diff
@@ -3483,6 +3521,198 @@ fn zstd() {
...
+
+fn zlib_decode(bytes: Vec<u8>) -> std::io::Result<String> {
+    let mut z = flate2::read::ZlibDecoder::new(&bytes[..]);
+    let mut s = String::new();
+    z.read_to_string(&mut s)?;
+    Ok(s)
+}
+
+#[test]
+fn compress_request_body() {
```
ducaale, line 3584: Would it be possible to break this into multiple test cases?

src/main.rs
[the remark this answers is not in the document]
ducaale: @zuisong Can you add a TODO comment about compressing the file body without buffering it?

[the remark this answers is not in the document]
zuisong: Happy Chinese New Year! 🥳🥳🥳

I'm currently on vacation, but I will update the code as soon as possible regarding your comments later.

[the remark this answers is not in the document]
ducaale: No rush at all. Enjoy your vacation!

tests/cases/compress_request_body.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    get_command()
+        .arg(server.base_url())
+        .arg("-x")
+        .arg(format!("@{}", filename.to_string_lossy()))
+        .assert()
+        .success();
+}
+#[test]
+fn test_cannot_combine_compress_with_multipart() {
```
zuisong, line 200: Should we handle the situation where multipart requests, specified by headers and body, are currently unprocessed? 
Reading the request header Content-Type before compression should handle these cases.

zuisong: Gave it a try, but it seems I can't construct a multipart request on my own at the moment. won't worry about this anymore.

```
xh post https://postman-echo.com/post  --raw='--16e4fcaf49bf49a5b0d13aeeb3b8f817
Content-Disposition: form-data; name="a"

1
--16e4fcaf49bf49a5b0d13aeeb3b8f817--
'  content-type:"multipart/form-data; boundary=16e4fcaf49bf49a5b0d13aeeb3b8f817"
```

zuisong:
I've incorporated all your suggested improvements. 😄 

Could you please review for any other areas needing improvement?

Should I squash into a single commit before merging the code?

src/main.rs
```diff
@@ -508,6 +510,23 @@ fn run(args: Cli) -> Result<i32> {
 
         let mut request = request_builder.headers(headers).build()?;
 
+        if args.compress >= 1 && request.headers().get(CONTENT_ENCODING).is_none() {
```
blyxxyz, line 513: - If the user passed `--compress` but also a `Content-Encoding` header we should print a warning to let them know that their request isn't compressed and why, just like with `--continue` and `Range`.
  ```rust
  log::warn!("--compress can't be used with a 'Content-Encoding:' header. --compress will be disabled.");
  ```
  Maybe we don't even need to mention it in the docs if we do this, not sure.

- HTTPie actually handles this differently, it forces the compressed Content-Encoding. But I think I like this behavior better. We should still put a comment about the difference.

- I think we can use [`contains_key`](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderMap.html#method.contains_key) here

src/main.rs
```diff
@@ -508,6 +510,23 @@ fn run(args: Cli) -> Result<i32> {
...
+                let body_bytes = body.buffer()?;
+                let mut encoder = ZlibEncoder::new(Vec::new(), Default::default());
+                encoder.write_all(body_bytes)?;
+                let output = encoder.finish()?;
+                if output.len() < body_bytes.len() || args.compress >= 2 {
+                    let _ = std::mem::replace(body, ReqwestBody::from(output));
+                    request
+                        .headers_mut()
+                        .entry(CONTENT_ENCODING)
+                        .or_insert(HeaderValue::from_static("deflate"));
```
blyxxyz, line 525:
```suggestion
                    *body = ReqwestBody::from(output);
                    request
                        .headers_mut()
                        .insert(CONTENT_ENCODING, HeaderValue::from_static("deflate"));
```

tests/cases/compress_request_body.rs
```diff
@@ -0,0 +1,209 @@
...
+                let body = zlib_decode(compressed_body).unwrap();
+                hyper::Response::builder()
+                    .header("date", "N/A")
+                    .header("Content-Type", "text/plain")
+                    .body(body.into())
+                    .unwrap()
+            }
+            "/normal" => {
+                let body = req.body_as_string().await;
+                hyper::Response::builder()
```
blyxxyz, line 34: Can we assert here that the request doesn't have a `Content-Encoding` header?

tests/cases/compress_request_body.rs
```diff
@@ -0,0 +1,209 @@
...
+        .stdout(indoc::formatdoc! {r#"
+            HTTP/1.1 200 OK
+            Date: N/A
+            Content-Type: text/plain
+            Content-Length: 1010
+
+            {{"key":"{c}"}}
+        "#, c = "1".repeat(1000),});
+}
+#[test]
```
blyxxyz, line 67:
```suggestion
}

#[test]
```

tests/cases/compress_request_body.rs
```diff
@@ -0,0 +1,209 @@
...
+        .unwrap();
+
+    get_command()
+        .arg(server.base_url())
+        .arg("-x")
+        .arg(format!("@{}", filename.to_string_lossy()))
+        .assert()
+        .success();
+}
+#[test]
```
blyxxyz, line 199:
```suggestion
}

#[test]
```

tests/cases/compress_request_body.rs
```diff
@@ -0,0 +1,209 @@
...
+        .arg(format!("@{}", filename.to_string_lossy()))
+        .assert()
+        .success();
+}
+
+#[test]
+fn compress_body_from_file_unless_compress_rate_less_1() {
+    let server = server::http(|req| async move {
+        assert_eq!("Hello world\n", req.body_as_string().await);
+        hyper::Response::default()
```
blyxxyz, line 178: Can you also check for the header here?

blyxxyz (commented):
Thank you, almost there!

zuisong:
thanks, I've incorporated all your suggested improvements.

ducaale:
The changes look good to me. Is there anything else we want to address before this can be merged @blyxxyz?

src/cli.rs
```diff
@@ -182,6 +182,15 @@ Example: --print=Hb"
...
+    ///  Note: Compression cannot be forced if the Content-Encoding request header is present.
```
blyxxyz, line 190:
```suggestion
    ///
    ///  Note: Compression cannot be used if the Content-Encoding request header is present.
```

I think this warning isn't specifically related to forcing?

blyxxyz (approved):
Looks great!

src/main.rs
```diff
@@ -508,6 +510,24 @@ fn run(args: Cli) -> Result<i32> {
 
         let mut request = request_builder.headers(headers).build()?;
 
+        if args.compress >= 1 {
+            if request.headers().contains_key(CONTENT_ENCODING) {
+                log::warn!("--compress can't be used with a 'Content-Encoding:' header. --compress will be disabled.");
```
blyxxyz, line 515:
```suggestion
                // HTTPie overrides the original Content-Encoding header in this case
                log::warn!("--compress can't be used with a 'Content-Encoding:' header. --compress will be disabled.");
```

PR #404, opened 2025-02-12
Remove cargo-cross in favour of using GitHub ARM runner

ducaale:


ducaale:
It seems the random CI failure might be related to a deadlock when emulating arm32 using qemu (I originally thought it was related to cargo-cross). I will continue investigating this locally.

PR #411, opened 2025-03-05
Fix crash on empty zstd response body

### LINKED ISSUE #409: panic with compressed response from server
When calling a service that returns a `zstd` compressed response, xh panics.

For example:

```
xh -v HEAD http://hello.com/file.css
HEAD /file.css HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Host: hello.com
User-Agent: xh/0.24.0

HTTP/1.1 200 OK
Cache-Control: max-age=604800
Connection: keep-alive
Content-Disposition: inline; filename=file.css
Content-Encoding: zstd
Content-Length: 998
Content-Type: text/css; charset=utf-8
Date: Mon, 03 Mar 2025 18:50:39 GMT
Last-Modified: Fri, 28 Feb 2025 12:28:50 GMT
Server: gunicorn
Vary: Accept-Encoding

thread 'main' panicked at src/decoder.rs:150:79:
called `Result::unwrap()` on an `Err` value: ReadFrameHeaderError(MagicNumberReadError(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Same request with curl works fine.

blyxxyz:
Fixes a panic for `xh head https://httpbin.dev/zstd`.

`ZstdDecoder::new()` returns a `Result`. We used to panic on this, but it needs to be a `Read` error instead, so we can suppress the error for an empty input the way we do for other decoders.

Our existing approach couldn't handle this, so I ended up refactoring the system. I think it's cleaner now, though still weird.

We now also preserve the original decoder error instead of `.to_string()`ing it, or strip it completely if there was an I/O error. That should improve the error reporting.

This PR also contains a fix for the binary data detection: in certain circumstances, like an invalid streaming brotli response, it would print the "binary data" note instead of an error.

Resolves #409.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -1,4 +1,6 @@
+use std::cell::Cell;
 use std::io::{self, Read};
+use std::rc::Rc;
 use std::str::FromStr;
 
 use brotli::Decompressor as BrotliDecoder;
@@ -6,7 +8,7 @@ use flate2::read::{GzDecoder, ZlibDecoder};
 use reqwest::header::{HeaderMap, CONTENT_ENCODING, CONTENT_LENGTH, TRANSFER_ENCODING};
 use ruzstd::{FrameDecoder, StreamingDecoder as ZstdDecoder};
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub enum CompressionType {
     Gzip,
     Deflate,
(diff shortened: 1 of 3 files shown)


ducaale (approved):
Looks good!

PR #416, opened 2025-04-10
download file parse filename add RFC 5987 support

### LINKED ISSUE #415: filename for utf8 content-disposition is not correct
I use xh --download to download a file.
and the filename is Download.bin.
It should be sql04071104.csv
Image

zuisong:
support percent-encoded filename

close #415

### DIFF
--- a/src/content_disposition.rs
+++ b/src/content_disposition.rs
@@ -1,4 +1,4 @@
-use percent_encoding::percent_decode;
+use percent_encoding::percent_decode_str;
 
 /// Parse filename from Content-Disposition header
 /// Prioritizes filename* parameter if present, otherwise uses filename parameter
@@ -10,8 +10,8 @@ pub fn parse_filename_from_content_disposition(content_disposition: &str) -> Opt
 
     // First try to find filename* parameter
     for part in parts.iter() {
-        if part.starts_with("filename*=") {
-            if let Some(filename) = parse_encoded_filename(part) {
+        if let Some(value) = part.strip_prefix("filename*=") {
+            if let Some(filename) = parse_encoded_filename(value) {
                 return Some(filename);
             }
         }
@@ -19,8 +19,8 @@ pub fn parse_filename_from_content_disposition(content_disposition: &str) -> Opt
 
     // If filename* is not found or parsing failed, try regular filename parameter
     for part in parts {
-        if part.starts_with("filename=") {
-            return parse_regular_filename(part);
+        if let Some(value) = part.strip_prefix("filename=") {
+            return parse_regular_filename(value);
         }
     }
 
@@ -29,10 +29,7 @@ pub fn parse_filename_from_content_disposition(content_disposition: &str) -> Opt
 
 /// Parse regular filename parameter
 /// Handles both quoted and unquoted filenames
-fn parse_regular_filename(part: &str) -> Option<String> {
-    let filename = part.trim_start_matches("filename=");
-    // Remove quotes if present
-    //
+fn parse_regular_filename(filename: &str) -> Option<String> {
     // Content-Disposition: attachment; filename="file with \"quotes\".txt"  // This won't occur
     // Content-Disposition: attachment; filename*=UTF-8''file%20with%20quotes.txt  // This is the actual practice
     //
@@ -41,6 +38,8 @@ fn parse_regular_filename(part: &str) -> Option<String> {
     // It's not a standard practice
     // It rarely occurs in real-world scenarios
     // When filenames contain special characters, they should use the filename* parameter
+
+    // Remove quotes if present
     let filename = if filename.starts_with('"') && filename.ends_with('"') && filename.len() >= 2 {
         &filename[1..(filename.len() - 1)]
     } else {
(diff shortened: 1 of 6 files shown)


src/content_disposition.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    None
+}
+
+/// Parse regular filename parameter
+/// Handles both quoted and unquoted filenames
+fn parse_regular_filename(part: &str) -> Option<String> {
+    let filename = part.trim_start_matches("filename=");
+    // Remove quotes if present
```
zuisong, line 34: should we deal with escaped characters? 

httpie has maked it.

```
❯ http  httpbingo.org/response-headers 'Content-Disposition==attachment; filename="fileaaa.pdf\""' -d -v
GET /response-headers?Content-Disposition=attachment%3B+filename%3D%22fileaaa.pdf%5C%22%22 HTTP/1.1
Accept: */*
Accept-Encoding: identity
Connection: keep-alive
Host: httpbingo.org
User-Agent: HTTPie/3.2.4



HTTP/1.1 200 OK
content-disposition: attachment; filename="fileaaa.pdf\""
content-length: 145
content-type: application/json; charset=utf-8
date: Thu, 10 Apr 2025 05:25:18 GMT
fly-request-id: 01JRF2HRJTM7WJFYQFHEKBF5RY-dfw
server: Fly/77d8e9253 (2025-04-09)
via: 1.1 fly.io, 1.1 fly.io

Downloading to fileaaa.pdf"
Done. 145 bytes in 00:0.02463 (5.9 kB/s)
```

zuisong: in this case. chrome rename filename as `fileaaa.pdf_` .

I think keeping this implementation is fine.

src/content_disposition.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    let encoded_filename = parts[2];
+
+    // Decode using percent-encoding
+    let decoded = percent_decode(encoded_filename.as_bytes())
+        .decode_utf8()
+        .ok()?;
+
+    Some(decoded.into_owned())
```
blyxxyz, line 76: We should also check that the charset is `UTF-8`. I don't think other charsets are worth supporting but if they show up we should ignore them.

src/content_disposition.rs
```diff
@@ -0,0 +1,127 @@
...
+        let header = "attachment; filename*=UTF-8''%E6%B5%8B%E8%AF%95.pdf";
+        assert_eq!(
+            parse_filename_from_content_disposition(header),
+            Some("测试.pdf".to_string())
+        );
+    }
+
+    #[test]
+    fn test_both_filenames() {
+        // When both filename and filename* are present, filename* should be preferred
```
blyxxyz, line 113: Can you also test that `filename` is used if `filename*` is present but unusable?

src/download.rs
```diff
@@ -60,7 +52,7 @@ fn get_file_name(response: &Response, orig_url: &reqwest::Url) -> String {
         .or_else(|| from_url(orig_url))
         .unwrap_or_else(|| "index".to_string());
 
-    let filename = filename.split(std::path::is_separator).next_back().unwrap();
+    let filename = sanitize_filename::sanitize(&filename);
```
blyxxyz, line 55: I see `sanitize-filename` has a `replacement` option. Maybe we can tell it to replace by `"_"` since that's what most browsers seem to do?

src/content_disposition.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Prioritizes filename* parameter if present, otherwise uses filename parameter
+pub fn parse_filename_from_content_disposition(content_disposition: &str) -> Option<String> {
+    let parts: Vec<&str> = content_disposition
+        .split(';')
+        .map(|part| part.trim())
+        .collect();
+
+    // First try to find filename* parameter
+    for part in parts.iter() {
+        if part.starts_with("filename*=") {
```
blyxxyz, line 13: if you do it like this:
```suggestion
        if let Some(value) = part.strip_prefix("filename*=") {
```
then you don't need `trim_start_matches` in the function

blyxxyz (commented):
Nice!

zuisong:
😁 Thanks, I've incorporated all your suggested improvements.

src/content_disposition.rs
```diff
@@ -0,0 +1,166 @@
...
+    // Percent-decode the encoded filename into bytes.
+    let decoded_bytes = percent_decode_str(encoded_filename).collect::<Vec<u8>>();
+
+    if charset.eq_ignore_ascii_case("UTF-8") {
+        if let Ok(decoded_str) = String::from_utf8(decoded_bytes) {
+            return Some(decoded_str);
+        }
+    } else if charset.eq_ignore_ascii_case("ISO-8859-1") {
+        // Use the encoding_rs crate to decode ISO-8859-1 bytes.
+        let decoded: String = decoded_bytes.iter().map(|&b| b as char).collect();
```
blyxxyz, line 78: The RFC says ISO-8859-1:1998 (which you do here) but I checked Firefox and Chromium and they actually decode it as Windows-1252. When they see this header:
```
content-disposition: attachment; filename*=iso-8859-1'en'a%99b
```
They decode the filename as `a™b`.

So this is the way to go:
```suggestion
        // RFC 5987 says to use ISO/IEC 8859-1:1998.
        // But Firefox and Chromium decode %99 as ™ so they're actually using
        // Windows-1252. This mixup is common on the web.
        // This affects the 0x80-0x9F range. According to ISO 8859-1 those are
        // control characters. According to Windows-1252 most of them are
        // printable characters.
        // They agree on all the other characters, and filenames shouldn't have
        // control characters, so Windows-1252 makes sense.
        if let Some(decoded_str) = encoding_rs::WINDOWS_1252
            .decode_without_bom_handling_and_without_replacement(&decoded_bytes)
        {
            return Some(decoded_str.into_owned());
        }
```

src/content_disposition.rs
```diff
@@ -0,0 +1,166 @@
...
+        return Some(decoded);
+    } else {
+        // Unknown charset. As a fallback, try interpreting as UTF-8.
+        if let Ok(decoded_str) = String::from_utf8(decoded_bytes) {
+            return Some(decoded_str);
+        }
```
blyxxyz, line 84: I tested this too so we might as well write down what I found:
```suggestion
        // Unknown charset. As a fallback, try interpreting as UTF-8.
        // Firefox also does this.
        // Chromium makes up its own filename. (Even if `filename=` is present.)
        if let Ok(decoded_str) = String::from_utf8(decoded_bytes) {
            return Some(decoded_str);
        }
```

src/content_disposition.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    None
+}
+
+#[cfg(test)]
+mod tests {
```
blyxxyz, line 91: Can you add a test for `%99` too?

blyxxyz (approved):
Works great! `sanitize-filename` is also a good addition.

One more tweak and then I think this is ready.

Issue #284, opened 2022-09-26
Add support for OAuth 2.0 authentication flows


jirutka:
Can you please consider adding support for OAuth 2.0, at least the client_credentials grant?

Example:

```sh
xh --auth-type oauth2 --auth <client_id>:<client_secret> --oauth-grant client_credentials --oauth-token-url https://example.org/oauth/token GET example.org/api/v1/hello
```

ducaale:
@jirutka Given that we don't support plugins yet, I think we could add auth types that people find useful. Do you know of any CLI tool that supports OAuth 2.0? I know there are a couple of OAuth HTTPie plugins, but they differ from your proposed implementation.

jirutka:
I don’t know about any, that’s basically the reason why I’d like to add it into xh.

blyxxyz:
In general we try to be compatible with HTTPie.

Similar issue for HTTPie (currently empty): httpie/httpie#1427

For reference, this OAuth 1.0 plugin exists (and is listed in the official docs): https://github.com/qcif/httpie-oauth1

While it's ugly I think we'd want to go for the same approach of cramming everything inside the `--auth` option. That doesn't bloat the options and makes it much easier to persist in session files.

A popular OAuth 2.0 crate exists: https://lib.rs/crates/yup-oauth2

jirutka:
It was just an example, I don’t insist on that choice of options at all. :)

ducaale:
There is https://github.com/ctron/oidc-cli which supports both `client_credentials` (confidential) and `authorization_code` (public) flows

```sh
# taken from examples in ctron/oidc-cli/README.md

$ oidc create confidential my-client \
  --issuer https://example.com/realm \
  --client-id foo \
  --client-secret bar

$ xh example.com/api $(oidc token -H my-client)
```

PR #420, opened 2025-05-31
adding jwt integration

niels-demeyer:
created a way to make it easier to request endpoints using JWT tokens.
- token is stored locally
- automatic refresh capabilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # xh
+
 [![Version info](https://img.shields.io/crates/v/xh.svg)](https://crates.io/crates/xh)
 [![Packaging status](https://repology.org/badge/tiny-repos/xh.svg)](https://repology.org/project/xh/versions)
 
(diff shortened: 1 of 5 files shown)


ducaale:
Hi, thanks for taking the time to work on this. While I'm open to adding native OAuth2 support, I would like us to agree on the approach first. See https://github.com/ducaale/xh/issues/284 which proposes adding `--auth-type=oauth2`.


A few notes:
- OAuth can be [messy](https://www.nango.dev/blog/why-is-oauth-still-hard) so I'm interested in a solution that allows users to add custom auth script e.g `--auth-type=custom --auth=./oauth2.sh`
- We already support persisting headers and cookies via `--session=<name>`
- We have `--auth-type=bearer --auth=<token>` for bearer token authentication
- We should avoid introducing too many flags for a single functionality
- When adding a new auth in the future, I recommend taking a look at how digest authentication is implemented using a middleware

I will close this PR for now but feel free to propose a solution in https://github.com/ducaale/xh/issues/284

PR #426, opened 2025-08-16
Remove x86_64-apple-darwin from CI

ducaale:
See https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#demoting-x86-64-apple-darwin-to-tier-2-with-host-tools


blyxxyz:
Do we need to do this? We could wait until the target loses host tools or visibly breaks in some other way. "Tier 2 with host tools" is still supported by rustup and so on.

(I do like the fallback in the install script.)

ducaale:
GitHub will also stop providing free Apple-Intel runners in November. Maybe we can wait until then?

PR #432, opened 2025-11-23
Add colors to --help/-h

starsep:
I used [clap-cargo](https://github.com/crate-ci/clap-cargo) as it's easy to use and provides sensible default styles.
User can disable colors via `NO_COLOR=1` environment variable

Output of `xh --help`:
| before | after | `NO_COLOR=1` |
| - | - | - |
| xh-before | xh-color | xh-no-color |

Edit: update after screenshot with styling without dependency

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -11,6 +11,8 @@ use std::str::FromStr;
 use std::time::Duration;
 
 use anyhow::{anyhow, Context};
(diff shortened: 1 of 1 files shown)


blyxxyz:
I'm okay with colors given that so much of the other output is colorful but is there a way to do this without adding a whole dependency?

PR #447, opened 2026-01-14
Document `ALL_PROXY` env variable

ducaale:
See https://github.com/seanmonstar/reqwest/pull/1856 and https://github.com/seanmonstar/reqwest/pull/2370 which added support `ALL_PROXY`

### DIFF
--- a/doc/xh.1
+++ b/doc/xh.1
@@ -258,7 +258,7 @@ If your proxy requires credentials, put them in the URL, like so: \-\-proxy http
 
 You can specify proxies for multiple protocols by repeating this option.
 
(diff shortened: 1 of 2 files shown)


doc/man-template.roff
blyxxyz: Looks like `HTTP_PROXY` also works nowadays, so we could list that casing instead.

HTTPie also supports it. But curl doesn't (for [CGI security reasons](https://github.com/curl/curl/blob/82e0d387a2e110d704bb18c42894749548b804b4/lib/url.c#L2064)), so `http_proxy` is still more standard.

blyxxyz (approved):

PR #450, opened 2026-02-17
Pretty-print XML responses

### LINKED ISSUE #231: Feature request: pretty print HTML / XML
It would be nice if HTML and XML were pretty printed the same way JSON is.

o1x3:
Closes #231

Uses `quick-xml` to parse and re-indent XML response bodies. Works for both buffered and streaming (`--stream`) output.

`quick-xml` 0.38 is already in the dependency tree (via `plist`/`syntect`), so this doesn't add a new crate to the lockfile.

### How it works

Reads XML events with `trim_text(false)` to preserve meaningful whitespace (e.g. `<p>Hello <b>world</b></p>`), but skips whitespace-only text nodes so that already-indented input gets re-indented without doubling up.

For the streaming+color path, a small `HighlightingWriter` adapter bridges quick-xml's event-based output with syntect's line-based highlighting. quick-xml writes formatted bytes and calls `flush()` after each event, and the adapter highlights and forwards them to the terminal on each flush.

Invalid XML falls back to syntax-highlight-only, same as how invalid JSON is handled.

### Format options

- `--format-options=xml.format:false` disables formatting
- `--format-options=xml.indent:2` changes indent (default 4, matching JSON)

These were already recognized by the CLI but rejected as "Unsupported".

### Not included

HTML pretty-printing. HTML needs error recovery, void elements, implicit closing. Fundamentally different from XML. Could be a separate PR if there's interest.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -140,7 +140,7 @@ impl Printer {
             format_json: format_options.json_format.unwrap_or(pretty.format()),
             json_indent_level: format_options.json_indent.unwrap_or(4),
             format_xml: format_options.xml_format.unwrap_or(pretty.format()),
-            xml_indent_level: format_options.xml_indent.unwrap_or(4),
+            xml_indent_level: format_options.xml_indent.unwrap_or(2),
             sort_headers: format_options.headers_sort.unwrap_or(pretty.format()),
             color: pretty.color(),
             stream: stream.into(),
@@ -579,10 +579,10 @@ impl From<&str> for ContentType {
     fn from(content_type: &str) -> Self {
         if content_type.contains("json") {
             ContentType::Json
-        } else if content_type.contains("html") {
-            ContentType::Html
         } else if content_type.contains("xml") {
             ContentType::Xml
+        } else if content_type.contains("html") {
+            ContentType::Html
         } else if content_type.contains("multipart") {
             ContentType::Multipart
         } else if content_type.contains("x-www-form-urlencoded") {
--- a/tests/cases/xml.rs
+++ b/tests/cases/xml.rs
@@ -16,10 +16,10 @@ fn xml_pretty_printing() {
         .stdout(indoc! {r#"
             <?xml version="1.0"?>
             <catalog>
-                <book id="1">
-                    <author>Gambardella</author>
-                    <title>XML Developer Guide</title>
-                </book>
+              <book id="1">
+                <author>Gambardella</author>
+                <title>XML Developer Guide</title>
+              </book>
             </catalog>
 
 
@@ -39,7 +39,7 @@ fn xml_pretty_printing_text_xml_content_type() {
         .assert()
         .stdout(indoc! {r#"
             <root>
-                <a>text</a>
+              <a>text</a>
             </root>
 
 
@@ -75,13 +75,13 @@ fn xml_custom_indent() {
     get_command()
         .args([
             "--print=b",
-            "--format-options=xml.indent:2",
+            "--format-options=xml.indent:6",
             &server.base_url(),
         ])
         .assert()
         .stdout(indoc! {r#"
             <root>
-              <a>text</a>
+                  <a>text</a>
             </root>
 
 
@@ -116,7 +116,7 @@ fn xml_declaration_preserved() {
         .stdout(indoc! {r#"
             <?xml version="1.0" encoding="UTF-8"?>
             <root>
-                <a>text</a>
+              <a>text</a>
             </root>
 
 
@@ -164,7 +164,7 @@ fn xml_mixed_content_preserved() {
         .assert()
         .stdout(indoc! {r#"
             <root>
-                <p>Hello <b>world</b> end</p>
+              <p>Hello <b>world</b> end</p>
             </root>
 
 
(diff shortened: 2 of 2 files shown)


src/formatting/mod.rs
```diff
@@ -23,6 +25,59 @@ pub fn get_json_formatter(indent_level: usize) -> jsonxf::Formatter {
...
+    let mut reader = Reader::from_reader(BufReader::new(input));
+    reader.config_mut().trim_text(false);
+    let mut writer = Writer::new_with_indent(Vec::new(), b' ', indent);
+    let mut buf = Vec::new();
+    loop {
+        match reader.read_event_into(&mut buf) {
+            Ok(Event::Eof) => break,
+            Ok(Event::Text(ref e)) if e.iter().all(|b| b.is_ascii_whitespace()) => {}
+            Ok(event) => writer.write_event(event).map_err(io::Error::other)?,
+            Err(e) => return Err(io::Error::new(io::ErrorKind::InvalidData, e)),
```
ducaale: Is there a way to recover from invalid xml errors while streaming? if not, we can limit pretty-printing to non-streamed responses

o1x3:
@ducaale both comments addressed. Please let me know if you want these squashed into one commit

src/formatting/mod.rs
```diff
@@ -23,6 +25,25 @@ pub fn get_json_formatter(indent_level: usize) -> jsonxf::Formatter {
...
+pub fn format_xml(indent: usize, text: &str, mut out: impl Write) -> io::Result<()> {
+    let mut reader = Reader::from_str(text);
+    reader.config_mut().trim_text(false);
+    let mut writer = Writer::new_with_indent(Vec::new(), b' ', indent);
+    loop {
+        match reader.read_event() {
+            Ok(Event::Eof) => break,
+            Ok(Event::Text(ref e)) if e.iter().all(|b| b.is_ascii_whitespace()) => {}
+            Ok(event) => writer.write_event(event).map_err(io::Error::other)?,
+            Err(e) => return Err(io::Error::new(io::ErrorKind::InvalidData, e)),
```
blyxxyz: If `e` is an `Error::Io` then its inner error should be returned.

(And then for full pedantic correctness `io::ErrorKind::Interrupted` should be handled with a `continue`, see e.g. https://github.com/ducaale/xh/blob/f30aaa1142dda8e6d6b7004370f7e0d9a036cfc4/src/printer.rs#L307-L315)

blyxxyz (commented):
Very clean implementation, thank you!

The code looks good, haven't tested it yet.

src/formatting/mod.rs
```diff
@@ -23,6 +25,25 @@ pub fn get_json_formatter(indent_level: usize) -> jsonxf::Formatter {
...
+        match reader.read_event() {
+            Ok(Event::Eof) => break,
+            Ok(Event::Text(ref e)) if e.iter().all(|b| b.is_ascii_whitespace()) => {}
+            Ok(event) => writer.write_event(event)?,
+            Err(e) => return Err(io::Error::new(io::ErrorKind::InvalidData, e)),
+        }
+    }
+    let mut buf = writer.into_inner();
+    buf.extend_from_slice(b"\n\n");
+    Ok(buf)
```
blyxxyz: For consistency with the JSON formatting I'd like to add the newlines outside the function:
```suggestion
    Ok(writer.into_inner())
```

blyxxyz (approved):
Thanks for bearing with me. This time I tested it with a few different documents, particularly https://github.com/shlomif/perl-XML-LibXML/blob/master/example/test.xhtml. It works great!

The formatting isn't identical to HTTPie though:
<details>
<summary>Original XML</summary>

```xml
</p></dd><dt><strong><a name="item_toString">toString</a></strong></dt><dd><p><strong>toString</strong> is a deparsing function, so the DOM Tree can be translated into a string,
ready for output. The optional <strong>$format</strong> parameter sets the indenting of the output. This parameter is expected to
be an <em>integer</em> value, that specifies the number of linebreaks for each node. For more
information about the formatted output check the documentation of <em>xmlDocDumpFormatMemory</em> in <em>libxml2/tree.h</em> .

</p>
```
</details>
<details>
<summary>xh output</summary>

```xml
      </dd>
      <dt>
        <strong>
          <a name="item_toString">toString</a>
        </strong>
      </dt>
      <dd>
        <p>
          <strong>toString</strong> is a deparsing function, so the DOM Tree can be translated into a string,
ready for output. The optional <strong>$format</strong> parameter sets the indenting of the output. This parameter is expected to
be an <em>integer</em> value, that specifies the number of linebreaks for each node. For more
information about the formatted output check the documentation of <em>xmlDocDumpFormatMemory</em> in <em>libxml2/tree.h</em> .

</p>
```
</details>
<details>
<summary>HTTPie output</summary>

```xml
      </dd>
      <dt>
        <strong>
          <a name="item_toString">toString</a>
        </strong>
      </dt>
      <dd>
        <p>
          <strong>toString</strong>
           is a deparsing function, so the DOM Tree can be translated into a string,
ready for output. The optional
          <strong>$format</strong>
           parameter sets the indenting of the output. This parameter is expected to
be an
          <em>integer</em>
           value, that specifies the number of linebreaks for each node. For more
information about the formatted output check the documentation of
          <em>xmlDocDumpFormatMemory</em>
           in
          <em>libxml2/tree.h</em>
           .
        </p>
      </dd>
```
</details>

HTTPie changes the whitespace more. Each tag gets its own line even if it's nested between text nodes, and leading and trailing whitespace are removed.

It might be nice to do that too but it doesn't look like quick-xml has an easy way to do it? `trim_text(true)` produces results like this, which is IMO worse:
```xml
<p>As an equivalent of<strong>createElement</strong>, but it creates a<strong>Text Node</strong>bound to the DOM.</p>
```

So I'm fine with merging this.

o1x3:
quick-xml gives you events one at a time (Text, Start, Text, End...). It doesn't give you "here's a mixed-content parent with all its children." You'd have to buffer children of each element, detect mixed content, then re-emit with custom splitting logic.

src/printer.rs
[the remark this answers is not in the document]
o1x3: Done, xml is now checked before html. Also added a test for application/xhtml+xml.

PR #462, opened 2026-05-01
Generate man page as markdown

### LINKED ISSUE #446: Combine README and man page
It is quite frustrating that README and the man page are so different so to get more detailed info on usage (env vars) i had to look into the man page. Would it be better to include the whole man page into the README file?

ducaale:
Since github doesn't support rendering roff files, we will use `--generate markdown` to generate `xh.1.md` which is then linked to from `README.md`. The markdown format is inspired by https://github.com/rtomayko/ronn/blob/master/man/ronn.1.ronn

I have also updated `--generate man` to output definitions using roff controls. This should improve how they are rendered in online man pages like https://man.archlinux.org/man/extra/xh/xh.1.en (See `--pretty`, `--generate`, etc)

Resolves #446

### DIFF
--- a/doc/xh.1.md
+++ b/doc/xh.1.md
@@ -115,7 +115,8 @@ Each `--OPTION` can be reset with a `--no-OPTION` argument.
 
 - `-s`, `--style`=`THEME`: Output coloring style.
 
-  [possible values: auto, solarized, monokai, fruity]
+  [possible values: `auto`, `solarized`, `monokai`, `fruity`]
+
 - `--response-charset`=`ENCODING`: Override the response encoding for terminal display purposes.
   
   Example: --response-charset=latin1.
@@ -186,7 +187,8 @@ Each `--OPTION` can be reset with a `--no-OPTION` argument.
 
 - `-A`, `--auth-type`=`AUTH_TYPE`: Specify the auth mechanism.
 
-  [possible values: basic, bearer, digest]
+  [possible values: `basic`, `bearer`, `digest`]
+
 - `-a`, `--auth`=`USER[:PASS] | TOKEN`: Authenticate as USER with PASS (-A basic|digest) or with TOKEN (-A bearer).
   
   PASS will be prompted if missing. Use a trailing colon (i.e. "USER:") to authenticate with just a username.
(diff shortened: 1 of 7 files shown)


doc/xh.1.md
```diff
@@ -0,0 +1,406 @@
...
+  
+  To construct a complex JSON object, the REQUEST_ITEM's key can be set to a JSON path instead of a field name. For more information on this syntax, refer to https://httpie.io/docs/cli/nested-json.
+
+## OPTIONS
+
+Each `--OPTION` can be reset with a `--no-OPTION` argument.
+
+- `-j`, `--json`: (default) Serialize data items from the command line as a JSON object.
+  
+  Overrides both --form and --multipart.
```
blyxxyz, line 77: It would be nice to backtick `--form` and `--multipart` here (and so on for other option mentions and for examples). But that also applies to the existing man page formatting so it's very separable from this PR.

ducaale:
This produces HTML that renders similar to https://man.archlinux.org/man/extra/xh/xh.1.en
```sh
mandoc -T html -O style=https://man.archlinux.org/static/archmanweb/man_page.css doc/xh.1
```

src/generation.rs
```diff
@@ -214,3 +316,152 @@ fn generate_manpages(app: &mut clap::Command) {
...
+
+    for line in body.lines() {
+        if let Some(line) = line.strip_prefix("    ") {
+            if !indent {
+                parsed.push(ParsedHelp::Indent);
+                indent = true;
+            }
+            if let Some(line) = line.strip_prefix("    ") {
+                if let Some(ParsedHelp::Definition(_, defintion)) = parsed.last_mut() {
+                    defintion.replace(line);
```
blyxxyz:
```suggestion
                if let Some(ParsedHelp::Definition(_, description)) = parsed.last_mut() {
                    description.replace(line);
```

src/cli.rs
```diff
@@ -1360,6 +1361,7 @@ pub enum Generate {
     CompletePowershell,
     CompleteZsh,
     Man,
+    Markdown,
```
blyxxyz: Maybe `ManMarkdown`/`man-markdown`?

blyxxyz:
With the new man format there are a few options that mandoc doesn't seem to hyperlink (compared to the old output where only `-4`/`-6` aren't hyperlinked). I can't spot a pattern.

image

Overall the output does look a bit better.

src/generation.rs
```diff
@@ -214,3 +316,152 @@ fn generate_manpages(app: &mut clap::Command) {
 
     print!("{manpage}");
 }
+
+#[derive(Debug, PartialEq)]
+enum ParsedHelp<'a> {
```
blyxxyz, line 321: I like the abstract format and the tests that showcase it, very clean.

blyxxyz (approved):

ducaale:
>With the new man format there are a few options that mandoc doesn't seem to hyperlink

Interesting. It looks like the letters `H`/`B`/`h`/`b`/`m` are hyperlinked inside `-p, --print=FORMAT` so mandoc is not able to hyperlink any options that use those as short option.

doc/xh.1.md
ducaale: Sounds good. I would also like to format examples similar to https://man.archlinux.org/man/curl.1.en

PR #464, opened 2026-05-29
Add support for custom auth plugins

### LINKED ISSUE #424: Plugin system for extending supported auth methods
The goal is to allow users to implement custom auth strategies that aren't currently supported. This would also serve as a substitute for HTTPie's auth plugin system.

Related issues that might benefit from this:
1. https://github.com/ducaale/xh/issues/284
2. https://github.com/ducaale/xh/issues/328
3. https://github.com/ducaale/xh/issues/414

I thought about relying on inline shell invocation (`--auth-type=bearer --auth=$(get-oauth-token qa)`), but that wouldn't work if the auth needs access to the request itself (e.g cURL's `--aws-sigv4`).

One possible approach is to support:

```
xh --auth-type=plugin:hmac-sha256 --auth=signature-secret
```

This would look for an executable like `xh-hmac-sha256` (similar to how cargo handles custom commands) and pass it the request details (i.e URL, headers, body, content of `--auth`). The plugin would output headers to include in the final request.

I would like to hear if this direction makes sense and any other alternatives we should consider.

ducaale:
This PR adds `--auth-plugin=NAME` that instructs `xh` to spawn an executable named `xh-plugin-NAME` and pass the following via stdin:

```json
{
  "url": "http://example.org/api/v1/hello",
  "auth": []
}
```

The plugin will then modify request by outputting JSON object via stdout:

```json
{
  "add_headers": [
    { "name": "x-signature", "value": "12345678" }
  ]
}
``` 

Some notes:
- multiple values can be passed to plugin by repeating `--auth` option e.g `--auth-plugin=oauth2 --auth=client_id:pluto --auth=client_secret:12345 --auth=scope:42`
- env variable `XH_PLUGIN=auth` is passed to plugin
- full path can be used instead of plugin name e.g `--auth-type=/usr/bin/oauth2`
- error message can be customised by outputting `{"error_message": "..."}`

Resolves #424


ctron:
I haven't tried it yet, I up slapped together an example of how this could work. Many thanks.

blyxxyz:
We might want to support plugins for other features besides authentication. Perhaps we can pass a `type` JSON key to the plugin and recommend checking it? In any case we need to work carefully.

> `plugin:NAME` e.g `plugin:hmac-sha256` which instructs xh to spawn an executable named `xh-plugin-NAME`

This makes sense but it seems useful to be able to specify a plugin by its file path, especially if it's not in `$PATH`. Maybe we could not apply the `xh-plugin-` prefix if there's a directory separator in the name? Still feels slightly weird. I wonder what other software does.

> stdin

This is probably the best solution but it makes it slightly awkward for plugins to take interactive user input. (On Unix you can still do this by opening `/dev/tty`.)

> `"body_base64": "eyJoZWxsbyI6IndvcmxkIn0="`

This can be expensive e.g. for large file uploads. How often will plugins need to know the body? Maybe we could figure out a way for the plugin to only request the body if it needs it?

If we leave this out now we can always add it later.

(I see the HMAC example uses it...)

> `"current_dir": "/home/user"`

Can't the plugin look at its own working directory?

ducaale:
>We might want to support plugins for other features besides authentication.

Would an env variable be enough? we're currently passing `XH_AUTH_PLUGIN=1` to the plugin

>Maybe we could not apply the xh-plugin- prefix if there's a directory separator in the name? Still feels slightly weird. I wonder what other software does.

Sounds good :+1:

>This is probably the best solution but it makes it slightly awkward for plugins to take interactive user input. (On Unix you can still do this by opening /dev/tty.)

Hmm, I haven't thought about this. Let me check the behaviour in windows as well.

>Can't the plugin look at its own working directory?

Yeah, you're right. Let me remove it.

>This can be expensive e.g. for large file uploads. How often will plugins need to know the body? Maybe we could figure out a way for the plugin to only request the body if it needs it?

I agree. One thing I was experimented with is to have the plugin configure itself, so if it receives the following:

```json
{
  "next_request": {
    "method": "POST",
    "url": "http://example.org/api/v1/hello",
    "headers": [],
    "body_base64": null
  },
  "auth": [],
  "state": null,
  "current_dir": "/home/user"
  "config": null
}
``` 

it can output something like:

```json
{ "set_config": { "requires_body": true } }
```

Which leads to the plugin being spawned again with the following:

```json
{
  "next_request": {
    "method": "POST",
    "url": "http://example.org/api/v1/hello",
    "headers": [],
    "body_base64": "eyJoZWxsbyI6IndvcmxkIn0="
  },
  "auth": [],
  "state": null,
  "current_dir": "/home/user"
  "config": { "requires_body": true }
}
```

blyxxyz:
> Would an env variable be enough?

Making it part of the payload seems most natural, you can directly express that as a serde enum or TypeScript tagged union.

Configuring could even be a separate mandatory step, so that we send a payload like `{"type": "configure"}` (no other fields) and the plugin sends back `{"requires_body": true}` or just `{}` if the defaults are fine. Though that increases the amount of boilerplate plugins need and it doubles the startup time.

We could also do JSONL, keep the plugin process running and keep sending it new payloads. (And take care not to deadlock if a naïve plugin e.g. stops listening too early because it doesn't take redirects into account?)

It smells like an ad hoc IPC system, we may be reinventing the wheel...

ctron:
> > Would an env variable be enough?
> 
> Making it part of the payload seems most natural, you can directly express that as a serde enum or TypeScript tagged union.

The idea of the env-var was to have a "normal" application just define an alias (symlink) and then trigger the plugin behavior by detecting the env-var. That would not work if you would first need to parse stdin.

> 
> We could also do JSONL, keep the plugin process running and keep sending it new payloads. (And take care not to deadlock if a naïve plugin e.g. stops listening too early because it doesn't take redirects into account?)
> 
> It smells like an ad hoc IPC system, we may be reinventing the wheel...

I think using stdio for things like this and a one shot approach is a fairly common pattern in Unix-ish OSes. Also for Windows. Creating a service and keeping things running makes things rather complicated.

blyxxyz:
> The idea of the env-var was to have a "normal" application just define an alias (symlink) and then trigger the plugin behavior by detecting the env-var. That would not work if you would first need to parse stdin.

I forgot that the env var name contained `AUTH`. I was imagining a system where an env var is used to tell the program that it's running as a plugin at all and the payload communicates what kind of plugin or which step in the process it's in. (Not sure that's actually better.)

It's unfortunate that env vars get inherited by the plugin's child processes but that's probably unlikely to cause problems in practice. (We could pass it as an argument instead.)

Instead of `XH_AUTH_PLUGIN=1` we could do `XH_PLUGIN=auth`. How useful that is depends on whether the same program might want to implement multiple plugin types.

> `{ "set_config": { "requires_body": true } }`

If we do this, do we also allow other commands in that same response (with the possibility that the plugin modifies headers twice)?

A poorly written plugin might confuse an empty body for a missing body, so maybe we should error if the plugin calls `set_config` twice to avoid infinite loops? (Could be too paranoid.)

ctron:
> Instead of `XH_AUTH_PLUGIN=1` we could do `XH_PLUGIN=auth`.

Sounds like a good idea!

ducaale:
>>Maybe we could not apply the xh-plugin- prefix if there's a directory separator in the name? Still feels slightly weird. I wonder what other software does.
>
>Sounds good 👍

Note that this would be in line with how git finds external helpers. From [`gitcredentials(7)`](https://man7.org/linux/man-pages/man7/gitcredentials.7.html#CUSTOM_HELPERS):
>1. If the helper string begins with "!", it is considered a shell
           snippet, and everything after the "!" becomes the command.
>2. Otherwise, if the helper string begins with an absolute path,
           the verbatim helper string becomes the command.
>3. Otherwise, the string "git credential-" is prepended to the
           helper string, and the result becomes the command.

ducaale:
Does it make sense to still use the `plugin:` prefix when specifying a full path? e.g `--auth-type=plugin:/path/to/plugin` vs `--auth-type=/path/to/plugin`

**Update**: `--auth-type` now accepts either `plugin-NAME` (interpreted as `xh-plugin-NAME`) or `/path/to/plugin`

ducaale:
Couple of questions to think about:
- How can we support multi-step authentication like digest or kerberos? git-credential-helpers use [`continue`](https://git-scm.com/docs/git-credential#Documentation/git-credential.txt-continue) keyword to retry a request. According to this [talk](https://www.youtube.com/watch?v=3Gy3-lA7xls), it is limited to 3 round trips.
- Should we allow plugins to persist headers if `--session=name_or_path` is used?
- How should we handle versioning? maybe we can pass xh version as env variable?

ducaale:
One thing we can do is to have a separate mode for advanced use cases e.g signing requests or multiple-step authentication.

Initially, plugin receives the following via stdin with `XH_PLUGIN=simple_auth`:
```json
{
  "url": "http://example.org/api/v1/hello",
  "auth": ["secret-token"]
}
```

Plugin can either specify what headers to add or upgrade to advanced mode:

```jsonc
// modify request **once**
{
  "add_headers": [{ "name": "x-api-key", "value": "secret-token" }]
}

// upgrade to advanced mode
{
  "action": "upgrade_to_advanced_auth"
}
```

In the latter case, we re-spawn plugin with `XH_PLUGIN=advanced_auth` and all communication happens via JSONL. In this mode, plugin can read headers, body, prompt user for more input, modify redirected requests, retry request if response contains `www-authenticate` header, etc

```jsonc
// xh
{ "type": "request", "url": "http://example.org/api/v1/hello", "auth": [] }
// plugin
{ "action": "prompt", "message": "Enter signing secret for example.com" }
// xh
{ "value": "secret-signing-1234" }
// plugin
{ "action": "read_body_base64" }
// xh
{ "value": "eyJoZWxsbyI6IndvcmxkIn0=" }
// plugin 
{ "action": "modify_headers", "add_headers": [{ "name": "x-signature", "value": "12345" }] }
// xh
{}
// plugin
{ "action": "send" }
// xh (redirect to http://example.org/api/v2/hello)
{ "type": "request", "url": "http://example.org/api/v2/hello", "auth": [] }
// plugin
{ "action": "send" }
// xh
{ "type": "end" }
```

ctron:
I guess I'm ok just with "simple auth". Maybe "advanced" should be another PR.

An area of improvement could be the way an error is reported. Maybe allow the plugin to return some "reason" when having a problem. So integrate the error response into `xh`, rather than having two CLIs on stderr.

ducaale:
>>This is probably the best solution but it makes it slightly awkward for plugins to take interactive user input. (On Unix you can still do this by opening /dev/tty.)
>
>Hmm, I haven't thought about this. Let me check the behaviour in windows as well.

On Windows, you can replace `/dev/tty` with `con` (thanks @hadiosman271 for finding this):

```py
if platform.system() == 'Windows':
    sys.stdin = open("con")
else:
    sys.stdin = open("/dev/tty")
sys.stderr.write(f"xh-plugin-token: token for {url.hostname}: ")
token = input("")
```

src/cli.rs
```diff
@@ -244,8 +244,29 @@ Example: --print=Hb"
...
-    /// Specify the auth mechanism.
-    #[clap(short = 'A', long, value_enum)]
+    /// Specify the auth mechanism. Supported auth types are "basic", "bearer" and "digest".
+    ///
+    /// Custom auth is supported via "plugin-NAME" or "/path/to/plugin". The former
+    /// looks for an executable named "xh-plugin-NAME" in "$PATH".
+    ///
+    /// The plugin receives request url and --auth values via stdin:
+    ///
+    /// { "url": "http://example.org/api/v1/hello", "auth": [] }
```
ducaale, line 258: I plan to fix how this JSON snippet is rendered in `xh.1.md` in a follow-up PR

src/auth.rs
```diff
@@ -97,6 +103,89 @@ impl Middleware for DigestAuthMiddleware<'_> {
...
+
+        let output = child
+            .wait_with_output()
+            .context("Failed to wait for plugin output")?;
+
+        if !output.status.success() {
+            if let Ok(PluginResponseErr { error_message }) = serde_json::from_slice(&output.stdout)
+            {
+                return Err(anyhow!(error_message));
+            }
```
blyxxyz, line 169: We should probably log stdout if it fails to parse.

src/auth.rs
```diff
@@ -97,6 +103,89 @@ impl Middleware for DigestAuthMiddleware<'_> {
...
+}
+
+impl AuthPlugin {
+    pub fn exec(name_or_path: OsString, url: String, auth: Vec<String>) -> Result<Self> {
+        let plugin_input = PluginInput::new(&url, &auth);
+
+        let plugin_path = if is_path(&name_or_path) {
+            std::path::PathBuf::from(&name_or_path)
+        } else {
+            std::path::PathBuf::from(format!("xh-{}", name_or_path.to_string_lossy()))
```
blyxxyz: I can live with this but it's possible to do it non-lossily with [`OsString::push()`](https://doc.rust-lang.org/std/ffi/struct.OsString.html#method.push) (and then convert to `PathBuf`).

blyxxyz:
I'll try to review this soon, sorry for the wait.

ducaale:
No worries, thanks!

src/auth.rs
```diff
@@ -97,6 +103,89 @@ impl Middleware for DigestAuthMiddleware<'_> {
...
+
+        if !output.status.success() {
+            if let Ok(PluginResponseErr { error_message }) = serde_json::from_slice(&output.stdout)
+            {
+                return Err(anyhow!(error_message));
+            }
+            if let Some(code) = output.status.code() {
+                return Err(anyhow!("Plugin exited with exit code {}", code));
+            } else {
+                return Err(anyhow!("Plugin exited no exit code"));
```
blyxxyz, line 174: Seems like this can only happen if the process was killed by a signal? [Determining the signal](https://doc.rust-lang.org/std/os/unix/process/trait.ExitStatusExt.html) is possible and useful but maybe not worth the effort. Let's at least put a comment:

```suggestion
                // Usually means termination by a signal
                return Err(anyhow!("Plugin exited no exit code"));
```

src/redirect.rs
```diff
@@ -6,39 +6,38 @@ use reqwest::header::{
 };
 use reqwest::{Method, StatusCode, Url};
 
-#[cfg(feature = "http-message-signatures")]
-use crate::cli::MessageSignature;
 use crate::middleware::{Context, Middleware};
 use crate::utils::{HeaderValueExt, clone_request};
 
-pub struct RedirectFollower {
+pub struct RedirectFollower<T> {
```
blyxxyz: Nitpick: `F` is more standard as a name for function type parameters

src/to_curl.rs
```diff
@@ -362,6 +362,9 @@ pub fn translate(args: Cli) -> Result<Command> {
                 cmd.arg("--oauth2-bearer");
                 cmd.arg(auth);
             }
+            AuthType::Plugin(..) => {
+                cmd.warn("Ignored -A/--auth-type".to_string());
+            }
```
blyxxyz: This only gets printed if `--auth` is used, but a plugin might not need `--auth` at all.

src/session.rs
```diff
@@ -273,7 +273,14 @@ impl Session {
                     raw_auth: Some(token.into()),
                 }
             }
+            auth::Auth::Plugin(auth_plugin) => {
+                let auth_headers = auth_plugin.headers.clone();
+                let mut session_headers = self.headers()?;
+                session_headers.extend(auth_headers);
+                self.save_headers(&session_headers)?;
+            }
```
blyxxyz: Saving the headers is probably undesirable for many plugins.

Not sure what makes sense here... three possibilities:

- Save all the `--auth` values (and extend the schema to accommodate this)
- Save nothing
- Make `--session` and auth plugins incompatible with each other (i.e. crash) until we make a decision

We could later extend the plugin protocol to let plugins manage their own state.

If we store the plugin name/path itself in the session file then loading a session files could cause arbitrary commands to run which we'd maybe want to avoid.

src/cli.rs
```diff
@@ -802,12 +823,31 @@ fn construct_url(
...
+        _ => {
+            let auth_type_str = auth_type.to_string_lossy();
+            if auth_type_str.starts_with("plugin-")
+                || auth_type_str.contains(std::path::is_separator)
+            {
+                Ok(AuthType::Plugin(auth_type))
+            } else {
+                Err(anyhow!("Unknown auth_type {:?}", auth_type))
+            }
+        }
```
blyxxyz: Maybe it'd be cleaner to have a separate `--auth-plugin=` option that overrides (and is overridden by) `--auth-type`?

src/auth.rs
```diff
@@ -97,6 +103,89 @@ impl Middleware for DigestAuthMiddleware<'_> {
     }
 }
 
+#[derive(Debug, Serialize)]
+struct PluginInput<'a> {
```
blyxxyz, line 105: If we go with the upgrade action design then we probably don't need protocol versioning but sending the xh version along still seems worthwhile.

Not sure if a string or a tuple (`[0, 27, 1]`) is best. In Python you can compare tuples easily:
```pycon
>>> [1, 0, 0] >= [0, 27, 1]
True
>>> [0, 0, 0] >= [0, 27, 1]
False
>>> [0, 27, 0] >= [0, 27, 1]
False
>>> [0, 27, 2] >= [0, 27, 1]
True
```
But people might also be tempted to only compare the minor version which would blow up once we reach 1.0.0.

src/auth.rs
```diff
@@ -97,6 +103,89 @@ impl Middleware for DigestAuthMiddleware<'_> {
...
+#[derive(Debug, Deserialize)]
+struct Header {
+    name: String,
+    value: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct PluginResponse {
+    add_headers: Vec<Header>,
+}
```
blyxxyz, line 125: We need to think carefully about how to handle unknown keys (for outdated versions of xh running newer plugins).

I don't think crashing on them is the way to go, there could be future keys for e.g. caching that are okay to ignore, but a `log::debug!()` would be helpful. (Maybe even `log::warn!()`.)

And then maybe `add_headers` should already not be required?

tests/cases/auth_plugin.rs
```diff
@@ -0,0 +1,149 @@
...
+use indoc::indoc;
+
+use crate::prelude::*;
+
+fn path_with_plugins_dir() -> OsString {
+    let path = env::var_os("PATH").expect("PATH var missing");
+    env::join_paths(env::split_paths(&path).chain([PathBuf::from(concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/tests/fixtures/plugins"
+    ))]))
```
blyxxyz: The test directory should go at the front, not the back, or other entries could theoretically get in the way.

blyxxyz (commented):
A few details to sort out but I think I'm happy with the approach and we're not locking ourselves into something we'll regret.

src/session.rs
ducaale: I will remove this logic for now. In the future, we can have plugins opt-in into saving header e.g

```jsonc
{
  "add_headers": [
    // TODO: think of a better name
    { "name": "x-signature", "value": "12345678", "can_persist_in_session": true }
  ]
}
```

ducaale: >If we store the plugin name/path itself in the session file then loading a session files could cause arbitrary commands to run which we'd maybe want to avoid.

We might want to ban plugins in `default_cli_args()` to avoid a similar thing

blyxxyz: I'm not as worried about that one because `~/.config` feels basically trusted while session files can live anywhere. But worth considering yeah.

PR #467, opened 2026-07-07
Fix --auth ignoring credentials with an empty username

upuddu:
`xh -a :password` (an empty username with a password) currently prompts for a password instead of using the credentials, unlike HTTPie which sends `Authorization: Basic OnBhc3M=`. The credential regex required at least one username character (`.+?`); this loosens it to `.*?` so an empty username before the colon is accepted, leaving the existing `user:`, `:`, and colons-in-password cases unchanged.


ducaale (approved):
Thanks!

PR #479, opened 2026-08-10
Stabilize httpsig CLI with curl-compatible defaults

zuisong:
Replace the experimental --unstable-m-sig-* flags with curl-style --httpsig-* options:

- --httpsig-keyid, --httpsig-key (hex or @file), --httpsig-algo, --httpsig-headers with space-separated components
- Keep only hmac-sha256 and ed25519, drop the other algorithms
- Derive method/authority/path/query components instead of @-shorthands
- Add MessageSignatureKey parsing and simplify message_signature.rs
- Update completions, man page, docs, and tests


Cargo.toml
```diff
@@ -97,11 +96,11 @@ hyper-util = { version = "0.1.3", features = ["server"] }
 http-body-util = "0.1.1"
 
 [features]
-default = ["online-tests", "rustls", "network-interface"]
+default = ["online-tests", "rustls", "network-interface", "http-message-signatures"]
```
ducaale: We can't enable this by default since httpsig is still in 0.0.x

src/redirect.rs
```diff
@@ -7,25 +7,35 @@ use reqwest::header::{
...
+    explicit_signature_headers: HeaderMap,
+    #[cfg(feature = "http-message-signatures")]
+    preserve_signature_headers: bool,
 }
 
 impl RedirectFollower {
     pub fn new(
         max_redirects: usize,
-        #[cfg(feature = "http-message-signatures")] message_signature: Option<MessageSignature>,
+        #[cfg(feature = "http-message-signatures")] httpsig: Option<HttpsigOptions>,
```
ducaale: Can we have this in a `on_redirect` closure similar to how it is done in https://github.com/ducaale/xh/pull/464?

src/main.rs
```diff
@@ -108,6 +108,8 @@ fn run(args: Cli) -> Result<ExitCode> {
     }
 
     let (mut headers, headers_to_unset) = args.request_items.headers()?;
+    // Keep the user-provided fields separate from xh's generated defaults so
+    // HTTP Message Signatures only cover headers the caller explicitly set.
```
ducaale: I don't think we need this comment

src/main.rs
```diff
@@ -424,6 +426,13 @@ fn run(args: Cli) -> Result<ExitCode> {
         }
     }
 
+    // Session headers are also user-configured fields. Capture them after
+    // session loading, while excluding cookies moved into the cookie jar.
+    let mut explicit_signature_headers = headers.clone();
+    #[cfg(feature = "http-message-signatures")]
+    let signature_headers_preprovided =
+        headers.contains_key("signature") || headers.contains_key("signature-input");
+
```
ducaale: Let's remove this unless there is a good reason

ducaale (commented):
Some initial comments

src/cli.rs
```diff
@@ -811,121 +812,272 @@ pub enum AuthType {
...
+            .collect::<Result<Vec<_>, _>>()?;
+
+        if components.is_empty() {
+            return Err("no HTTP message signature components specified".to_string());
+        }
+        if components.len() > MAX_MESSAGE_SIGNATURE_COMPONENTS {
+            return Err(format!(
+                "too many HTTP message signature components (maximum {MAX_MESSAGE_SIGNATURE_COMPONENTS})"
+            ));
+        }
```
ducaale: I don't think we need this

src/main.rs
zuisong: If the snapshot (explicit_signature_headers) is removed, do we need to handle the case where a user asks to sign a header (e.g. --httpsig-headers="user-agent:") but never explicitly set that header?

zuisong: I think we can remove explicit_signature_headers. Our goal can be restated as: sign only what is actually sent in the request.

src/redirect.rs
zuisong: done

src/main.rs
```diff
@@ -590,34 +590,20 @@ fn run(args: Cli) -> Result<ExitCode> {
...
-                &mut request,
-                key_id,
-                key_material,
-                (!m_sig_components.is_empty()).then_some(m_sig_components.as_slice()),
-                m_sig_algorithm,
-            )?;
+        if args.httpsig.is_requested() {
+            if !args.httpsig.has_key_pair() {
+                return Err(anyhow!(
+                    "HTTP message signatures require both --httpsig-keyid and --httpsig-key."
```
ducaale: Can we let clap handle this validation?

src/cli.rs
```diff
@@ -811,121 +812,265 @@ pub enum AuthType {
...
-        Some((self.m_sig_id.as_deref()?, self.m_sig_key.as_deref()?))
+    /// Return the key identifier and key source as one validated pair.
+    #[cfg(feature = "http-message-signatures")]
+    pub fn key_pair(&self) -> Option<(&str, &MessageSignatureKey)> {
+        Some((self.httpsig_key_id.as_deref()?, self.httpsig_key.as_ref()?))
     }
 
+    /// Return the explicitly selected algorithm, if any.
+    #[cfg(feature = "http-message-signatures")]
     pub fn algorithm(&self) -> Option<MessageSignatureAlgorithm> {
```
ducaale: Let's remove this getter

src/cli.rs
```diff
@@ -811,121 +812,265 @@ pub enum AuthType {
...
+        self.httpsig_algorithm
     }
 
-    pub fn has_components(&self) -> bool {
-        self.m_sig_comp
-            .iter()
-            .any(|components| !components.0.is_empty())
+    /// Return the validated component list supplied by the user.
+    #[cfg(feature = "http-message-signatures")]
+    pub fn components(&self) -> Option<&[MessageSignatureComponent]> {
```
ducaale: Same as above

src/cli.rs
```diff
@@ -1868,40 +2009,117 @@ mod tests {
     }
 
     #[test]
-    fn parse_repeated_message_signature_components() {
+    fn repeated_message_signature_components_use_last_value() {
```
ducaale: Remove test

src/cli.rs
```diff
@@ -1868,40 +2009,117 @@ mod tests {
...
-        assert_eq!(cli.m_sig.m_sig_comp[0].0, vec!["@method", "@path"]);
-        assert_eq!(cli.m_sig.m_sig_comp[1].0, vec!["date"]);
+        assert_eq!(
+            cli.httpsig.httpsig_headers.unwrap().0,
+            vec![MessageSignatureComponent::Header("date".to_string())]
+        );
+    }
+
+    #[test]
+    fn message_signature_components_match_curl_syntax() {
```
ducaale: same as above

src/cli.rs
```diff
@@ -1868,40 +2009,117 @@ mod tests {
...
+                duplicate,
+                "get",
+                "example.org",
+            ])
+            .is_err()
+        );
+    }
+
+    #[test]
+    fn message_signature_key_matches_curl_syntax() {
```
ducaale: same as above

src/cli.rs
```diff
@@ -1868,40 +2009,117 @@ mod tests {
...
+                ])
+                .is_ok(),
+                "unexpectedly rejected {key:?}"
+            );
+        }
     }
 
     #[test]
+    #[cfg(feature = "http-message-signatures")]
     fn parse_message_signature_algorithm() {
```
ducaale: same as above

src/cli.rs
```diff
@@ -811,121 +812,265 @@ pub enum AuthType {
...
-    EcdsaP256Sha256,
-    #[clap(name = "ecdsa-p384-sha384")]
-    EcdsaP384Sha384,
-    #[clap(name = "rsa-v1_5-sha256")]
-    RsaV15Sha256,
-    #[clap(name = "rsa-pss-sha512")]
-    RsaPssSha512,
+}
+
+impl fmt::Display for MessageSignatureAlgorithm {
```
ducaale: Can we remove `impl Display` for the following?
- `MessageSignatureComponents`
- `MessageSignatureComponent`
- `MessageSignatureAlgorithm`
- `MessageSignatureKey`

We can move some of the logic to `to_curl.rs` (similar to `http-version`). You will also need to replace display with debug for errors

src/cli.rs
```diff
@@ -811,121 +812,265 @@ pub enum AuthType {
...
     }
+}
 
+/// A curl-compatible HTTP Message Signature key source.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MessageSignatureKey {
+    /// Hex-encoded key material supplied directly on the command line.
+    Hex(String),
+    /// A file path supplied with the curl `@filename` syntax.
+    File(String),
```
ducaale: We usually parse args as `OsString` for paths input. See `--session` (or `--auth-type` in https://github.com/ducaale/xh/pull/464)

src/main.rs
```diff
@@ -699,19 +680,24 @@ fn run(args: Cli) -> Result<ExitCode> {
...
+                    args.max_redirects.unwrap_or(10),
+                    #[allow(unused)]
+                    |previous_url, mut request| {
+                        #[cfg(feature = "http-message-signatures")]
+                        {
+                            // Sign only same-origin redirects. A cross-origin
+                            // redirect is a new trust boundary and must not
+                            // receive the old signature.
+                            if !redirect::is_cross_domain_redirect(request.url(), previous_url)
+                                && args.httpsig.has_key_pair()
```
ducaale: Can we replace `has_key_pair` with `args.httpsig.is_requested()`? Also, it would be nice if the guard returned an Option/Result so we can avoid `message-signature: Missing key or key identifier` error inside `sign_request`

src/cli.rs
```diff
@@ -811,121 +812,213 @@ pub enum AuthType {
...
-#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq)]
+impl FromStr for MessageSignatureComponent {
+    type Err = String;
+
+    fn from_str(component: &str) -> Result<Self, Self::Err> {
+        if component.contains(['"', '\\']) {
+            return Err(format!(
+                "invalid HTTP message signature component `{component}`"
+            ));
+        }
```
ducaale: I think we can remove this check since invalid chars in header name will be caught by reqwest

src/main.rs
ducaale: Also, I wonder if we can remove `signature` and `signature-input` header here instead of inside `redirect.rs`?

src/cli.rs
```diff
@@ -811,121 +812,213 @@ pub enum AuthType {
...
+            return Err("HTTP message signature key file path cannot be empty".to_string());
+        }
+        return Ok(MessageSignatureKey::File(OsString::from(path)));
     }
 
-    pub fn has_components(&self) -> bool {
-        self.m_sig_comp
-            .iter()
-            .any(|components| !components.0.is_empty())
+    if !key.is_empty() && key.len() % 2 == 0 && key.bytes().all(|byte| byte.is_ascii_hexdigit()) {
```
ducaale: Can we have `util::hex` function that we can use here and in `message_signature.rs`?

zuisong:
Sorry for the delay — I've been tied up with some work stuff recently, hope it's not too late.

I've addressed all the feedback from the second review round:
- Extracted utils::is_hex, now shared between --httpsig-key parsing and key file decoding;
- Moved the removal of signature / signature-input out of redirect.rs into the redirect callback, so it stays httpsig-specific;
- sign_request now returns early when no key pair is present instead of raising the Missing key or key identifier error; the redirect callback uses the is_requested() guard, and I dropped has_key_pair() (clap's requires already guarantees the key pair is always present);
- Dropped the invalid-character check for httpsig header components and let reqwest handle it.

Let me know if there's anything else to tweak. I'll mark the PR ready for your final review 🙏

src/message_signature.rs
```diff
@@ -1,50 +1,78 @@
...
+        if let MessageSignatureComponent::Header(name) = component {
+            let header_name = HeaderName::from_bytes(name.as_bytes())
+                .with_context(|| format!("message-signature: Invalid header component: {name}"))?;
+            for value in request.headers().get_all(&header_name) {
+                http_request
+                    .headers_mut()
+                    .append(header_name.clone(), value.clone());
+            }
+        }
+    }
```
ducaale, line 51: Can we check missing headers so we don't need to call `validate_header_components`?

src/message_signature.rs
```diff
@@ -1,50 +1,78 @@
...
+/// Signing covers only headers that are actually present on the request, so
+/// the signature matches exactly what will be sent.
+pub fn sign_request(request: &mut Request, httpsig: &HttpsigOptions) -> Result<()> {
+    // Match curl: a caller-provided Signature or Signature-Input means the
+    // request is already signed and automatic signing is skipped entirely.
+    if request.headers().contains_key("signature")
+        || request.headers().contains_key("signature-input")
+    {
+        return Ok(());
+    }
```
ducaale: Can remove this early return and let `sign_request` override these headers?

src/main.rs
```diff
@@ -699,19 +680,25 @@ fn run(args: Cli) -> Result<ExitCode> {
...
+                client = client.with(RedirectFollower::new(
+                    args.max_redirects.unwrap_or(10),
+                    #[allow(unused)]
+                    |previous_url, mut request| {
+                        #[cfg(feature = "http-message-signatures")]
+                        if args.httpsig.is_requested() {
+                            // A redirect invalidates the existing signature, so
+                            // drop it and re-sign only same-origin redirects; a
+                            // cross-origin redirect is a new trust boundary and
+                            // must not receive the old signature.
```
ducaale: Can we remove this comment?

src/message_signature.rs
```diff
@@ -1,50 +1,78 @@
...
     signature_params.set_alg(&algorithm);
-
-    // Ensure keyid is included in Signature-Input
+    // curl always includes the user-provided key identifier in Signature-Input.
     signature_params.set_keyid(key_id);
 
-    // Preferred path: use upstream sync signing helper.
+    // httpsig-hyper signs http::Request, while the rest of xh uses reqwest::Request.
+    // When the caller supplies Content-Digest, the body is covered through that
+    // field; only request metadata needs to be copied into this temporary request.
```
ducaale: Can you remove this comment?

src/message_signature.rs
```diff
@@ -64,282 +92,148 @@ pub fn sign_request(
     Ok(())
 }
 
-/// Resolves and expands message components for signature coverage.
+/// Resolve curl component names to the exact set covered by this request.
```
ducaale: I feel like some of the comments unnecessarily mention curl. In this case, we are we using the same defaults as curl but I'm not sure why we're calling them "curl components"

Do you mind updating this and other similar comments?

src/message_signature.rs
```diff
@@ -1,50 +1,78 @@
...
+    {
+        return Ok(());
+    }
 
-    let (signing_key, algorithm) = build_signing_key(&key, key_id, algorithm_override)?;
+    // clap's `requires` ensures a key pair is present whenever any --httpsig-*
+    // option is given, so this guard only handles the defensive case.
+    let Some((key_id, key_source)) = httpsig.key_pair() else {
+        return Ok(());
+    };
```
ducaale: Can we pass key pair, algorithm and components [individually](https://www.youtube.com/watch?v=0BXuYlNrUmE) so we don't need this check? we can replace `.is_requested()` with a method that returns those

src/cli.rs
```diff
@@ -811,121 +812,201 @@ pub enum AuthType {
...
+
+impl MessageSignatureComponent {
+    /// Return the canonical identifier expected by the RFC 9421 signing library.
     #[cfg(feature = "http-message-signatures")]
-    pub fn flattened_components(&self) -> Vec<String> {
-        self.m_sig_comp
-            .iter()
-            .flat_map(|components| components.0.iter().cloned())
-            .collect()
+    pub fn rfc_name(&self) -> &str {
```
ducaale: Can we inline this where needed?

src/cli.rs
```diff
@@ -811,121 +812,201 @@ pub enum AuthType {
...
+            Self::Authority => "@authority",
+            Self::Path => "@path",
+            Self::Query => "@query",
+            Self::Header(name) => name,
+        }
+    }
+
+    /// Whether this component names the given HTTP field.
+    #[cfg(feature = "http-message-signatures")]
+    pub fn is_header(&self, expected: &str) -> bool {
```
ducaale: Same as above

src/cli.rs
ducaale, line 983: Same as above

ducaale (commented):
No worries. I have more nitpicks:

ducaale:
At the moment, curl doesn't handle computing `content-digest` header. However, most servers including mastodon expect it when there is a body.

Should we automatically compute it (using `sha-256`) if header is missing and user passes `content-digest:` as component?

zuisong:
> At the moment, curl doesn't handle computing `content-digest` header. However, most servers including mastodon expect it when there is a body.
> 
> Should we automatically compute it (using `sha-256`) if header is missing and user passes `content-digest:` as component?

Agreed—this was in an earlier version, and I’m also in favor of restoring it.

src/cli.rs
```diff
@@ -811,121 +812,171 @@ pub enum AuthType {
...
+    Hex(String),
+    File(OsString),
+}
 
-    pub fn algorithm(&self) -> Option<MessageSignatureAlgorithm> {
-        self.m_sig_alg
+fn parse_message_signature_key(key: OsString) -> Result<MessageSignatureKey, String> {
+    let key = key
+        .to_str()
+        .ok_or("HTTP message signature key must be hexadecimal or @FILE")?;
```
blyxxyz: If we unconditionally convert to string we can just use a `StringValueParser`.

That'll give a more appropriate error message for cases like `@\xFF`.

src/cli.rs
```diff
@@ -811,121 +812,171 @@ pub enum AuthType {
...
+        return Ok(MessageSignatureKey::File(OsString::from(path)));
     }
 
-    pub fn has_components(&self) -> bool {
-        self.m_sig_comp
-            .iter()
-            .any(|components| !components.0.is_empty())
+    if crate::utils::is_hex(key) {
+        return Ok(MessageSignatureKey::Hex(key.to_string()));
     }
```
blyxxyz: It might be cleaner to [parse the key immediately instead of just validating it](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

src/cli.rs
```diff
@@ -811,121 +812,171 @@ pub enum AuthType {
...
+}
 
-    pub fn key_pair(&self) -> Option<(&str, &str)> {
-        Some((self.m_sig_id.as_deref()?, self.m_sig_key.as_deref()?))
-    }
+/// HTTP Message Signature key source.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MessageSignatureKey {
+    Hex(String),
+    File(OsString),
```
blyxxyz: If you replace `OsString` by `PathBuf` the intent is a bit clearer and it saves a few conversions in the tests.

src/main.rs
```diff
@@ -590,34 +590,15 @@ fn run(args: Cli) -> Result<ExitCode> {
         }
 
         #[cfg(not(feature = "http-message-signatures"))]
-        if args.m_sig.m_sig_id.is_some()
-            || args.m_sig.m_sig_key.is_some()
-            || args.m_sig.m_sig_alg.is_some()
-            || args.m_sig.has_components()
-        {
+        if args.httpsig.signing_data().is_some() {
             return Err(anyhow!(
```
blyxxyz: Do we want to keep feature-gating this?

src/message_signature.rs
```diff
@@ -1,44 +1,54 @@
...
         .method(request.method())
         .uri(request.url().as_str())
         .body(reqwest::Body::default())
         .context("message-signature: Failed to build temporary HTTP request")?;
-    *http_request.headers_mut() = request.headers().clone();
+    for component in &components {
+        if let MessageSignatureComponent::Header(name) = component {
+            let header_name = HeaderName::from_bytes(name.as_bytes())
+                .with_context(|| format!("message-signature: Invalid header component: {name}"))?;
+            if !request.headers().contains_key(&header_name) {
```
blyxxyz: curl [says](https://curl.se/docs/manpage.html#--httpsig-headers):
> Header components are taken from "-H" / "[--header](https://curl.se/docs/manpage.html#--header)" options only. Headers curl adds by default (such as "User-Agent") are not signed unless you set them explicitly

Maybe we should do the same? It's backward-compatible to broaden it later.

